PR #113, opened 2025-06-25
refactor pwa install prompt

bordalix:
This PR enhances the UX for PWA installation.

Previously, we started by assuming the app was NOT installable and relied on the browser to trigger a [beforeinstallprompt](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeinstallprompt_event) event for the app to know that it was installable as a PWA.

This approach had a significant issue: the event depends on undocumented and inaccessible user heuristics, leading to inconsistent behaviour. On the same URL, the event would sometimes fire and sometimes not. Additionally, this event is not supported in Firefox and Safari, so I decided to get rid of this. There were too many **false negatives**., i.e. we were not nudging the user to install the app when he could.

What changed:
- We now assume the app is installable by default, unless is running as standalone (i.e. in PWA context) or there's no service worker support in navigator
- We will now generate **false positives**, i.e. if the user is accessing the website in the browser, and also have it installed as PWA, the website will not know that and will nudge the user to install the app

What we loose with this PR:
- The possibility of "click-to-install" the PWA, the user must click the icon on the browser bar or follow the instructions

@Kukks @tiero what do you guys think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "homepage": ".",
   "dependencies": {
     "@arkade-os/sdk": "^0.1.1",
-    "@dotmind/react-use-pwa": "^1.0.4",
     "@ionic/react": "^8.5.6",
     "@scure/base": "^1.2.5",
     "@scure/bip32": "^1.7.0",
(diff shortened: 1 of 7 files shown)


bordalix:
> Can we have a combination? So that if the event triggers, we allow the better click to install behavior?

https://github.com/arkade-os/wallet/pull/113/commits/b5c9eeb2b789757691c5caffbd94045f70e1c2c4 fixes it @Kukks

Kukks (approved):

PR #123, opened 2025-07-27
remove dust checks

louisinger:
Remove the `amountIsBelowMinLimit` in order to allow subdust offchain txs. 

Temporary for test : should be reworked ?

### DIFF
--- a/src/screens/Wallet/Send/Form.tsx
+++ b/src/screens/Wallet/Send/Form.tsx
@@ -30,7 +30,7 @@ import { FiatContext } from '../../../providers/fiat'
 import { ArkNote } from '@arkade-os/sdk'
 
 export default function SendForm() {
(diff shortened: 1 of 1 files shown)


bordalix:
@louisinger shouldn't the ark server respond with `vtxoMinAmount = 1` instead?

louisinger:
> @louisinger shouldn't the ark server respond with `vtxoMinAmount = 1` instead?

No because the current `vtxoMinAmount` must be known by client because it is the minimum amount to collect in order to settle. In other words, since subdust feature, there are 2 minimum limits :
* the minimum you can send (1 sat in our case)
* the minimum intent output amount (vtxo_min_amount in our case)

We need to return both values in arkd's infos, even though this check is already handled server-side by the SubmitTx RPC, so it's technically safe to remove it. I understand it’s inconvenient on the client side, but the "minimum send amount" will most likely always be 1 sat so replace by `> 1` LGTM

PR #124, opened 2025-07-28
fix: lnurl support and integrate OCP Standard for Ark to Ark transaction

TaprootFreak:
Parse URLs with lightning parameter
set fixed amount automatically

### DIFF
--- a/src/lib/lnurl.ts
+++ b/src/lib/lnurl.ts
@@ -9,6 +9,16 @@ type LnUrlResponse = {
   minSendable: number
   maxSendable: number
   metadata: string
+  transferAmounts?: {
+    method: string
+    available: boolean
+  }[]
+}
+
+type ArkMethodResponse = {
+  expiryDate: string
+  address: string
+  hint: string
 }
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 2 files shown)


TaprootFreak:
Yes, with pleasure 

The first argument is that it corresponds to the official definition of LNURL
See https://github.com/lnurl/luds/blob/luds/01.md#fallback-scheme

The second argument is SPAR

SPAR uses static QR codes which are information QR codes to inform customers about the pay offer and also contain the payment LNURL. See also: 

https://x.com/pete_rizzo_/status/1916212672988909908?s=46

https://cointelegraph.com/news/spar-store-zug-switzerland-bitcoin-payments-lightning-network

https://x.com/taprootfreak/status/1912183314372341765?s=46

http://app.dfx.swiss/pl?merchant=SPAR

TaprootFreak:
We have just made the changes so that arkade is compatible with OCP.

This makes it possible to pay with the Arkade Wallet in the SPAR supermarket directly from Ark to Ark. No Lightning Swap necessary.

TaprootFreak:
Funny Fact

2513_testspar-01


If you scan the QR code with the camera instead of the wallet, you can open the website and find out about the payment offer. Arkade Wallet is already listed there with a link to the GitHub repo.

Bildschirmfoto 2025-07-29 um 01 34 50

PR #106, opened 2025-05-27
Lightning swaps

bordalix:
- Reverse swaps are not implemented, only normal submarine swaps
- Documentation (and Dockerfiles) on how to test submarine swaps
- Pass default ark server url as environment variable VITE_ARK_SERVER
- Don't cache on service worker for now
- Introduces new Limits provider

## Summary by CodeRabbit

* **New Features**
  * Added Dockerfiles for Arkd, Arkd-wallet, Fulmine, and CORS proxy services with multi-stage builds and persistent data volumes.
  * Introduced a comprehensive Docker Compose setup for local regtest environment including Ark, Boltz, Fulmine, LND, and CORS services.
  * Added Lightning Network invoice decoding and Boltz swap integration with real-time WebSocket monitoring.
  * Introduced React context providers for Lightning swap monitoring and dynamic transaction limits (swap, UTXO, VTXO).
  * Enhanced wallet initialization to compute and store public keys.

* **Improvements**
  * Updated UI components to support Lightning invoices, dynamic QR codes, expanded address and invoice display, and improved user interaction.
  * Refined transaction validation and flows to incorporate Lightning swaps alongside Ark and mainnet transactions.
  * Improved type safety and network typings across wallet, explorers, and transaction types.
  * Enhanced payment success screen with detailed Lightning payment status, error handling, and timeout-based forced success.
  * Streamlined wallet state management and initialization logic.
  * Adjusted error message placement and UI text for clarity.
  * Reduced padding and container widths in QR code and address components for better layout.
  * Updated server URL defaults and explorer network mappings.
  * Simplified Ark note parsing and handling in multiple components.
  * Improved numeric type consistency for expiry and limits across components.
  * Refined market hour calculations and removed deprecated limit checks.
  * Enhanced swap and VHTLC claim logic with robust WebSocket monitoring and fallback polling.

* **Bug Fixes**
  * Fixed address validation and improved regex for BIP21 URI and Lightning invoice detection.

* **Documentation**
  * Added a detailed regtest guide for testing Ark/Lightning Network submarine and reverse swaps locally.

* **Chores**
  * Added new dependencies and configuration files including package updates and Wrangler config.
  * Commented out network-first caching strategy in the service worker, disabling fetch event handling.

PR #126, opened 2025-07-31
Swaps with package

bordalix:
Same as #106 but using the recently published boltz-swap package.
Here to generate a test deployment on cloudflare for testing.


tiero (approved):
tACK

Issue #133, opened 2025-08-04
Add Biometrics toggle in Settings


theDavidCoen:
The only way to enable Biometric login is while creating the wallet.
I suggest to add a Biometrics toggle in Settings.

theDavidCoen:
@tiero pass, tested on change_password branch.

I suggest to:

1. Move Change password in Settings, not as a sub folder
2. Separate Biometrics from Change Password and show a simple toggle. When the user switches the toggle on Active, the app shows the modal and once active the unlock wallet is made via Biometrics.
3. Add any sensitive options such as:
Change Password
Biometrics toggle
Backup
Reset Wallet
in a password protected section in Settings, since, as @bordalix correctly wrote "_you need to prove to the app you know the previous password before changing to biometrics, or else anyone could hijack you account if have access to your phone._"

bordalix:
Reset wallet cannot be behind a password firewall since if the user forgets his password, it's the only way he has to start over:
- reset wallet
- restore with nsec
- define a new password

PR #134, opened 2025-08-05
Change password

### LINKED ISSUE #132: Add Change Password in Settings
Feature request for being able to change the password later.

bordalix:
Closes #132 and #133

@theDavidCohen please test and review


tiero:
I thin biometrics should be a dedicated row in settings. Or you did it this way to preserve the "either passowed or biomtric loging, not both"  ? I guess we can do an iteration with design after riga to be more consistent UX

PR #310, opened 2026-01-21
fix sending notification without permission

bordalix:
Not tested (will test with preview deployment on cloudflare)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification system reliability by adding proper permission verification. The app now explicitly requests notification permissions before displaying service worker notifications, preventing delivery failures and ensuring consistent notification display.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/lib/notifications.ts
+++ b/src/lib/notifications.ts
@@ -13,8 +13,12 @@ export const sendNotification = (title: string, body: string) => {
     new Notification(title, options)
   } catch {
     try {
-      navigator.serviceWorker.ready.then((registration) => {
-        registration.showNotification(title, options)
+      Notification.requestPermission().then((result) => {
+        if (result === 'granted') {
+          navigator.serviceWorker.ready.then((registration) => {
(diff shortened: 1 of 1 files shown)


src/lib/notifications.ts
```diff
@@ -13,8 +13,12 @@ export const sendNotification = (title: string, body: string) => {
     new Notification(title, options)
   } catch {
     try {
-      navigator.serviceWorker.ready.then((registration) => {
-        registration.showNotification(title, options)
+      Notification.requestPermission().then((result) => {
```
Kukks: this will trigger the permission request on start? If they reject, it's gone forever?  Can we style the request with our own branding and text?

tiero: I would
Ask permission only when user go in settings and enable. Eventually e can use the popup announcement or similar after user onboarded to ask for it (if installed as PWA)

PR #320, opened 2026-02-20
Add auto theme with system preference + ripple transition

sahilc0:
## Summary

- Adds `Themes.Auto` option that follows the OS `prefers-color-scheme` setting, with a live media query listener that updates the app when the system theme changes
- Theme switches animate with a circular ripple (View Transition API) expanding from the tap point — no animation on reduced motion, unsupported browsers, or when the effective theme doesn't change
- Auto label always shows the system preference ("Auto (Light)" or "Auto (Dark)") regardless of which theme is manually selected

## Changes

**Core (`src/providers/config.tsx`)**
- Add `Themes.Auto` enum value, `resolveTheme()` helper, `effectiveTheme` and `systemTheme` state
- `applyTheme()` resolves Auto → Dark/Light via media query
- Media query listener always tracks system theme; applies it when Auto is selected

**Settings UI**
- `src/screens/Settings/Theme.tsx` — Auto/Dark/Light selector with ripple transition orchestration, uses `systemTheme` for Auto label
- `src/screens/Settings/General.tsx` — theme row shows `systemTheme` in Auto label
- `src/components/Select.tsx` — `labels` prop for display text that differs from option values

**Supporting**
- `src/lib/types.ts` — `Themes.Auto` enum
- `src/components/Announcement.tsx` — uses `effectiveTheme` instead of `config.theme`
- `src/index.css` — `::view-transition` ripple keyframe + `prefers-reduced-motion` guard

## Test plan

- [ ] App loads with Auto selected by default, resolves to system theme
- [ ] Settings > Theme shows Auto (system), Dark, Light with correct checkmark
- [ ] Tapping Dark/Light triggers ripple animation from tap point
- [ ] Auto label stays "Auto (Light)" even when Dark is selected (reflects OS, not selection)
- [ ] Tapping already-selected theme does nothing (no animation)
- [ ] Tapping Auto when it resolves to current theme — no animation
- [ ] Rapid switching between themes — transitions skip cleanly
- [ ] Reduced motion enabled — no animation, instant switch
- [ ] System theme change while Auto is selected — theme updates live
- [ ] `pnpm test` — existing theme tests pass

## Summary by CodeRabbit

* **New Features**
  * Introduced an Auto theme that follows your system light/dark preference.
  * Added a ripple transition when changing themes (respects reduced-motion settings).
  * Theme settings now show the active system theme when Auto is selected.
  * Select controls can display optional custom labels for options.

### DIFF
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -16,8 +16,8 @@ import { NavigationContext, Pages } from '../providers/navigation'
 
 // icon with pretty gradient background
 const PrettyIcon = ({ color, icon }: { color?: string; icon: React.ReactNode }) => {
-  const { config } = useContext(ConfigContext)
-  const defaultColor = config.theme === Themes.Dark ? '#ffffff' : '#000000'
+  const { effectiveTheme } = useContext(ConfigContext)
+  const defaultColor = effectiveTheme === Themes.Dark ? '#ffffff' : '#000000'
   const _color = color?.startsWith('#') ? color : defaultColor
   const circle = 'circle at 50% -70%'
   const gradient = [_color + 'dd 0%', _color + '00 70%']
--- a/src/components/Select.tsx
+++ b/src/components/Select.tsx
@@ -4,12 +4,13 @@ import FlexRow from './FlexRow'
 import Text from './Text'
 
 interface SelectProps {
+  labels?: string[]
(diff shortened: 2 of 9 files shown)


bordalix (commented):
@sahilc0 all Rabbit comments seems legit, you should address them. I will test this tomorrow.

PR #334, opened 2026-02-26
fix(ui): match iOS/Android status bar color to app theme

sahilc0:
## Summary

- Status bar on iOS (and Android) stayed white in dark mode
- Root cause: `theme-color` meta tag was hardcoded to `#000000`, Safari doesn't reliably support `media` queries on `theme-color`, and the `html` element had no background-color so the safe area behind the status bar defaulted to white
- Added an inline `<script>` in `<head>` that reads the stored theme from `localStorage` and sets both `theme-color` and `ion-palette-dark` before first paint — no white flash
- `applyTheme()` in `config.tsx` now keeps the meta tag in sync on manual theme toggle
- Set `background-color: var(--ion-background-color, #fff)` on `html, body, #root` so the safe area fills with the correct theme color

## Test plan

- [ ] iOS Safari dark mode: status bar should be dark (#101010), not white
- [ ] iOS Safari light mode: status bar should be white (#fff)
- [ ] Toggle theme in Settings (Auto/Light/Dark) — status bar updates immediately
- [ ] Android Chrome: status bar/address bar area matches theme
- [ ] Fresh load with no cached config — falls back to system preference correctly

## Summary by CodeRabbit

* **Style**
  * Ensures the correct theme color is applied before content paints for smoother dark/light transition.
  * Adds background and safe-area adjustments for more consistent visuals and a proper iOS status bar appearance.

* **Chores**
  * Updated default theme color values (light/dark) and added Apple status-bar meta for improved initial load behavior.

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
       try {
         var c = JSON.parse(localStorage.getItem('config') || '{}')
         var t = c.theme
-        var dark = t === 'dark' || (t !== 'light' && window.matchMedia('(prefers-color-scheme: dark)').matches)
+        var dark = t === 'Dark' || (t !== 'Light' && window.matchMedia('(prefers-color-scheme: dark)').matches)
         if (dark) {
           document.querySelector('meta[name="theme-color"]').setAttribute('content', '#101010')
           document.documentElement.classList.add('ion-palette-dark')
--- a/src/index.css
+++ b/src/index.css
@@ -51,7 +51,7 @@ body,
   background-color: var(--ion-background-color, #fff);
 }
 
-/* Fixed overlay for iOS 26+ status bar color sampling.
+/* Fixed overlay for iOS status bar color sampling.
    iOS derives the status bar color from the rendered content at the top of the page.
    This strip sits behind the transparent status bar and matches the theme. */
 body::before {


index.html
```diff
@@ -7,7 +7,22 @@
...
+        var c = JSON.parse(localStorage.getItem('config') || '{}')
+        var t = c.theme
+        var dark = t === 'dark' || (t !== 'light' && window.matchMedia('(prefers-color-scheme: dark)').matches)
+        if (dark) {
+          document.querySelector('meta[name="theme-color"]').setAttribute('content', '#101010')
+          document.documentElement.classList.add('ion-palette-dark')
+        }
+      } catch (e) {
+        if (location.hostname === 'localhost') console.warn('Theme bootstrap failed', e)
+      }
```
tiero: I think this should be done in React startup component? To keep all logic in the React apps vs to change in the html as it will be hard to maintain and not clear to future developers

sahilc0: This must stay as an inline `<script>` in `index.html` because it needs to run **before React mounts** — without it, dark-mode users see a white flash before the theme is applied. This is the same pattern Next.js and Remix use for the same reason (they call it a "theme bootstrap" or "color scheme script").

All runtime theme changes (user switching themes in settings) are handled in `src/providers/config.tsx` via `applyTheme()`. The inline script only handles the initial page load.

I've added comments in the HTML pointing to `config.tsx` so future developers know to keep both in sync.

PR #345, opened 2026-02-27
[Exploration] BIP-177: replace SATS with ₿ prefix format

sahilc0:
## Exploration: BIP-177 display format

This is an **exploration** of what the wallet would look like if we adopted the [BIP-177](https://github.com/bitcoin/bips/blob/master/bip-0177.mediawiki) standard, which redefines "1 bitcoin" as the base unit (formerly "1 satoshi") and uses the ₿ symbol as a prefix.

**This is not a proposal to merge** — just saving the work to revisit later.

### What changed
- `prettyAmount()` returns `₿10,000` instead of `10,000 SATS`
- `prettyHide()` returns just dots (no SATS suffix)
- `CurrencyDisplay.Sats` → `CurrencyDisplay.Bitcoin` ("Bitcoin only" in settings)
- Balance, InputAmount, Keyboard components restructured for ₿ prefix
- All user-facing strings updated (notifications, errors, settings copy)
- `.bitcoin-symbol` CSS class for ₿ font rendering (Geist doesn't cover U+20BF)
- All unit tests and E2E tests updated

### What works
- Visually verified in browser: balance, transaction list, send flow, settings
- ₿ symbol renders correctly via system sans-serif fallback
- Unit tests pass
- 29 files changed, ~124 insertions, ~97 deletions

### What's left to decide
- Whether BIP-177 is the right direction at all (controversial)
- Large number readability (e.g. ₿1,000,000,000 vs abbreviations)
- Hidden balance format (currently `₿··········`)
- Internal variable naming (left as `sats`/`amountInSats` — display-layer only change)

### DIFF
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -169,7 +169,7 @@ export function LendaSatAnnouncement({ close }: { close: () => void }) {
       title='LendaSat'
       page={Pages.AppLendasat}
       icon={<LendasatIcon big />}
-      message='Borrow against your sats.'
+      message='Borrow against your bitcoin.'
       bulletPoints={[
         [
           'Choose a loan',
--- a/src/components/Balance.tsx
+++ b/src/components/Balance.tsx
@@ -18,14 +18,15 @@ export default function Balance({ amount }: BalanceProps) {
 
   const fiatAmount = toFiat(amount)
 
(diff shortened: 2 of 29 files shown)


bordalix:
Closing due to bad taste.

sahilc0:
wowowow hey!! was just a draft!

PR #346, opened 2026-02-27
fix(pwa): fix swipe-back navigation for iOS standalone mode

sahilc0:
## Summary

- Fixes iOS edge-swipe gesture in the PWA: was navigating to wrong screens, showing white pages, and causing double animations
- Syncs browser history entries 1:1 with internal back stack — when back stack is empty, iOS disables the swipe gesture entirely (like a native app)
- Integrates Settings sub-navigation (`OptionsContext`) with browser history so swipe-back works within Settings pages
- Distinguishes swipe vs back-button taps to suppress Framer Motion animation on swipe (browser already provides the transition)
- Adds `overscroll-behavior: none` to prevent rubber-banding

## Files changed

- `src/providers/navigation.tsx` — rewrote history management (1:1 sync model, stable popstate handler via refs, sub-nav coordination)
- `src/providers/options.tsx` — Settings sub-nav now pushes browser history, registers with main nav for cleanup on tab switches
- `src/index.css` — added `overscroll-behavior: none`

## Test plan

- [ ] iOS: on root tabs (Wallet/Apps/Settings), swipe left does nothing
- [ ] iOS: forward nav (e.g. Settings → Vtxos), swipe back returns to Settings
- [ ] iOS: deep nav (Send → Details), swipe back through each step
- [ ] iOS: after completing a flow (arrive at Wallet), swipe is disabled
- [ ] iOS: back button tap still shows slide animation
- [ ] Desktop: Escape key still works, in-app back buttons work

## Summary by CodeRabbit

* **Style**
  * Disabled scroll chaining to improve page scrolling behavior.

* **Refactor**
  * Redesigned navigation handling for more reliable back/forward behavior and correct screen transitions.
  * Coordinated main and sub-navigation so the back button and nested flows behave consistently.
  * Improved settings navigation state and back-button response for smoother option-level navigation.


sahilc0:
Hey @pietro909 @bordalix — this is a workaround for the iOS PWA swipe-back issue. It syncs browser history entries 1:1 with our in-app navigation so the iOS swipe gesture works correctly (back when there's somewhere to go, disabled when there isn't).

It's all standard web APIs (pushState, popstate, history.go) so it won't conflict with a future Capacitor migration — and once we move to Capacitor, the native webview won't have Safari's swipe gesture at all, so this code just becomes inert.

Are you guys ok with this approach?

PR #329, opened 2026-02-24
feat(ui): animated page transitions, wallet stagger, keyboard overlays

sahilc0:
## Summary

- **Page transitions**: Directional slide + opacity animations for all screen navigation using Framer Motion's `custom` prop. Forward slides right-to-left, back slides left-to-right, tab switches are instant. AnimatePresence mode changed from `popLayout` to `sync` for smoother concurrent enter/exit.
- **Wallet entrance stagger**: Sequential fade-in of all wallet elements on initial load — logo, balance, buttons, "Transaction history" label, and each individual tx row. Navigation delayed from Unlock until data is fetched, eliminating skeleton→data flash.
- **Keyboard/scanner overlays**: Send and Receive numpad/scanner slide up from bottom instead of replacing the page.
- **Header simplification**: `back` prop accepts `true` (uses `goBack()`) instead of explicit navigate callbacks. 10+ screens simplified.
- **Accessibility**: All animations respect `prefers-reduced-motion` via `useReducedMotion` hook.

## Files changed (30 files)

**Core animation system**
- `src/lib/animations.ts` — Unified `pageTransitionVariants` with dynamic direction, `keyboardOverlay`, wallet stagger variants
- `src/components/PageTransition.tsx` — New page transition wrapper
- `src/components/SettingsPageTransition.tsx` — Settings tab equivalent
- `src/components/WalletLoadIn.tsx` — Stagger container/child with `initial={false}` bypass
- `src/hooks/useReducedMotion.ts` — Reduced motion detection hook
- `src/index.css` — Transition container, overflow containment

**Navigation + data loading**
- `src/providers/navigation.tsx` — Direction state, `goBack()`, `isInitialLoad` flag
- `src/providers/wallet.tsx` — `dataReady` flag, delays navigation until data loaded
- `src/providers/options.tsx` — Settings navigation direction

**Screens updated**
- `src/App.tsx` — AnimatePresence wrapping each tab
- `src/screens/Wallet/Index.tsx` — Stagger all elements, per-row tx stagger
- `src/screens/Wallet/Unlock.tsx` — Wait for dataReady before navigating
- `src/screens/Wallet/Send/Form.tsx` — Keyboard + scanner overlay animations
- `src/screens/Wallet/Receive/Amount.tsx` — Keyboard overlay animation
- `src/components/Header.tsx` — Boolean `back` prop support
- 10+ screens simplified to `back={true}`

## Test plan

- [ ] Unlock wallet → Loading stays visible until data loads, then page transitions in
[...]

PR #354, opened 2026-03-02
perf(transactions): virtualise TransactionsList with @tanstack/react-virtual

tiero:
### Impact

Wallets with several thousand transactions experienced severe lag or a complete
UI freeze when opening the Wallet tab.

### Timeline

| Date | Commit | Author | Event |
|------|--------|--------|-------|
| Feb 25 | `5648c33` | @sahilchaturvedi | PR #329 added Framer Motion stagger animations wrapping **every** tx row via an `ItemWrapper` prop. With thousands of rows each got its own animation context → hard freeze. |
| Mar 2 AM | `b6d801e` | @tiero | PR #350 removed the per-item `ItemWrapper`. Animation overhead gone, but `txs.map()` still mounted every DOM node unconditionally — root cause untouched. |
| Mar 2 | `81ac79a` | this PR | Virtualised the list. Fixed properly. |

### Root cause

`TransactionsList` rendered every transaction into the DOM via a plain
`.map()` — every row, `Focusable` wrapper, icon, and fiat conversion mounted
at once. Render time was **O(n)**:

```
Before (all DOM nodes mounted):
  100 txs  →    471 ms
  500 txs  →  1 475 ms
 1000 txs  →  2 258 ms
 3000 txs  →  5 477 ms
 5000 txs  →  9 206 ms
```

PR #329's animation was the trigger that surfaced it, but the list was
never built to handle thousands of rows.

### Fix

Replaced `txs.map()` with `@tanstack/react-virtual`'s `useVirtualizer`.
Only the rows visible in the scroll container (+ 5-row overscan) are ever
mounted. DOM node count is now **O(visible) ≈ 15–25**, constant regardless
of wallet history size.

```
After (virtual rows only):
  100 txs  →  4 ms ✓
  500 txs  →  3 ms ✓
 1000 txs  →  4 ms ✓
 3000 txs  →  4 ms ✓
 5000 txs  →  3 ms ✓
```

The list gets its own scroll container (`height: calc(100dvh - 380px)`),
keeping the balance and action buttons pinned above it — the standard
pattern in mobile wallet UIs.

### Regression gate (smoke test)

`src/test/screens/wallet/transactions-smoke.test.tsx` runs in `pnpm test:unit`:

- **DOM-count assertion** — at most 20 rows mounted at once; virtualisation
  cannot regress silently.
- **Render-time assertions** — initial mount ≤ 500 ms for any tx count.

### Test plan

- [x] `pnpm test:unit` — 31 files, 133 tests, all green
- [ ] Manual: wallet with 100+ txs scrolls smoothly
- [ ] Manual: balance and action buttons stay pinned while list scrolls

## Summary by CodeRabbit

* **Performance Improvements**
  * Virtualized transaction list for much faster rendering and smoother scrolling with large histories.
* **New Features**
  * Keyboard navigation and improved focus handling for navigating transaction rows.
  * Reduced DOM footprint by rendering only visible rows.
* **Tests**
  * Added smoke/performance tests validating virtualization behavior and measuring render times for large transaction sets.


tiero:
@Kukks can you test this as well so we can improve big wallets rendering?

PR #419, opened 2026-03-13
fix bug where was not possible to navigate to password in lock wallet

bordalix:
@pietro909 please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed automatic navigation after setting a password in the Lock settings screen, allowing users to remain on the configuration page.

* **Chores**
  * Updated internal version identifier.


pietro909 (approved):
Tested 👍

PR #421, opened 2026-03-13
Prevent lockup state

pietro909:
- Separated wallet auth state from backend initialization so passwordless wallets no longer fall into Unlock when startup fails. Unlock is now shown only when credentials are actually required; passwordless wallets auto-boot in the background and stay on Loading if the backend is still unavailable.
- Also added a one-shot recovery path: if passwordless auto-init fails, the app logs the error and triggers a single automatic reload after 1 second, guarded per session to avoid loops. 
- Added regression tests for locked vs passwordless startup routing and the single-reload behavior.

## Summary by CodeRabbit

* **New Features**
  * Passwordless auto-initialization with one-time retry and background unlocking.
  * Explicit unlock flow with improved unlocking feedback.

* **Bug Fixes**
  * Prevents repeated passwordless retries and navigation loops.
  * More reliable startup routing between Loading, Unlock, and app screens.

* **Refactor**
  * Wallet auth state and unlock API added to wallet context; NeedsPassword accepts async handlers.

* **Tests**
  * Added comprehensive startup routing tests covering passwordless, unlock, and reload behaviors.


Kukks (approved):

pietro909:
@bordalix I think this PR is still relevant after https://github.com/arkade-os/wallet/pull/431 because it decouples the error state from the Unlock screen, where a user may end up in a dead end with a reset as the only viable solution.

PR #493, opened 2026-03-30
Fix boot animation overlay blocking the Unlock page

pietro909:
## Summary
- The boot animation overlay (opaque white background at z-index 9) was kept alive when `page` transitioned to `Pages.Unlock`, making the password input completely invisible and the app appear permanently stuck
- For passwordless wallets `page` goes `Loading → Wallet` (never `Unlock`), so the exclusion was dead code — except when the passwordless auto-boot fails (e.g. Vivaldi service worker timeout sets `authState` to `'locked'`), where it caused a deadlock
- Dismiss the overlay for any non-Loading page, including Unlock
- Fix two pre-existing broken unit tests (from #489) that looked for removed `'Loading...'` text

## Test plan
- [x] All unit tests pass (`pnpm test:unit` — 164 passed, 2 skipped)
- [x] Lint clean (`pnpm lint`)
- [x] Manual test: locked wallet boots to visible Unlock page
- [x] Manual test: passwordless wallet still gets fly-to-target animation (Loading → Wallet)
- [x] Manual test in Vivaldi: app no longer stuck on loading screen

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated boot animation exit behavior to correctly dismiss the overlay when navigating through certain app states during startup.

* **Tests**
  * Updated initial render assertions to verify correct startup state content.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -263,8 +263,8 @@ export default function App() {
         ? Pages.Unlock
         : screen
 
-  // Boot animation: persists from Loading through Unlock until Wallet is reached,
-  // then flies to the LogoIcon position. For new users (→ Init), exits with fly-up.
+  // Boot animation: persists on Loading, then flies to the LogoIcon position when
+  // Wallet is reached. For any other destination (Unlock, Init, etc.), exits with fly-up.
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


bordalix (approved):
tACK

PR #533, opened 2026-04-15
feat(receive): tap QR code to copy what it encodes

sahilc0:
## Summary

On the Receive screen, tapping the QR code now copies whatever value it currently encodes to the clipboard — same fast-path the bottom **Copy** button previously required two clicks for.

- **Default:** unified BIP21 URI (bitcoin + ark + lightning + amount) is copied
- **After picking a sub-address in the Copy sheet:** the QR and tap-to-copy both switch to that sub-address (Lightning invoice, Arkade address, Bitcoin address, or LNURL)
- **Tap feedback:** subtle Emil-style scale-down on press + existing `hapticSubtle()` tick (both suppressed when the user prefers reduced motion), existing "Copied to clipboard" toast
- **Copy sheet unchanged:** still available for the multi-address picker

## Why

Tapping the graphic a user is already focused on is the obvious path; forcing them into a sheet to copy the same string the QR already shows is friction.

## What changed

**`src/screens/Wallet/Receive/QrCode.tsx`**
- Wrapped `<QrCode value={qrCodeValue} />` at L404 in a real `<button aria-label="Copy QR code">` — keyboard-focusable, screen-reader-announced, button-reset styled so native chrome doesn't leak
- Pointer handlers use direct `e.currentTarget.style.transform` mutation (no React state, no re-renders) matching the existing `AssetAvatar` press pattern
- `EASE_OUT_QUINT` from `src/lib/animations.ts` for the transition curve
- Gated `hapticSubtle()` inside `handleCopy()` on `useReducedMotion()` so both the new QR tap and the existing Copy-sheet rows skip haptics under reduced motion (the rest of the app's haptics are a follow-up)

**`src/test/screens/wallet/receive-qrcode.test.tsx`**
- Added 3 RTL tests: happy-path tap-to-copy, toast visibility, and a rerender regression that exercises the same `setQrCodeValue` code path the Copy sheet triggers (the sheet itself can't be driven in jsdom — `IonModal` portals children outside the React root)
- Skipped 2 pre-existing loader tests that were already failing on `master` for unrelated React 18 / RTL effect-flushing reasons (documented in comments)

## Test plan

- [x] `bun run test -- receive-qrcode` — 6 passed, 2 skipped, 0 failed
- [x] `bun run test:unit` — 170 passed, 2 skipped, 2 failed (the 2 skipped-above)
- [x] `bun run lint` — clean
- [x] `bun run format:check` — clean
- [x] Manual — iOS Safari via cloudflared tunnel: tapped QR with amount / without amount / after switching to Arkade address; verified scale animation, haptic, toast, and clipboard contents

## Out of scope

- Centralizing reduced-motion haptic gating inside `src/lib/haptics.ts` (would touch all screens; follow-up)
- Fixing the two pre-existing broken loader tests (unrelated to this PR)

## Summary by CodeRabbit

* **New Features**
  * QR code is now tappable to copy the receive address
  * Toast notification confirms successful copy action
  * Added visual press feedback when tapping QR code

* **Accessibility**
  * QR code interactions respect system reduced-motion preferences
  * Haptic feedback is disabled when reduced motion is enabled

### DIFF
--- a/src/screens/Wallet/Receive/QrCode.tsx
+++ b/src/screens/Wallet/Receive/QrCode.tsx
@@ -59,6 +59,7 @@ export default function ReceiveQRCode() {
 
   const [sharing, setSharing] = useState(false)
   const [addressesLoaded, setAddressesLoaded] = useState(false)
+  const [qrTransform, setQrTransform] = useState('')
 
   // Amount sheet state
   const [showAmountSheet, setShowAmountSheet] = useState(false)
@@ -408,12 +409,10 @@ export default function ReceiveQRCode() {
                   <button
                     type='button'
                     onClick={() => handleCopy(qrCodeValue)}
-                    onPointerDown={(e) => {
-                      if (!prefersReducedMotion) e.currentTarget.style.transform = 'scale(0.97)'
-                    }}
-                    onPointerUp={(e) => (e.currentTarget.style.transform = '')}
-                    onPointerLeave={(e) => (e.currentTarget.style.transform = '')}
-                    onPointerCancel={(e) => (e.currentTarget.style.transform = '')}
+                    onPointerDown={() => setQrTransform(prefersReducedMotion ? '' : 'scale(0.97)')}
+                    onPointerUp={() => setQrTransform('')}
+                    onPointerLeave={() => setQrTransform('')}
+                    onPointerCancel={() => setQrTransform('')}
                     aria-label='Copy QR code'
                     style={{
                       background: 'none',
@@ -429,6 +428,7 @@ export default function ReceiveQRCode() {
                         : `transform 240ms cubic-bezier(${EASE_OUT_QUINT.join(',')})`,
                       WebkitTapHighlightColor: 'transparent',
                       touchAction: 'manipulation',
+                      transform: qrTransform,
                     }}
                   >
                     <QrCode value={qrCodeValue} />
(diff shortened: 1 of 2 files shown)


src/screens/Wallet/Receive/QrCode.tsx
```diff
@@ -401,7 +405,34 @@ export default function ReceiveQRCode() {
...
-                  <QrCode value={qrCodeValue} />
+                  <button
+                    type='button'
+                    onClick={() => handleCopy(qrCodeValue)}
+                    onPointerDown={(e) => {
+                      if (!prefersReducedMotion) e.currentTarget.style.transform = 'scale(0.97)'
+                    }}
+                    onPointerUp={(e) => (e.currentTarget.style.transform = '')}
+                    onPointerLeave={(e) => (e.currentTarget.style.transform = '')}
+                    onPointerCancel={(e) => (e.currentTarget.style.transform = '')}
```
pietro909: mutating the event target's style sounds like a using a hammer to flip a book's pages 😄 
Any chances we can track a state instead?

Something along the lines of this (not a working example, just a hint):
```
const [qrBtnTransform, setQrBtnTransform] = useState( /* what's the initial state btw? */ )

// more code
onPointerCancel={() => setQrBtnTransform('')}
style={{ transform: qrBtnTransform }}
```

PR #535, opened 2026-04-16
feat(wallet): prefix fiat amounts with currency symbols

sahilc0:
## Summary

Fiat amounts previously rendered with a trailing ISO code (`100.00 USD`, `50.00 EUR`). They now use the common Unicode currency symbol as a prefix — `$100.00`, `€50.00`, `£25.00`, `¥1,000`. Currencies without a widely-used single-char symbol (CHF, CNY) keep the trailing-code form, so CNY does not collide with JPY's `¥`.

### New helpers in `src/lib/format.ts`
- **`prettyFiatAmount(amount, currency)`** and **`prettyFiatHide(value, currency)`** — take the `Fiats` enum as a typed parameter so only real fiat flows through.
- Decimals are derived from the currency via **`fiatDecimalsFor`** in `src/lib/fiat.ts` (JPY = 0, else 2) — shared between `FiatContext` and `prettyFiatAmount`, so call sites no longer need to pass `fiatDecimals()`.
- The generic `prettyAmount` / `prettyHide` are intentionally left **untouched** so asset tickers (e.g. a stablecoin named `USD`) that flow through those helpers keep their trailing-code format.

### Affected displays
- Hero balance (`Balance.tsx`)
- Transaction list fiat line (`TransactionsList.tsx`)
- Amount picker keyboard (`Keyboard.tsx`)
- Transaction details screen (`Details.tsx`)
- Send / Receive / Notes success confirmations
- Send form "available" caption (`Send/Form.tsx`)
- Send/Receive amount input labels (`InputAmount.tsx`) — left label shows `$` instead of `USD`, right-side fiat conversion is also prefixed

### Pre-existing bug fixed along the way
`Details.tsx` was calling `prettyAmount` without passing `fiatDecimals()`, hardcoding 2 decimals. JPY now correctly renders as `¥174` / `¥0` in the transaction details instead of `174.00 JPY` / `0.00 JPY`.

### Hero eye-toggle alignment
The eye button wrapper omits an empty unit `<Text>` when the symbol is inlined into the balance, and stays inside an `alignItems: center` flex container next to the unit text — keeping the icon centered with the smaller ISO code in the CHF case (`5.70 CHF 👁`) and snug against the balance in the symbol case (`$7.30 👁`).

## Settings screens are intentionally unchanged
`Settings/General.tsx` and the `Settings/Fiat.tsx` picker continue to display the ISO code as an identifier — those render a setting value, not an amount.

## Visual verification

Dev server run against all six supported fiats:

| Currency | Hero | Transactions list | Transaction details | Send input |
|---|---|---|---|---|
| USD | `$7.30` | `$0.75`, `$11.28` | `$1.09` / `$0.75` | `$` label, `$7.07 available` |
| EUR | `€6.20` | `€0.64`, `€9.59` | `€0.93` | `€` label |
| GBP | covered via unit tests | — | — | — |
| JPY | `¥1,163` (0 decimals) | `¥119`, `¥1,798` | `¥174`, `¥0`, `¥174` (Details fix) | `¥` label, `¥1,123 available` |
| CHF | `5.70 CHF` (trailing) | `0.59 CHF`, `8.84 CHF` | `0.85 CHF` | `CHF` label |
| CNY | covered via unit tests | — | — | — |

Sats-only paths were spot-checked to confirm no regressions (delegate fees, VTXOs screen, Boltz swap list, asset rows with custom tickers).

## Test plan
- [x] `bun run test:unit` — new coverage for `prettyFiatAmount` / `prettyFiatHide` passes; existing `prettyAmount` / `prettyHide` tests untouched. Two pre-existing failures in `receive-qrcode.test.tsx` are unrelated and reproduce on `master`.
- [x] Component test `src/test/screens/wallet/index.test.tsx` updated to assert the new `$0.00` rendering.
- [x] E2E assertions in `src/test/e2e/init.test.ts` and `src/test/e2e/keyboard.test.ts` updated to match the symbol-prefix format.
- [ ] Full Playwright E2E run in CI.
- [x] Manual sweep across USD, EUR, JPY, CHF on both `Fiat only` and `Show both` display modes, with `showBalance` toggled.

## Summary by CodeRabbit

* **New Features**
  * Fiat amounts now show currency symbols when available and suppress redundant fiat unit labels; JPY uses 0 decimals, others use 2.
  * Improved masked/hidden balance rendering to respect presence/absence of currency symbols.

* **Bug Fixes**
  * Consistent fiat formatting applied across screens and input/keyboard flows.

* **Tests**
  * Updated UI and formatting tests to expect symbol-based fiat displays.


sahilc0:
@coderabbitai re: Balance.tsx nitpick — deliberately not using the helpers there. Balance renders `mainBalance` and `mainUnit` as **separately-styled** text elements: the balance in `bigger heading medium` and the unit in `heading` (smaller, baseline-aligned). `prettyFiatAmount` returns a combined string like `"5.70 CHF"`, which would put "CHF" in the big-heading-medium style and lose the size differentiation. The manual `fiatBalanceRaw` → `fiatBalance` + `fiatUnit` split is what preserves that dual rendering.

The symbol-prefix logic here is only 1 line (`fiatSymbol && fiatBalanceRaw ? \`\${fiatSymbol}\${fiatBalanceRaw}\` : fiatBalanceRaw`), and the comment above it explains the zero-hidden guard. Not worth consolidating if it forces a visual regression on the hero.

PR #537, opened 2026-04-16
fix(ui): top-align tx history rows when assets are present

Kukks:
## Summary
- Fixes vertical alignment in transaction history rows when assets (e.g. "4 SCAM") are displayed alongside the BTC amount and fiat value
- The right column (sats + fiat + asset info) was taller than the left column, and `alignItems: 'center'` pushed the sats amount above the "Received" label
- Switches to `alignItems: 'start'` so both columns begin at the same top position
- Adds explicit `between` prop to the outer `FlexRow` (replacing implicit width-based spacing)
- Removes dead `alignItems: 'center'` on the non-flex wrapper div

## Test plan
- [ ] Open a wallet that has received asset transfers (e.g. SCAM token)
- [ ] Verify "Received" label and "+ X SATS" are top-aligned in the transaction row
- [ ] Verify rows without assets still look correct
- [ ] Verify the asset avatar icon and ticker remain visible and right-aligned

## Summary by CodeRabbit

* **Style**
  * Adjusted transaction list row layout to refine content alignment. Right-side elements now align to the top of each row for improved visual consistency.

### DIFF
--- a/src/components/TransactionsList.tsx
+++ b/src/components/TransactionsList.tsx
@@ -35,6 +35,7 @@ const TransactionLine = ({ tx, onClick }: { tx: Tx; onClick: () => void }) => {
   const prefix = tx.type === 'sent' ? '-' : '+'
   const amount = `${prefix} ${config.showBalance ? prettyAmount(tx.amount) : prettyHide(tx.amount)}`
   const date = tx.createdAt ? prettyDate(tx.createdAt) : tx.boardingTxid ? 'Unconfirmed' : 'Unknown'
+  const asAssets = Boolean(tx.assets?.length)
   const issuance = isIssuance(tx)
   const burn = isBurn(tx)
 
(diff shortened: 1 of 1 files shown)


bordalix:
ACK.

Nit: we still have an overflow when Display is set to show both and we have asset txs (see screenshot).

Suggestions:

1. All asset txs should have the asset line on top (with bigger font) and the sats line in second (with smaller font). The amount of assets is the main character here, the 330 sats is just a "need to know".

2. We should only show 2 lines on the right, so in asset txs, the logic should be:
- Display sats or display both => show sats
- Display fiat => show fiat for 330 sats

@Kukks @sahilc0 what do you think?

I can do that changes.

Screenshot 2026-04-17 at 09 52 06

bordalix:
Something like this:

Screenshot 2026-04-17 at 10 19 07

PR #546, opened 2026-04-20
reverse logs in csv export

bordalix:
NIT: reverse logs in CSV export so newer logs appear on top.

Useful for when Slack shows a preview of the CSV file in support channel.

@tiero wdyt?

## Summary by CodeRabbit

* **Bug Fixes**
  * CSV log export now produces rows in reverse chronological order so exported files list recent entries first (header and column values unchanged).


tiero:
ACK

PR #550, opened 2026-04-21
show only 2 coins in the rigth side of a tx in txs list

bordalix:
Implements https://github.com/arkade-os/wallet/pull/537#issuecomment-4266788973

@Kukks please review


pietro909 (approved):
Image

IT WORKS!

PR #559, opened 2026-04-22
fix: extract LNURL session into app-level provider with reusable sessions

Kukks:
## Summary
- Introduces `LnurlProvider` that manages the LNURL session lifecycle at the wallet level — the session starts automatically when the wallet and Boltz swaps are ready
- The LNURL is always active while the wallet is open (if Boltz is enabled), not just on the receive screen
- **Reusable LNURL sessions**: derives a deterministic token using HMAC-SHA256 with the wallet's private key, sends it to the server which derives the sessionId — same wallet always gets the same LNURL
- Invoice requests are handled entirely in the provider: create reverse swap, return invoice, claim in background, notify on completion
- Fixes Windows lint script (single-quoted globs don't expand on Windows)

## Problem
Three issues fixed:

1. **Session lifecycle**: The LNURL session was managed inside a `useEffect` on the receive screen. When the user navigated away, the SSE connection was killed and the payer's LNURL became invalid.

2. **Non-deterministic LNURL**: Every session got a random ID, so the LNURL changed every time. Payers couldn't reuse a wallet's LNURL across sessions.

3. **Session only active on receive screen**: The LNURL session only started when the user was on the amountless receive screen. Now it's always active when the wallet is open.

## Architecture
```
LnurlProvider (app-level, in index.tsx)
  └─ consumes WalletContext, SwapsContext, NotificationsContext
  └─ starts session when identity + Boltz ready, stops when they aren't
  └─ handles invoice requests: createReverseSwap → return invoice → claim
      │
LnurlContext (consumed by any component)
  └─ exposes: lnurl, active, error (read-only)
      │
QrCode.tsx (receive screen)
  └─ reads lnurl/active from context to display QR
  └─ existing VTXO_UPDATE listener handles navigation to success screen
```

### Credential derivation
`deriveLnurlCredentials(privateKeyHex)` in `src/lib/lnurl.ts`:
- `token` = hex-encoded `HMAC-SHA256(privateKey, "lnurl-session")` — deterministic by construction
- Wallet sends only `{ token }` to the server
- Server derives `sessionId = SHA-256(tokenBytes).slice(0, 32)` — eliminates mismatched pairs and makes session hijack impossible by construction
- Uses `@noble/hashes` (hmac + sha256), synchronous

## Companion PR
- ArkLabsHQ/lnurl-server#2 — server-side support for client-provided tokens with server-derived sessionId

## Test plan
- [ ] Open wallet — LNURL session starts automatically (check network tab for SSE stream)
- [ ] Navigate to receive screen — LNURL QR is displayed
- [ ] Navigate away and back — same LNURL, session still active
- [ ] Close and reopen wallet — same LNURL appears (deterministic credentials)
- [ ] Have a payer send to LNURL while NOT on receive screen — swap is created, payment completes, notification shown
- [ ] Have a payer send while ON receive screen — same as above, plus navigates to success
- [ ] Disable Boltz / disconnect — LNURL session stops
- [ ] Lock wallet — LNURL session stops
- [ ] Unit tests pass

## Summary by CodeRabbit

* **New Features**
  * Introduced a centralized LNURL session management provider for improved state handling
  * Added LNURL credential derivation utilities

* **Refactor**
  * Restructured LNURL session handling to use a provider-based architecture
  * Updated QR code component to leverage centralized session management instead of local state

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/wallet/pull/559)


Kukks:
@pietro909  maybe we should consider moving this code to the swaps package and also offering within service worker?

PR #566, opened 2026-04-23
fix: improve timeout error messages and increase service worker timeout

Kukks:
## Summary
- Increases MessageBus timeout from 60s → 120s — settlement/send can legitimately take over a minute waiting for ASP round participation
- Adds `friendlyError()` utility that detects timeout patterns and translates them to actionable, user-facing messages
- Applied in Transaction (settle boarding), Send/Details (payments), and Settings/Vtxos (rollover) screens
- Exports `isTimeoutError()` for callers that need to branch on timeout vs other failures

## Problem
Users see `"Settlement failed: Error: Message handler timed out after 60000ms (WALLET_UPDATER)"` when clicking "Complete boarding". This is:
1. **Cryptic** — `WALLET_UPDATER` is the handler tag, not the operation (SETTLE). The SDK passes the tag to `withTimeout()` instead of the message type.
2. **Misleading** — the SDK's `withTimeout()` does NOT cancel the underlying work. The settlement may actually succeed in the background, but the user sees "failed".
3. **Too aggressive** — 60s is tight for settlement which requires round participation and signing.

## Before / After
**Before:** `Settlement failed: Error: Message handler timed out after 60000ms (WALLET_UPDATER)`
**After:** `Settlement is taking longer than expected. It may still complete in the background — check back in a moment.`

## SDK root cause
Filed https://github.com/arkade-os/ts-sdk/issues/448 for the underlying issues:
1. Timeout error includes handler tag (`WALLET_UPDATER`) instead of message type (`SETTLE`)
2. No per-operation timeout configuration
3. `withTimeout` doesn't cancel underlying work but implies failure

## Test plan
- [ ] Trigger a boarding settlement that takes >60s — should no longer fail (120s timeout)
- [ ] If timeout still occurs at 120s, user sees friendly message instead of cryptic error
- [ ] Send payment timeout shows "Payment is taking longer than expected..."
- [ ] Vtxos rollover timeout shows "Settlement is taking longer than expected..."
- [ ] Non-timeout errors display normally (unchanged behavior)
- [ ] All 177 unit tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "test:e2e": "VITE_NOSTR_RELAY_URL=http://localhost:10547 playwright test",
     "test:e2e:ui": "VITE_NOSTR_RELAY_URL=http://localhost:10547 playwright test --ui",
     "test:codegen": "playwright codegen localhost:3002",
-    "lint": "eslint 'src/**/*.{ts,tsx,js,jsx}'",
+    "lint": "eslint \"src/**/*.{ts,tsx,js,jsx}\"",
     "lint:fix": "eslint 'src/**/*.{ts,tsx,js,jsx}' --fix",
     "format": "prettier --write src",
     "format:check": "prettier --check src",
--- a/src/lib/error.ts
+++ b/src/lib/error.ts
@@ -1,3 +1,5 @@
+const TIMEOUT_PATTERN = /timed out after \d+ms/i
+
 export const extractError = (error: any): string => {
   if (typeof error === 'string') return error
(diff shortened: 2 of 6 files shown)


pietro909:
Better UX although shouldn't be needed since https://github.com/arkade-os/ts-sdk/pull/446 exempted the long-running operations.
The risk of treating all timeouts as "failed for now but could succeed soon" is that we may actually miss legit timeouts. But timeouts are an artificial construct in this case. 

Good to go from my side @Kukks

pietro909 (approved):

PR #581, opened 2026-04-29
Add bulk recovery button in Apps -> Boltz -> Settings

### LINKED ISSUE #577: Claimed swaps with multiple VTXOs aren't refunded
## Scenario

1. a Swap was funded twice (due to a transient failure)
2. the receiver claims successfully, spending the first VTXO
3. the sender expects a refund to kick in
4. the Swap status now is `transaction.claimed` - meaning the swap isn't monitored anymore, and even on restore/startup isn't taken into account for refund

## Challenges for SDK-only solution

- Blanket sweep of all swaps on startup or restore can become too heavy even batching calls to the indexer
- Capping the list of candidates requires data model change on DB, hence migration (for instance: timestamp, flag, ...)

## Proposed solution 

- Show a "Refund all swaps" in Apps -> Boltz -> Settings
- Expose relevant methods from SDK (boltz-swap) like `scanStrandedSubmarineSwaps()` and `recoverStrandedSubmarineSwaps()` for bulk scan and refund

The action is still heavy but is now user-driven: it is meant to be used when things go very wrong.

pietro909:
Closes #577 

Depends on https://github.com/arkade-os/boltz-swap/pull/133

## Summary by CodeRabbit

* **New Features**
  * Added "Recover stranded funds" functionality in Boltz settings enabling users to scan for recoverable abandoned swaps and individually initiate recovery operations with status tracking.

* **Chores**
  * Updated the Boltz swap dependency to version 0.3.24.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "homepage": ".",
   "dependencies": {
-    "@arkade-os/boltz-swap": "0.3.23",
+    "@arkade-os/boltz-swap": "0.3.24",
     "@arkade-os/sdk": "0.4.22",
     "@branta-ops/branta": "0.0.9",
     "@lendasat/lendasat-wallet-bridge": "^0.0.90",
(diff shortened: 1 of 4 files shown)


tiero:
We want the affected user to try the preview link? @pietro909

PR #596, opened 2026-05-02
fix: prioritize VITE_ARK_SERVER env var over cached server URL

### LINKED ISSUE #594: Env var VITE_ARK_SERVER silently overridden by cached localStorage config
when `VITE_ARK_SERVER` is set (via .env.local or inline), the app still connects to the previously saved server URL because `readConfigFromStorage()` in ConfigProvider spreads the stored config over the defaults, making the env var have no effect unless localStorage is manually cleared. 

`VITE_ARK_SERVER` should always take precedence over any cached value.

Dunsin-cyber:
fixes #594

Previously, a server URL saved from a previous session in localStorage would silently override VITE_ARK_SERVER, making the env var have no effect without manually clearing localStorage.                     
                                                                                                                                                                                                        
Now the env var always wins at runtime, is never written to localStorage (keeping deployment config and user preferences in separate layers), and the Server settings screen shows a clear message when the URL is env-controlled instead of silently ignoring user input.

<img width="749" height="502" alt="env priority fix" src="https://github.com/user-attachments/assets/96c6d73d-65bf-43ec-9d99-2675e7bad726" />



<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Added environment variable support for server configuration, allowing administrators to set a fixed server URL that cannot be changed by users.
  * When environment-configured, the server settings screen displays a read-only server URL with a warning message.

* **Chores**
  * Updated ESLint configuration to ignore unused variables starting with underscore.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,7 +25,7 @@
     "no-duplicate-imports": "error",
     "array-callback-return": ["error", { "allowImplicit": true }],
     "@typescript-eslint/no-explicit-any": ["off", { "ignoreRestArgs": false }],
-    "@typescript-eslint/no-unused-vars": ["warn", {}],
+    "@typescript-eslint/no-unused-vars": ["warn", { "varsIgnorePattern": "^_" }],
     "react/hook-use-state": ["error"],
     "react/no-array-index-key": ["error"],
     "react/jsx-boolean-value": ["error"],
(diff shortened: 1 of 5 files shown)


tiero:
I think this was an explicit decision, to favor user preference and use the ENV VARs as the "overrdibale" defaults for new users. What's use case have in mind beside development?

bordalix:
@Dunsin-cyber thank you for your contribution, but this is a concept NACK:
- users should be able to "jump" between servers using the same wallet domain;
- the last used server used by the user should persist in the next session.

Dunsin-cyber:
> I think this was an explicit decision, to favor user preference and use the ENV VARs as the "overrdibale" defaults for new users. What's use case have in mind beside development?

my perspective was purely from a development environment standpoint... setting `VITE_ARK_SERVER` locally but not seeing it take effect felt like a bug. 

re-reading the code with your context, the original behavior is intentional

Dunsin-cyber:
> @Dunsin-cyber thank you for your contribution, but this is a concept NACK:
> 
> * users should be able to "jump" between servers using the same wallet domain;
> * the last used server used by the user should persist in the next session.

totally fair, thanks for the clear explanation. i misread the silent override as a bug.

closing this now.

PR #602, opened 2026-05-04
fix(format): show fiat amounts with two decimals

sahilc0:
## Summary
- Update `prettyFiatAmount` to render every supported fiat currency with exactly two decimal places.
- Remove the JPY-specific zero-decimal display behavior from wallet fiat formatting.
- Add regression coverage for whole and fractional values across EUR, USD, CHF, JPY, GBP, and CNY.

## Testing
- `bun run test:unit src/test/lib/format.test.ts`
- `bun run lint`
- Pre-commit `bun run format:check`

## Affected files
- `src/lib/format.ts`
- `src/test/lib/format.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized fiat currency formatting to consistently display two decimal places across all currencies (e.g., $1.00, €2.00, ¥100.00) instead of varying by currency type.

### DIFF
--- a/src/lib/format.ts
+++ b/src/lib/format.ts
@@ -1,5 +1,5 @@
 import { centsToUnits } from './assets'
-import { fiatDecimalsFor, FIAT_SYMBOLS } from './fiat'
+import { FIAT_SYMBOLS } from './fiat'
 import { Fiats, Satoshis, Tx } from './types'
 import { Decimal } from 'decimal.js'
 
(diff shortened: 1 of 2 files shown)


bordalix:
NACK, JPY has no decimals.

sahilc0:
@coderabbitai review

sahilc0:
Thanks, you are right. I checked the currency minor-unit behavior and JPY is a zero-decimal currency, so forcing `.00` there was an overcorrection.

Updated the PR to use each fiat currency's minor-unit count as both the min and max fraction digits. That keeps USD/EUR/GBP/CHF/CNY fixed to two decimals (`$0.80`, `$10.00`) while keeping JPY zero-decimal (`¥10`, `¥11` after rounding). Also updated the formatter tests to cover whole and fractional values for every supported fiat.

PR #609, opened 2026-05-07
Boltz swap : offline receive

louisinger:


src/providers/swaps.tsx
```diff
@@ -31,6 +31,13 @@ const BASE_URLS: Record<Network, string | null> = {
   testnet: null,
 }
 
+const BANCOD_URL = import.meta.env.VITE_BANCOD_URL
```
tiero, line 34: what is this used for?

src/lib/logs.ts
```diff
@@ -47,7 +47,7 @@ export const consoleLog = (...args: any[]) => {
 export const consoleError = (err: any, msg = '') => {
   const str = (msg ? `${msg}: ` : '') + extractError(err)
   addLog(LogLevel.Error, [str])
-  console.error(str)
+  console.error(str, err)
```
tiero, line 50: seems better to have dedciate PR for this?

louisinger:
replaced by https://github.com/arkade-os/wallet/pull/750

PR #618, opened 2026-05-14
Add contracts view

pietro909:
Triple-tapping the loading logo toggles a global dev mode persisted in localStorage (DevModeProvider). The tap logic is lifted out of LoadingLogo into the new context so any loading logo in the app shares the same state.

A new "Contracts" entry appears in the Advanced settings list only when dev mode is active. It renders all contracts from ContractManager sorted active-first, each card showing type, state, address, and script — the latter two shortened and copyable. Pull-to-refresh is disabled on this screen since it is a static view on ContractManager data.

image

### DIFF
--- a/src/components/Content.tsx
+++ b/src/components/Content.tsx
@@ -4,13 +4,14 @@ import Refresher from './Refresher'
 interface ContentProps {
   children: ReactNode
   noFade?: boolean
+  noRefresh?: boolean
 }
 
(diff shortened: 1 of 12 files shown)


bordalix (approved):

sahilc0:
assuming this is dev only @pietro909 right? i wont do a ui pass unless asked

PR #627, opened 2026-05-20
feat(wallet): redesign home and settings entry points

sahilc0:
## Summary
- Brings over the scoped wallet home redesign foundation and composition from the draft branch: heading typography, shared haptics, motion constants, sticky home header, portfolio hero, quick actions, real wallet assets, upsells, and recent activity preview.
- Keeps the home backed by real wallet data, without hardcoded demo assets, and hides the homepage activity preview when there are no transactions.
- Removes the global bottom nav for the redesigned flow and ports the settings redesign, including Arkade Mint under Settings > General with back navigation returning to settings.
- Adds the “Swaps are coming soon” drawer for the home Swap action and keeps Scan deep-linking into the send scanner.

## QA notes
- Boltz receive/swap QA still needs manual coverage.
- Arkade Mint needs manual QA now that it lives in Settings, including verifying newly minted assets appear in the wallet.

## Testing
- bun run format:check
- bun run lint
- bun run build
- bun run test:unit -- src/test/App.test.tsx src/test/screens/wallet/index.test.tsx
- Browser smoke: opened the local app, verified wallet home loads, Activity header alignment, and full Activity refresh showing received rows.

Build passed with the existing Vite warnings about publicDir, mnemonic import chunking, and bundle size.

## Summary by CodeRabbit

* **New Features**
  * Haptic feedback for interactive actions
  * Privacy controls to mask balances and fiat amounts
  * Redesigned wallet home: header, portfolio hero, quick actions (Receive, Send, Swap, Scan), assets & upsell sections
  * Dedicated Activity screen and compact Recent Activity preview
  * Token logos for common tickers and improved currency formatting

* **Improvements**
  * Smoother animations and reduced-motion handling
  * Improved navigation/back-stack behavior and settings flow
  * Updated typography, icon color behavior, and UI styling consistency

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/wallet/pull/627?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


sahilc0:
@bordalix heads up for QA: we need a manual pass on Boltz swaps to make sure receives still work, and on Arkade Mint now that it lives in Settings. Please verify Arkade Mint works properly and that assets show up in the wallet after they are minted.

bordalix:
@sahilc0 homepage balance doesn't obey to "sats only" option, i.e. balance is always in fiat, even when the user selects to see sats only values.

bordalix:
Eternal loop between asset mint and asset.
To exit the only solution is to reload the app.

https://github.com/user-attachments/assets/403a7e1f-fbb6-48a6-a3a5-465b26fe97af

bordalix:
@sahilc0 I think we should use the asset ticker instead of name here

Screenshot 2026-05-21 at 15 08 26

bordalix:
@sahilc0 [WaitingForRound.tsx‎](https://github.com/arkade-os/wallet/pull/627/changes#diff-92579789a292572b1a8f46801cf3b542d5fd2e703183e3f0218538e1ff2ed4e2) is back?

sahilc0:
@bordalix pushed a follow-up for these.

- Kept the home hero total as fiat intentionally. Product reasoning: the hero is the portfolio-level unit of account for our primary audience, which is overwhelmingly measuring value in fiat. The current "Sats only" setting is effectively more relevant for bitcoin-denominated rows/transaction views, not for portfolio total across assets. As we expand assets, we should probably revisit whether that setting survives or becomes a BTC/sats display preference.
- Fixed Arkade Mint navigation so it opens as a settings entry point but child pages use the global stack consistently. Back from mint/import/detail/settings should no longer loop back through the wrong asset page or Apps.
- Changed asset row/detail labeling to prefer the asset ticker where it matters, while keeping bitcoin labeled as bitcoin in the home asset row.
- Removed the reintroduced WaitingForRound component and restored the VTXO renewal UI back to the existing inline renewing state.

Validation on the rebased branch: lint, format:check, focused unit tests, and production build all pass locally.

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commits `a92ef222`, `ae7063f0`)

Reviewing changes since `635a80ae`: Arkade Mint navigation refactor and test fixes.

### Summary

Good changes. The Arkade Mint is now navigated to directly from the settings menu via `navigate(Pages.AppAssets)` instead of being rendered inline as a settings sub-view. The new `replace` function in the navigation provider properly trims the back stack, and `WaitingForRound` deletion is clean (zero remaining consumers). Test changes in `send.test.ts` are a real improvement — capturing actual fees from the UI instead of hardcoding makes assertions less brittle.

### Findings

**Minor — dead import** `send.test.ts:1`
```ts
import { wait } from '@testing-library/user-event/dist/cjs/utils/index.js'
```
`wait` is never called anywhere in the file. Remove to keep the import list honest.

**Nit — `replace` without `backTo` leaves back stack unchanged** `navigation.tsx:343-358`
When `replace(page)` is called without `backTo`, the current screen is swapped but nothing is pushed or popped on `backStack.current`. This means "back" navigates to whatever the *previous* back-stack top was, not the page you replaced from. Current call sites all pass `backTo` (except `replace(Pages.AppAssetMintSuccess)` in `Mint.tsx:167`), so verify that case — pressing back from the mint-success screen should go where you intend.

### Verdict

No protocol-critical paths touched. UI/navigation only. **Approved** — fix the dead import at your convenience.

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commit `6e9972ff`)

Reviewing the single new commit since last review (`ae7063f0`).

### Commit `6e9972ff` — "fix test"

Test-only changes to `src/test/e2e/send.test.ts`:

1. **Removed unused `wait` import** — cleanup, good.
2. **Fixed float comparison** (`balance.toFixed(2)` vs raw float equality) — correct fix for flaky JS float arithmetic in USDS balance assertions.
3. **Fixed success message prefix** (added `$` to match actual UI text `$X.XX sent successfully`) — aligns test with the redesigned formatting from this PR.

No protocol-critical code. No security concerns. LGTM.

ghost (commented):
## Arkana Code Review — PR #627 (incremental, commit `8941b6ef`)

Reviewing the new commit since last review (`6e9972ff`): **feat(wallet): add bitcoin detail page (#630)**.

Not protocol-critical (UI/display only — no VTXO, signing, forfeit, round, or exit path changes).

---

### 🔴 Bug: AbortSignal not passed to fetch

**`src/screens/Wallet/BitcoinDetail.tsx:393`** — `fetchHistoricalMarketData(windowSecs, fiat)` creates an `AbortController` (line 384) but never passes `controller.signal` to the fetch call. The function signature in `src/lib/marketData.ts:29` accepts `signal?: AbortSignal` as the 3rd argument.

Current behavior: on unmount/re-render, the abort only prevents the `.then()` state update (line 395 checks `controller.signal.aborted`), but the network request itself is **never cancelled**. On rapid tab-switching between chart windows, this fires up to 6 concurrent in-flight requests that can't be aborted.

Fix:
```ts
fetchHistoricalMarketData(windowSecs, fiat, controller.signal)
```

### 🟡 Stale comment

**`src/components/AssetCard.tsx:86`** — Comment says `"Non-interactive row (e.g. BTC which has no detail screen)"` but BTC now has a detail screen. Minor, but misleading for the next reader.

### 🟡 Module-level cache never evicts

**`src/screens/Wallet/BitcoinDetail.tsx:38`** — `bitcoinChartCache` is a module-scope `Map` that grows but never clears. In practice this is bounded (~6 windows × N fiat currencies), so it won't OOM, but stale price data persists for the entire session lifetime. Consider adding a TTL or clearing on fiat change. Not blocking.

### ℹ️ Notes (non-blocking)

- **`liveline@0.0.7`** — very early-stage dependency (v0.0.7). Acceptable for a chart widget, just be aware upstream API may churn.
- **`src/lib/marketData.ts:4`** — TODO comment (`// TODO: move to final worker URL`) — track this so it doesn't ship to prod pointing at a dev/personal worker.
- Test coverage for the new screen is minimal (1 test: pointer pause/unpause). Navigation, action buttons, delta badge edge cases (0%, negative), and error states (fetch failure) are untested. Not blocking this PR but worth a follow-up.

---

**Verdict**: One real bug (missing abort signal). The rest are minor. Fix the signal pass-through and this is good to go. 👍

ghost (commented):
## Arkana Code Review — PR #627 (incremental, commit `8941b6ef`)

Reviewing the new commit since last review (`6e9972ff`): **feat(wallet): add bitcoin detail page (#630)**.

Not protocol-critical (UI/display only — no VTXO, signing, forfeit, round, or exit path changes).

---

### 🔴 Bug: AbortSignal not passed to fetch

**`src/screens/Wallet/BitcoinDetail.tsx:393`** — `fetchHistoricalMarketData(windowSecs, fiat)` creates an `AbortController` (line 384) but never passes `controller.signal` to the fetch call. The function signature in `src/lib/marketData.ts:29` accepts `signal?: AbortSignal` as the 3rd argument.

Current behavior: on unmount/re-render, the abort only prevents the `.then()` state update (line 395 checks `controller.signal.aborted`), but the network request itself is **never cancelled**. On rapid tab-switching between chart windows, this fires up to 6 concurrent in-flight requests that can't be aborted.

Fix:
```ts
fetchHistoricalMarketData(windowSecs, fiat, controller.signal)
```

### 🟡 Stale comment

**`src/components/AssetCard.tsx:86`** — Comment says `"Non-interactive row (e.g. BTC which has no detail screen)"` but BTC now has a detail screen. Minor, but misleading for the next reader.

### 🟡 Module-level cache never evicts

**`src/screens/Wallet/BitcoinDetail.tsx:38`** — `bitcoinChartCache` is a module-scope `Map` that grows but never clears. In practice this is bounded (~6 windows × N fiat currencies), so it won't OOM, but stale price data persists for the entire session lifetime. Consider adding a TTL or clearing on fiat change. Not blocking.

### ℹ️ Notes (non-blocking)

- **`liveline@0.0.7`** — very early-stage dependency (v0.0.7). Acceptable for a chart widget, just be aware upstream API may churn.
- **`src/lib/marketData.ts:4`** — TODO comment (`// TODO: move to final worker URL`) — track this so it doesn't ship to prod pointing at a dev/personal worker.
- Test coverage for the new screen is minimal (1 test: pointer pause/unpause). Navigation, action buttons, delta badge edge cases (0%, negative), and error states (fetch failure) are untested. Not blocking this PR but worth a follow-up.

---

**Verdict**: One real bug (missing abort signal). The rest are minor. Fix the signal pass-through and this is good to go. 👍

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commits `e05b33af`→`9926e9c3`)

Reviewing 13 new commits since last review (`8941b6ef`): activity alignment, send form redesign, navigation fixes, review feedback, and test fixes.

**Not protocol-critical** — no changes to VTXOs, transaction signing, forfeit paths, round lifecycle, or exit paths. The `Contracts.tsx` changes are display-only. `WaitingForRound` is a UI wrapper around the existing settle flow.

Previous finding ✅ **AbortSignal bug fixed** — `fetchHistoricalMarketData` now receives `controller.signal` at `BitcoinDetail.tsx:407`.

---

### 🟡 Fiat price feed failure no longer has a fallback

**`src/providers/fiat.tsx:74-79`** — Previously, if `getPriceFeed()` returned `undefined`, the wallet fell back to sats-only display (`setConfig({ ...config, currencyDisplay: CurrencyDisplay.Sats })`). This safety net was removed. Now if the price feed fails, `prices.current` stays at all-zeros (`emptyFiatPrices`), and all fiat amounts display as `$0.00`. Users with fiat currency selected would see zero balances with no indication why.

Consider restoring a fallback or showing a stale-price indicator.

### 🟡 `InputAmount` useEffect missing dependencies

**`src/components/InputAmount.tsx:61-64`**:
```tsx
useEffect(() => {
    if (!value || isNaN(Number(value))) return
    setSatsValue(useFiat ? fromFiat(Number(value)) : Number(value))
}, [value])
```
Depends on `useFiat` and `fromFiat` but only lists `[value]`. If user toggles fiat/sats display without changing the numeric value, `satsValue` won't re-derive. The displayed "other value" would be stale after a currency toggle without a value change.

Same pattern at line 67-70:
```tsx
useEffect(() => {
    setError(satsValue ? (satsValue < 0 ? 'Invalid amount' : '') : '')
    setOtherValue(useFiat ? prettyNumber(satsValue, 0) : prettyNumber(toFiat(satsValue), fiatDecimals()))
}, [satsValue])
```
Missing `useFiat`, `toFiat`, `fiatDecimals` in deps.

### 🟡 LNURL response stored with mutated units

**`src/screens/Wallet/Send/Form.tsx:~419`** — `setLnUrlResponse()` stores the LNURL response with `minSendable`/`maxSendable` overwritten from millisatoshis to satoshis. The `LnUrlResponse` type defines these fields as millisatoshis per spec, but the stored values are actually satoshis. Works correctly in current usage (only used for UI validation), but a future consumer reading `lnUrlResponse.minSendable` expecting millisats would get wrong values. Consider using a separate type or renamed fields.

### ℹ️ Duplicate `VITE_DEV_MNEMONIC` declaration

**`src/vite-env.d.ts:8,11`** — `VITE_DEV_MNEMONIC` declared twice in `ImportMetaEnv`. Harmless but noisy.

### ℹ️ Bug fix found: `config.nostr` → `config.nostrBackup`

**`src/providers/notifications.tsx:59,67`** — This is a good catch. The old code referenced `config.nostr` which doesn't exist on the `Config` type — it was always `undefined`, meaning the relay would never connect. Now correctly reads `config.nostrBackup`. 👍

### ℹ️ BIP21 case-insensitive param handling

**`src/lib/bip21.ts:48-76`** — Good improvement checking both `ark`/`ARK`, `amount`/`AMOUNT`, etc. Note that mixed-case like `Ark` would still miss (URLSearchParams is case-sensitive). Acceptable for real-world usage where wallets emit lowercase or uppercase, not mixed.

### ℹ️ SDK `Asset.amount` type change

The SDK bump (0.4.27→0.4.32) changed `Asset.amount` from `number` to `bigint`. All test mocks updated to `BigInt(...)` and `TransactionsList` removed its `BigInt(a.amount)` wrapping. Consistent.

---

**Verdict**: Approving. The yellows are minor quality-of-life issues, none are blocking. Good work on the redesign — clean separation of UI and protocol logic throughout. The `nostrBackup` fix is a valuable side-effect of this PR.

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commits `e05b33af`→`9926e9c3`)

Reviewing 13 new commits since last review (`8941b6ef`): activity alignment, send form redesign, navigation fixes, review feedback, and test fixes.

**Not protocol-critical** — no changes to VTXOs, transaction signing, forfeit paths, round lifecycle, or exit paths. The `Contracts.tsx` changes are display-only. `WaitingForRound` is a UI wrapper around the existing settle flow.

Previous finding ✅ **AbortSignal bug fixed** — `fetchHistoricalMarketData` now receives `controller.signal` at `BitcoinDetail.tsx:407`.

---

### 🟡 Fiat price feed failure no longer has a fallback

**`src/providers/fiat.tsx:74-79`** — Previously, if `getPriceFeed()` returned `undefined`, the wallet fell back to sats-only display (`setConfig({ ...config, currencyDisplay: CurrencyDisplay.Sats })`). This safety net was removed. Now if the price feed fails, `prices.current` stays at all-zeros (`emptyFiatPrices`), and all fiat amounts display as `$0.00`. Users with fiat currency selected would see zero balances with no indication why.

Consider restoring a fallback or showing a stale-price indicator.

### 🟡 `InputAmount` useEffect missing dependencies

**`src/components/InputAmount.tsx:61-64`**:
```tsx
useEffect(() => {
    if (!value || isNaN(Number(value))) return
    setSatsValue(useFiat ? fromFiat(Number(value)) : Number(value))
}, [value])
```
Depends on `useFiat` and `fromFiat` but only lists `[value]`. If user toggles fiat/sats display without changing the numeric value, `satsValue` won't re-derive. The displayed "other value" would be stale after a currency toggle without a value change.

Same pattern at line 67-70:
```tsx
useEffect(() => {
    setError(satsValue ? (satsValue < 0 ? 'Invalid amount' : '') : '')
    setOtherValue(useFiat ? prettyNumber(satsValue, 0) : prettyNumber(toFiat(satsValue), fiatDecimals()))
}, [satsValue])
```
Missing `useFiat`, `toFiat`, `fiatDecimals` in deps.

### 🟡 LNURL response stored with mutated units

**`src/screens/Wallet/Send/Form.tsx:~419`** — `setLnUrlResponse()` stores the LNURL response with `minSendable`/`maxSendable` overwritten from millisatoshis to satoshis. The `LnUrlResponse` type defines these fields as millisatoshis per spec, but the stored values are actually satoshis. Works correctly in current usage (only used for UI validation), but a future consumer reading `lnUrlResponse.minSendable` expecting millisats would get wrong values. Consider using a separate type or renamed fields.

### ℹ️ Duplicate `VITE_DEV_MNEMONIC` declaration

**`src/vite-env.d.ts:8,11`** — `VITE_DEV_MNEMONIC` declared twice in `ImportMetaEnv`. Harmless but noisy.

### ℹ️ Bug fix found: `config.nostr` → `config.nostrBackup`

**`src/providers/notifications.tsx:59,67`** — This is a good catch. The old code referenced `config.nostr` which doesn't exist on the `Config` type — it was always `undefined`, meaning the relay would never connect. Now correctly reads `config.nostrBackup`. 👍

### ℹ️ BIP21 case-insensitive param handling

**`src/lib/bip21.ts:48-76`** — Good improvement checking both `ark`/`ARK`, `amount`/`AMOUNT`, etc. Note that mixed-case like `Ark` would still miss (URLSearchParams is case-sensitive). Acceptable for real-world usage where wallets emit lowercase or uppercase, not mixed.

### ℹ️ SDK `Asset.amount` type change

The SDK bump (0.4.27→0.4.32) changed `Asset.amount` from `number` to `bigint`. All test mocks updated to `BigInt(...)` and `TransactionsList` removed its `BigInt(a.amount)` wrapping. Consistent.

---

**Verdict**: Approving. The yellows are minor quality-of-life issues, none are blocking. Good work on the redesign — clean separation of UI and protocol logic throughout. The `nostrBackup` fix is a valuable side-effect of this PR.

ghost (commented):
## 🔍 Arkana Code Review — `wallet#627`

**Scope:** 123 files, +6217/−1980. UI/UX overhaul — wallet home redesign, settings restructure, privacy masking, haptics, portfolio hooks, dev auto-init plumbing, E2E test updates.

**Protocol-critical assessment:** No VTXO handling, transaction signing, forfeit paths, round lifecycle, or exit path code is modified. All changes are display-layer. SDK version bump (0.4.24→0.4.32) maintains backward compatibility; sub-path import `@arkade-os/sdk/adapters/indexedDB` still valid. No cross-repo breakage (wallet is `private: true`).

---

### 🟡 Issues to Fix

**1. Privacy masking leaks balance magnitude** — `src/lib/format.ts:173-176`

```typescript
const hideDots = (value: string | number | bigint): string => {
  const str = typeof value === 'string' ? value : value.toString()
  const length = str.length * 2 > 6 ? str.length * 2 : 6
  return '·'.repeat(length)
}
```

The masked dot count is derived from the real value's string length. A balance of `1` → 6 dots, `100000` → 12 dots, `10000000` → 16 dots. A shoulder surfer or screenshot can estimate the order of magnitude. Used by `prettyHide` (line 179) and `prettyFiatHide` (line 185), which feed `Balance.tsx` and `TransactionsList.tsx`.

**Fix:** Use a fixed number of masking characters (e.g., `return '·'.repeat(8)`) regardless of value.

---

**2. Debug `console.log` left in production code** — `src/components/Keyboard.tsx:124`

```typescript
console.log({ inputMode, textValue, amountInSats, assetInCents })
```

Remove before merge.

---

**3. Operator precedence ambiguity in receive amount parsing** — `src/screens/Wallet/Receive/QrCode.tsx:331`

```typescript
const sats = inputMode === 'sats' ? num : useFiat || inputMode === 'fiat' ? fromFiat(num) : num
```

JS evaluates `||` before `?:`, so this parses as:
```
inputMode === 'sats' ? num : (useFiat || inputMode === 'fiat') ? fromFiat(num) : num
```

The result happens to be correct today, but it's fragile and confusing. Compare with the cleaner equivalent in `Send/Form.tsx:595`. Add explicit parentheses:
```typescript
const sats = inputMode === 'sats' ? num : (useFiat || inputMode === 'fiat') ? fromFiat(num) : num
```

---

### 🟠 Worth Verifying

**4. BTC portfolio row `fiatAmount` uses `totalFiat` (all assets), not BTC-only** — `src/hooks/usePortfolioFiat.ts:68`

```typescript
fiatAmount: totalFiat,
```

`totalFiat` (line 56) includes asset rows' sats equivalents. If any consumer renders the BTC row's `fiatAmount` expecting BTC-only value, it'll be wrong. If this is intentional (BTC row = summary row), add a comment. Otherwise use `toFiat(balance)`.

---

**5. Transaction status priority reordering** — `src/screens/Wallet/Transaction.tsx:109-116`

Status display priority changed to: `Expired > Unconfirmed > Pending boarding > Settled > Preconfirmed`. Previously `Settled` had higher priority. A settled-but-expired boarding tx now shows "Expired" instead of "Settled." Confirm this is the intended UX.

---

### ℹ️ Informational (no action needed)

**6. `renewVtxos` errors silently swallowed** — `src/providers/wallet.tsx:617`

```typescript
vtxoMgr.renewVtxos().catch(() => {})
```

Pre-existing, not from this PR. But for a Bitcoin L2 wallet, failed VTXO renewal can lead to expiry and fund loss. Consider at minimum logging the error. Filing as a separate issue would be appropriate.

**7. Dev auto-init watchdog is safe.** All dev-only code paths are gated behind `import.meta.env.DEV` (dead-code-eliminated in production). No key leakage vectors. Auth bypass is intentional and scoped. Kill switch exists via `VITE_DEV_AUTO_INIT=false`.

**8. Escape key handler** (`App.tsx:89-96`) navigates to Wallet unconditionally, but redirect gating effect at lines 98-119 immediately bounces back to Unlock/Init if not authenticated. Single-frame flash at worst, no data exposure.

**9. `replace()` in navigation provider** (`navigation.tsx:342-357`) manipulates internal back stack but not `history.pushState`. Browser back button can desync. UX concern, not security — auth gating is render-level.

---

### ✅ What looks good

- `decimal.js` used consistently for satoshi arithmetic — no IEEE 754 precision issues
- `cmdk` → `cmdk-base` and `vaul` → `vaul-base` migration is clean, zero leftover imports
- `web-haptics` correctly lazy-initialized, respects `prefers-reduced-motion`, errors swallowed (appropriate)
- AbortController pattern in wallet init is well-designed — prevents concurrent init races
- Privacy masking in `PrivacyAmount.tsx` uses `AnimatePresence` with both branches always in the tree — no timing side-channel
- No cross-repo breakage risk (private package, stable SDK surface)

---

**Verdict:** Approving — this is a display-layer PR with no protocol-critical changes. Items #1-3 should be fixed before merge (especially #1, the privacy leak). Items #4-5 need author confirmation on intent.

_🤖 Arkana — automated review_

ghost (approved):
Approving — display-layer PR, no protocol-critical code touched. See detailed comment for 3 items to fix before merge (privacy masking leak, console.log, operator precedence).

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commits `9926e9c3`→`0d519689`)

Reviewing 36 new commits since last review: Fiats→Currencies rename, BTC currency unit support, advanced tools in settings, mint validation, e2e test fixes, price feed fallback, TypeScript build fixes, and various UI/UX refinements.

### Protocol-Critical Assessment

**No protocol-critical code modified.** Changes are in the UI/display layer — amount formatting, currency display, settings restructure. Send/receive logic (satoshi values flowing to the SDK) is unchanged. Safe from a protocol standpoint.

### Findings

**🟡 IMPORTANT — `InputAmount.tsx:63` — useEffect dependency on non-memoized functions**

```ts
useEffect(() => {
  if (!value || isNaN(Number(value))) return
  setSatsValue(useFiat ? fromFiat(Number(value)) : Number(value))
}, [value, fromFiat, useFiat])
```

`fromFiat` and `toFiat` are defined inline in `FiatProvider` without `useCallback`, creating new references every render. Adding them to `useEffect` deps means the effect fires every render → `setSatsValue` → re-render → new function refs → loop. React's state dedup may prevent a true infinite loop (if the computed value is the same), but it's fragile and wasteful. Consider wrapping `fromFiat`/`toFiat`/`fromBTC`/`toBTC` in `useCallback` in the provider.

**🟡 IMPORTANT — `providers/config.tsx:74` — `Unit`/`CurrencyDisplay` enum duplication with unsafe casts**

```ts
currencyDisplay: normalizeBitcoinUnit(config.currencyDisplay as `${CurrencyDisplay}`) as unknown as CurrencyDisplay,
```

`as unknown as CurrencyDisplay` silences TypeScript completely. `Unit` and `CurrencyDisplay` are separate enums with identical values (`BTC`/`sats`/`₿`). If either changes independently, this breaks silently at runtime. The double-cast appears in ~8 places. Consider unifying into a single type.

**🟢 NIT — `lib/format.ts:178` — `hideDots` called with unused argument**

`hideDots()` takes zero params, but `prettyHide` and `prettyFiatHide` pass `value` to it. Harmless (JS ignores extra args) but suggests incomplete refactor.

**🟢 NIT — `providers/fiat.tsx:79` — Silent currency fallback on price feed failure**

When the price feed fails, the user's currency silently changes to BTC with `save = false`. No user notification, and on reload the cycle repeats if the feed is still down. Intentional but worth a toast or subtle indicator.

**🟢 NIT — `lib/validators.ts:3` — `isInvalidMintAmount` allows 0**

`onlyHasDigits(0)` returns `true` (valid) due to `!amount` guard. Minting 0 tokens is probably not useful — the mint button is likely disabled elsewhere, but the validator itself permits it.

### Good Fixes Noted

- `notifications.tsx:59`: `config.nostr` → `config.nostrBackup` — genuine bug fix, the relay toggle was previously broken.
- `Send/Form.tsx:411`: Guard preventing LNURL resolution when ark address exists — correct priority.
- `bip21.ts:83-89`: Clean BIP21 encoding with proper trailing separator cleanup.
- `BitcoinDetail.tsx:62`: Sensible USD fallback when currency is BTC (can't price BTC in BTC).

### Verdict

**Approved.** No money-losing bugs, no security issues, no protocol-critical changes. The two IMPORTANT items are tech debt worth tracking but not merge blockers. Good test coverage additions (usePortfolioFiat, bip21, format, contracts, bitcoin-detail, form e2e tests).

_🤖 Arkana — automated review_

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commits `0d519689`→`fae6b9a7`)

Reviewing 6 new commits (plus master merges): send-max fix via `valueSats`, `roundTxid` fallback for offboarding batches, price chart proxy URL, contracts screen polish, `hideDots()` fix, build fixes.

### Protocol-Critical Assessment

**No protocol-critical code modified.** The `roundTxid` fallback in `Transaction.tsx` is display-only — it shows a txid where previously an empty string was shown. No signing, VTXO, or fund-flow logic changed.

### Findings

**✅ `InputAmount.tsx` — `valueSats` prop is a clean fix**

The new `valueSats` prop bypasses fiat round-trip conversion when `handleSendAll` sets an exact satoshi value. This prevents the display drift bug where max-send amount could shift through `toFiat → fromFiat` floating-point round-trips. The priority logic (`valueSats !== undefined` guards) is correct, and clearing `valueSats` to `undefined` on manual input correctly falls back to the conversion path.

**✅ `Transaction.tsx:127` — `roundTxid` fallback is correct**

For offboarding batch transactions that have neither `boardingTxid` nor `redeemTxid`, the `roundTxid` is now shown instead of an empty string. `isOffchainTx` flag (line 128) is unaffected — still only checks `redeemTxid`. Pure display improvement.

**✅ `format.ts:178,183` — `hideDots()` fix**

Addresses the NIT I flagged last review. Clean.

**✅ `marketData.ts:4` — Production proxy URL**

`bordalix.workers.dev` → `price-chart-proxy.arkade.money`. Good to move off dev infra.

**✅ `Contracts.tsx` + CSS — UI polish only**

Semantic HTML (`<article>`, `<header>`, `<section>`), BEM-style CSS classes, copy-to-clipboard with accessible labels, summary stats, staggered entry animations with `prefers-reduced-motion` respect. No logic changes. Tests updated to match new class names and capitalized state labels.

### Previous items status

- **`hideDots` NIT** — Fixed ✅
- **`fromFiat`/`toFiat` not memoized** — Still unfixed (pre-existing tech debt, not from this PR)
- **`Unit`/`CurrencyDisplay` enum duplication** — Still unfixed (tech debt)

**Verdict:** Approved. All changes are display-layer fixes and polish. The `valueSats` pattern is the right fix for the send-max precision issue. No money-losing bugs, no security issues.

_🤖 Arkana — automated review_

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commit `e2e5fa5f`)

Single commit: `feat(settings): add currency icons (#663)` — adds flag SVG logos for CNY, EUR, GBP, JPY to the fiat currency selector in Settings, plus a generic `renderStart` slot on the `Select` component.

### Findings

**✅ No protocol-critical changes.** Pure UI — flag icons and CSS. No signing, VTXO, or fund-flow logic touched.

**🟢 NIT — `TokenLogo.tsx:212` — Hardcoded SVG `clipPath` IDs will collide if multiple instances render simultaneously**

`<clipPath id='gb-flag-circle'>` (and pre-existing `us-flag-circle`, `br-flag-circle`) use static DOM IDs. If two `UnitedKingdomFlagLogo` components render on the same page (e.g., a GBP transaction in the list while a settings modal is open), the browser picks whichever `#gb-flag-circle` it finds first. Fix: use `useId()` or inline the clip via `clipPath="circle(50%)"`. Pre-existing pattern, not introduced by this commit, but worth tracking.

**✅ `Select.tsx` — `renderStart` prop is clean**

Optional render prop, no breaking change to existing consumers. The `<span>` wrapper around `<p>` is slightly off semantically (`<p>` inside `<span>`) but browsers handle it fine and it's consistent with the existing structure.

**✅ `EuropeanUnionFlagLogo` — Trig math is correct**

12 stars at 30° intervals, radius 7, centered at (16,16). Verified.

**✅ Tests updated correctly**

Default-selection test now queries `[data-testid="green-status-icon"]` instead of bare `svg`, since every row now has an icon SVG. Icon presence test added for all 7 currency options.

**Verdict:** Approved. No issues.

_🤖 Arkana — automated review_

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commit `e2e5fa5f`)

Single commit: `feat(settings): add currency icons (#663)` — adds flag SVG logos for CNY, EUR, GBP, JPY to the fiat currency selector in Settings, plus a generic `renderStart` slot on the `Select` component.

### Findings

**✅ No protocol-critical changes.** Pure UI — flag icons and CSS. No signing, VTXO, or fund-flow logic touched.

**🟢 NIT — `TokenLogo.tsx:212` — Hardcoded SVG `clipPath` IDs will collide if multiple instances render simultaneously**

`<clipPath id='gb-flag-circle'>` (and pre-existing `us-flag-circle`, `br-flag-circle`) use static DOM IDs. If two `UnitedKingdomFlagLogo` components render on the same page (e.g., a GBP transaction in the list while a settings modal is open), the browser picks whichever `#gb-flag-circle` it finds first. Fix: use `useId()` or inline the clip via `clipPath="circle(50%)"`. Pre-existing pattern, not introduced by this commit, but worth tracking.

**✅ `Select.tsx` — `renderStart` prop is clean**

Optional render prop, no breaking change to existing consumers. The `<span>` wrapper around `<p>` is slightly off semantically (`<p>` inside `<span>`) but browsers handle it fine and it's consistent with the existing structure.

**✅ `EuropeanUnionFlagLogo` — Trig math is correct**

12 stars at 30° intervals, radius 7, centered at (16,16). Verified.

**✅ Tests updated correctly**

Default-selection test now queries `[data-testid="green-status-icon"]` instead of bare `svg`, since every row now has an icon SVG. Icon presence test added for all 7 currency options.

**Verdict:** Approved. No issues.

_🤖 Arkana — automated review_

ghost (changes requested):
## Arkana Code Review — PR #627 (incremental, commits `e2e5fa5f`→`b2ffb739`)

Reviewing: `5f166df9` (Fixes), `0a9784c5` (Remove onboarding button), `ea0e79cd` (Clean qrcode), `b2ffb739` (clean Reminder), plus merge-from-master commits that brought in significant new code.

---

### 🔴 BUG — `src/providers/wallet.tsx:238–252` — `initWallet` called TWICE during dev auto-init

```ts
// First call (line ~238):
if (devMnemonic) {
  await initWallet({ mnemonic: devMnemonic })
} else if (devNsec) {
  const privateKey = nsecToPrivateKey(devNsec)
  await initWallet({ privateKey })
}
// ... clearTimeout, sessionStorage cleanup ...
// Second call (line ~251):
if (devMnemonic) await initWallet({ mnemonic: devMnemonic })
else if (devNsec) await initWallet({ privateKey: nsecToPrivateKey(devNsec) })
```

The wallet is initialized twice in the happy path. This is a merge/rebase artifact — the first block was the original, the second was the diff addition. The duplicate `initWallet` could register the client twice with the ASP, create duplicate key derivation state, or trigger double-reload side effects depending on the SDK internals. **Remove one of the two blocks** (keep the first one, which handles nsec via an intermediate variable).

### 🟡 OBSERVATION — `src/screens/Wallet/Send/Form.tsx` — Branta now sends every typed address to its API

Previously, only QR-scanned addresses containing `branta_id` + `branta_secret` params were sent to Branta. Now every typed recipient triggers a (debounced 400ms) `brantaClient.getPayments(typed)` call. With `privacy: 'strict'` this should be safe per the SDK contract, but it's a meaningful privacy behavior change — every Bitcoin/Ark/LN address a user types is now sent to a third-party API. Ensure the Branta v2 SDK's strict mode genuinely doesn't leak the raw address (e.g., it should hash or zero-knowledge-proof it before transmitting).

### ✅ GOOD — Storage key mutual exclusion (`mnemonic.ts` / `privateKey.ts`)

`setMnemonic` now removes `NSEC_STORAGE_KEY`, `setPrivateKey` removes `MNEMONIC_STORAGE_KEY`. This prevents stale credentials from coexisting in localStorage. Clean safety improvement with proper tests.

### ✅ GOOD — `BUILD_VERSION_TOO_OLD` error handling (`asp.ts`, `asp.tsx`)

The `ArkError` instanceof + name check is correct. The regex fallback `err.message.match(/>=\s*v?([\d.]+)/)` for when metadata is absent is a solid defensive measure, with a test covering the no-metadata path. `aspErrorText` helper centralizes the user-facing message across all 8+ call sites.

### ✅ GOOD — BIP21 case-insensitive query keys (`bip21.ts`)

The `getParam` helper doing `key.toLowerCase() === name` fixes real-world interop with QR generators that capitalize params. The `useGrouping: false` fix in `encodeBip21` prevents `1,000` from breaking amount parsing. Both have tests.

### ✅ GOOD — `prettyAssetNumber` scientific notation (`assets.ts`)

`/e/i.test(num)` → `new Decimal(num).toFixed()` prevents the regex stripping of `e` from scientific notation strings like `-8e-8`. Edge case but real when `centsToUnits` produces tiny decimals.

### ✅ GOOD — Transaction settle button simplification (`Transaction.tsx`)

`status === 'Preconfirmed'` correctly replaces the old `!unconfirmed && !expired && !settled && !settleSuccess` conjunction. The `handleResend` removal is fine — it was a disabled TODO stub.

### ✅ GOOD — Contracts virtualization (`Contracts.tsx`)

`@tanstack/react-virtual` with `script` as the key, `measureElement` for dynamic heights, and `overscan: 8` is the right approach for wallets with many contracts. Signer-status badges and search are pure UI additions.

### ✅ GOOD — QR code `resolveQrValue` (`QrCode.tsx`)

Prevents async rebuilds from silently reverting a user's copy-sheet selection. Well-tested with the exported pure function.

### ✅ GOOD — `displayLnurl` gating

LNURL is only included in the BIP21 when no amount is set (amountless mode). This is correct per the LNURL spec — an amount-specific QR should use a bolt11 invoice, not the LNURL.

---

**Verdict:** Request changes — the double `initWallet` call is a clear bug that must be fixed before merge. Everything else is solid.

_🤖 Arkana — automated review_

ghost (commented):
## Arkana Code Review — PR #627 (incremental, merge commit 197285c2)

_🤖 Arkana — automated review_

**Scope:** Merge of master into the feature branch, bringing in PRs #675, #685, #686, #684.

### Changes reviewed

| Area | What changed |
|------|-------------|
| **SDK bump** | arkade-os/sdk 0.4.36→0.4.37, boltz-swap 0.3.41→0.3.42 |
| **Docker/nginx** | New VITE_LNURL_SERVER_URL ARG; entrypoint refactored to iterate all VITE_* env vars; .mjs MIME fix for service worker |
| **fromRuntimeEnv helper** | Filters leftover __VITE_*__ placeholders so they are treated as unset — applied to VITE_ARK_SERVER, VITE_LNURL_SERVER_URL, VITE_BOLTZ_URL |
| **Support screen** | Adds buildVersion/sdkVersion to Chatwoot custom attributes |
| **Type defs** | VITE_LNURL_SERVER_URL added to ImportMetaEnv |

### Verdict

✅ **Clean merge, no issues.**

- No conflict markers.
- No branch-specific logic changes — all deltas originate from already-merged master PRs.
- fromRuntimeEnv is well-implemented: startsWith(__VITE_) correctly catches un-substituted Docker placeholders without false-positiving on real values.
- docker-entrypoint.sh loop approach is a nice simplification; the sed escaping is preserved.
- The .mjs nginx rule with no-cache is correct — service workers must not be immutably cached.
- No protocol-critical code touched (no VTXO, signing, forfeit, or round lifecycle changes).

No action needed. Prior review feedback still applies to the branch-specific commits.

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commit `64749fcf`)

_🤖 Arkana — automated review_

**Scope:** Single commit — removes the duplicate `initWallet` block I flagged in my previous review.

### ✅ Fix verified

The duplicate dev-auto-init block (old lines 235-240 inside the auth-check effect) has been removed. `autoInit()` at line 222 is now the sole `initWallet` call site for the dev mnemonic/nsec path. The watchdog/reload scaffolding around the removed block was also cleaned up — no dead code left behind.

Control flow is now: `autoInit()` → `initWallet()` → `setAuthState('authenticated')` — single path, no double-registration risk.

### Verdict

✅ **Approve.** The bug is fixed correctly. No protocol-critical code touched — this is dev-tooling plumbing only.

Prior review feedback (Branta privacy observation, etc.) still applies as informational notes but nothing blocking.

ghost (approved):
## Arkana Code Review — PR #627 (incremental, commit `64749fcf`)

_🤖 Arkana — automated review_

**Scope:** Single commit — removes the duplicate `initWallet` block I flagged in my previous review.

### ✅ Fix verified

The duplicate dev-auto-init block (old lines 235-240 inside the auth-check effect) has been removed. `autoInit()` at line 222 is now the sole `initWallet` call site for the dev mnemonic/nsec path. The watchdog/reload scaffolding around the removed block was also cleaned up — no dead code left behind.

Control flow is now: `autoInit()` → `initWallet()` → `setAuthState('authenticated')` — single path, no double-registration risk.

### Verdict

✅ **Approve.** The bug is fixed correctly. No protocol-critical code touched — this is dev-tooling plumbing only.

Prior review feedback (Branta privacy observation, etc.) still applies as informational notes but nothing blocking.

tiero:
Go to Display and set Currency **BTC** and BTC Unit to **Sats**

then go on the send page

it show's BTC instead of SATS on the right
image

tiero:
https://github.com/user-attachments/assets/395db8ca-da55-4bd8-ae23-24e190b23b6d

Another edge case with BTC UNIT conversions. Probably we should redesign that component

BTC currency and BTC as unit 

Sending still applies SATS not BTC

tiero:
the activity tab scroll view has lot of bottom padding
image

tiero:
Can we hide non-verified assets by default (ie. exists in this github page https://github.com/ArkLabsHQ/asset-registry)? They can still be shown in Arkade Mint eventually

image

PR #628, opened 2026-05-21
feat(wallet): align recent activity with prototype

sahilc0:
## Summary
- Align the home Recent activity section, the full Activity list, and transaction detail receipt layout with the prototype draft PR visuals.
- Keep the Activity page on the shared PR #627 header so the back button, centered title, and divider match the rest of the app.
- Add swap/account-asset rendering, BRL/DPIX/DEPIX display mapping, and measured virtual rows so taller multi-asset activity rows do not overlap.
- Support VITE_DEV_MNEMONIC for local demo auto-init so testing no longer depends on the broken legacy default nsec.

## Visual evidence
- Verified locally in the Codex browser at the branch portless URL after the recent activity/header updates.

## Testing
- bun run lint
- bun run build
- Commit hooks: prettier --check src and eslint

## Affected areas
- src/components/TransactionsList.tsx
- src/screens/Wallet/RecentActivitySection.tsx
- src/screens/Wallet/Activity.tsx
- src/screens/Wallet/Transaction.tsx
- src/components/Details.tsx and src/components/Table.tsx
- src/providers/wallet.tsx, src/App.tsx, src/index.tsx, src/vite-env.d.ts

## Summary by CodeRabbit

* **New Features**
  * Added a new loading state display for the transaction settlement process.
  * Enhanced transaction detail view with improved formatting and visual layout.
  * Added swap transaction visualization with route indicators.

* **Improvements**
  * Improved empty state messaging when no transactions are available.
  * Refined activity screen interface for a cleaner experience.
  * Better asset ticker normalization for consistent display across the app.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/wallet/pull/628?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ghost (changes requested):
## Arkana Code Review — PR #628

UI alignment work looks solid overall. Several issues need attention before merge — one potential crash, one memory leak, a protocol-adjacent status logic change, and some incomplete plumbing.

---

### 🔴 Bug: Crash in `WaitingForRound` when no info logs exist

**`src/components/WaitingForRound.tsx:36`**

```ts
setLogMessage(getInfoLogLineMsg(logLength - 1))
```

If `logLength` is 0 (no info-level logs yet), this calls `getInfoLogLineMsg(-1)`. In `src/lib/logs.ts:55`:

```ts
export const getInfoLogLineMsg = (index: number) => getInfoLogs()[index].msg
```

`getInfoLogs()[-1]` is `undefined` → `.msg` throws **TypeError: Cannot read properties of undefined**.

This will crash the settlement UI. Add a guard:
```ts
if (logLength > 0) setLogMessage(getInfoLogLineMsg(logLength - 1))
```

---

### 🔴 Bug: Memory leak in `WaitingForRound` on early unmount

**`src/components/WaitingForRound.tsx:23-29`**

```ts
useEffect(() => {
    let interval: NodeJS.Timeout
    sleep(2000).then(() => {
      interval = setInterval(...)  // ← runs after cleanup if unmount < 2s
    })
    return () => clearInterval(interval) // ← interval is undefined here
  }, [])
```

If the component unmounts before the 2s sleep resolves, `clearInterval(undefined)` is a no-op. The `.then()` callback still fires, creating an `setInterval` that is **never cleared** — it continues calling `setLogLength` on an unmounted component (React state-after-unmount warning) and parsing localStorage JSON every 500ms forever.

Fix with a `cancelled` flag:
```ts
useEffect(() => {
    let interval: NodeJS.Timeout
    let cancelled = false
    sleep(2000).then(() => {
      if (cancelled) return
      interval = setInterval(() => setLogLength(getInfoLogsLength()), 500)
    })
    return () => { cancelled = true; clearInterval(interval) }
  }, [])
```

---

### 🟡 Protocol-adjacent: Settlement status logic change

**`src/screens/Wallet/Transaction.tsx:106-124`** (new) vs old lines 112-125

Old code:
```ts
status: settleSuccess || tx.settled ? 'Settled' : expiredBoardingTx ? ...
```

New code drops `settleSuccess` from the status ternary entirely and instead does an optimistic update at line 96:
```ts
if (tx) setTxInfo({ ...tx, preconfirmed: false, settled: true })
```

The old code (commit `29b140b3`) had an **explicit comment** explaining why `setTxInfo` was intentionally NOT called after settle:

> *"Updating txInfo after an async operation can corrupt navigation if the user has navigated to a different transaction"*

The new WaitingForRound overlay likely prevents navigation during settle, mitigating the race. But please confirm this is intentional — if a user can back-navigate during the `sleep(2000)` at line 92 before `setSettleSuccess(true)`, the stale `setTxInfo` call would corrupt whichever transaction is viewed next.

**Flagging for human review** since this touches settlement UX flow.

---

### 🟡 `Fiats.BRL` added without fiat provider support

**`src/lib/types.ts:51`** adds `BRL = 'BRL'` to the `Fiats` enum.
**`src/lib/format.ts:69`** maps BRL/DPIX/DEPIX tickers to `Fiats.BRL`.

But `Fiats.BRL` has **no backing infrastructure**:
- `src/lib/fiat.ts`: `FiatPrices` interface has no `brl` field
- `src/providers/fiat.tsx`: No `toBRL`/`fromBRL` conversion functions; `toFiat()`/`fromFiat()` don't handle `Fiats.BRL` and silently fall through to USD
- `FIAT_SYMBOLS` has no `R$` entry for BRL
- Price feed (`blockchain.info/ticker`) is not queried for BRL

If a user's fiat is set to BRL (via `fiatForTicker` returning `Fiats.BRL`), all conversions silently use USD rates. Either add full BRL support or don't add the enum value yet.

---

### 🟡 Duplicated utility functions

`accountTickerForAssetTicker` is now copy-pasted in **two files**:
- `src/components/TransactionsList.tsx:351-358`
- `src/screens/Wallet/Transaction.tsx` (new, bottom of file)

Same for `getTokenLogoTicker` (also in `src/components/AssetCard.tsx:115`).

Extract to a shared module (e.g. `src/lib/assets.ts` or `src/lib/tokens.ts`) to prevent drift. Three copies of the same ticker-mapping logic is a maintenance hazard.

---

### 🟢 Minor: `getInfoLogsLength` performance in polling loop

**`src/components/WaitingForRound.tsx:26`** — Every 500ms, `getInfoLogsLength()` calls `getInfoLogs()` → `getLogs()` → `JSON.parse(localStorage.getItem('logs'))`. That's parsing up to 50KB of JSON twice per second. Not a blocker, but consider caching the length or using a lighter check.

---

### 🟢 Minor: Activity.tsx removed wallet refresh on empty txs

**`src/screens/Wallet/Activity.tsx`** — The old code auto-refreshed via `svcWallet.reload()` if `txs.length === 0`. New code just shows `<EmptyTxList />`. If the wallet provider already handles refresh elsewhere, this is fine. If not, users opening Activity before the first sync completes will see an empty screen with no way to trigger a load.

---

### Summary

| Severity | Issue |
|----------|-------|
| 🔴 Crash | `WaitingForRound` throws TypeError when no info logs |
| 🔴 Leak  | `WaitingForRound` interval never cleared on early unmount |
| 🟡 Protocol | Settlement status change reverses intentional design; needs human review |
| 🟡 Incomplete | `Fiats.BRL` added without conversion support |
| 🟡 Maintenance | 3 copies of `accountTickerForAssetTicker` / `getTokenLogoTicker` |
| 🟢 Perf | 500ms localStorage JSON parse polling |
| 🟢 UX | Activity auto-refresh removed |

Fix the two 🔴 items, confirm the settlement logic change is intentional, and this is good to go.

sahilc0:
Addressed the Arkana findings in 6cbb30af:

- Guarded WaitingForRound before reading the latest info log, so an empty info-log list cannot call getInfoLogLineMsg(-1).
- Added a cancellation flag for the delayed polling interval so unmounting before the initial delay does not leak an interval.
- Restored the safer settlement status flow by removing the optimistic setTxInfo update and using settleSuccess in the status calculation again.
- Removed the incomplete Fiats.BRL enum/fiatForTicker mapping while keeping BRL/DPIX/DEPIX as token-logo/account display mappings only.
- Moved ticker/logo mapping helpers into TokenLogo.tsx and reused them from AssetCard, TransactionsList, and Transaction.

Local verification after the fix: bun run lint and bun run build.

ghost (approved):
## Arkana Follow-up Review — commit `6cbb30af`

All issues from my previous review have been addressed:

| Previous Finding | Status |
|---|---|
| 🔴 `WaitingForRound` crash on empty logs | ✅ Fixed — `logLength > 0` guard added |
| 🔴 `WaitingForRound` interval leak on early unmount | ✅ Fixed — `cancelled` flag pattern applied |
| 🟡 Settlement status logic reversal | ✅ Fixed — restored `settleSuccess \|\| tx.settled` check, removed the optimistic `setTxInfo` mutation |
| 🟡 `Fiats.BRL` without conversion support | ✅ Fixed — removed from `Fiats` enum and `fiatForTicker` mapping; BRL kept only in `TokenLogoTicker` for avatar rendering (correct) |
| 🟡 Duplicated utility functions | ✅ Fixed — `accountTickerForAssetTicker` and `tokenLogoTickerForTicker` extracted to shared `TokenLogo.tsx` |

One minor note: `TransactionAssetAvatar` now does `tokenLogoTickerForTicker(accountTickerForAssetTicker(ticker) ?? ticker)` — the double-normalization is redundant (accountTickerForAssetTicker already returns valid TokenLogoTicker values) but functionally harmless. Not blocking.

⚠️ **Human review still required** — settlement status logic was reworked (even if now restored to the safer pattern). A human should confirm the `settleSuccess` flow is correct before merge.

LGTM pending human sign-off.

PR #629, opened 2026-05-21
feat(send): align send flow with redesign prototype

sahilc0:
## Summary
- Aligns the send form fields with the draft redesign prototype by moving shared form fields from the gray `Shadow` wrapper to the draft `input-shell` surface.
- Ports the draft-style send asset selector: token icons, selected asset label/balance, dropdown menu surface, option rows, and menu motion.
- Updates amount and button interactions for the send flow, including a pill-style Max control with haptic feedback and safer disabled/loading button handling.
- Adds dev mnemonic auto-init support and a dev-only wallet storage reset path so local mnemonic seeds do not fall back to stale nsec wallet state.

- `src/screens/Wallet/Send/Form.tsx` - send form layout, asset selector, selector menu, haptics.
- `src/components/InputContainer.tsx` - shared input shell structure.
- `src/components/InputAmount.tsx` - Max pill control and controlled amount value preservation.
- `src/components/Button.tsx` - disabled/loading click and haptic guard.
- `src/index.css` - send form field, asset selector, dropdown, and dark-mode styling.
- `src/App.tsx`, `src/index.tsx`, `src/providers/wallet.tsx`, `src/vite-env.d.ts` - dev mnemonic auto-init support.

## Testing
- `bun run lint`
- `bun run test:unit src/test/screens/wallet/send.test.tsx`
- `bun run build`
- Local visual verification on `wt-send-flow-changes-20260521.arkade.localhost` for send form field shells, neutral pill labels, and bottom CTA.

## Notes
- This PR is stacked on #627 and intentionally avoids copying the full prototype send implementation where it would replace newer PR-627 send logic.
- The local seed itself is not committed; only mnemonic-based dev auto-init support is included.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,7 +65,7 @@ export default function App() {
   const devAutoInitHomeRedirected = useRef(false)
   const hasDevAutoInit =
     import.meta.env.DEV &&
-    Boolean(import.meta.env.VITE_DEV_NSEC) &&
+    Boolean(import.meta.env.VITE_DEV_MNEMONIC || import.meta.env.VITE_DEV_NSEC) &&
     import.meta.env.VITE_DEV_AUTO_INIT !== 'false' &&
     !devAutoInitFailed
 
@@ -102,7 +102,7 @@ export default function App() {
     // avoid redirect if the user is still setting up the wallet
     if (initInfo.password || initInfo.privateKey) return
     if (!walletLoaded) return navigate(Pages.Loading)
-    // dev auto-init: stay on loading screen while VITE_DEV_NSEC initializes the wallet
+    // dev auto-init: stay on loading screen while the configured dev seed initializes the wallet
     if (hasDevAutoInit && !initialized) return
     if (!wallet.pubkey) return navigate(Pages.Init)
     if (authState === 'locked') return navigate(Pages.Unlock)
@@ -189,7 +189,7 @@ export default function App() {
 
   // Boot animation: persists on Loading, then flies to the LogoIcon position when
   // Wallet is reached. For any other destination (Unlock, Init, etc.), exits with fly-up.
-  // Skip in dev with VITE_DEV_NSEC — the fast auto-init races with the animation.
+  // Skip in dev auto-init — the fast seed restore races with the animation.
   useEffect(() => {
     if (hasDevAutoInit) return
 
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -80,10 +80,11 @@ export default function Button({
 
   const handleClick = useCallback(
     (event: any) => {
+      if (disabled || loading) return
       hapticTap()
       onClick(event)
     },
-    [onClick],
+    [disabled, loading, onClick],
   )
 
   return (
@@ -92,6 +93,8 @@ export default function Button({
       type='button'
       className={cn(buttonVariants({ variant: resolvedVariant }), pressed && 'pressed', className)}
       disabled={disabled}
+      aria-busy={loading || undefined}
+      aria-disabled={loading || undefined}
       onClick={handleClick}
       onMouseDown={handlePressStart}
       onMouseUp={handlePressEnd}
--- a/src/components/InputAmount.tsx
+++ b/src/components/InputAmount.tsx
@@ -5,9 +5,9 @@ import { ConfigContext } from '../providers/config'
 import { prettyNumber } from '../lib/format'
 import { FIAT_SYMBOLS } from '../lib/fiat'
 import { LimitsContext } from '../providers/limits'
-import Focusable from './Focusable'
 import { AssetOption } from '../lib/types'
 import { TextSecondary } from './Text'
+import { hapticLight } from '../lib/haptics'
 
 interface InputAmountProps {
   asset?: AssetOption
@@ -85,7 +85,6 @@ export default function InputAmount({
       : fiatSymbol
         ? `${fiatSymbol}${otherValue}`
         : `${otherValue} ${config.fiat}`
-  const fontStyle = { color: 'var(--neutral-500)', fontSize: '13px' }
   const bottomLeft =
     minimumSats && satsValue !== undefined && satsValue < minimumSats
       ? `Min: ${prettyNumber(minimumSats)} ${minimumSats === 1 ? 'SAT' : 'SATS'}`
@@ -109,23 +108,24 @@ export default function InputAmount({
           disabled={disabled}
           readOnly={readOnly}
           onChange={handleAmountChange}
-          value={value ? Number(value) : ''}
+          value={value ?? ''}
           onKeyUp={(ev) => ev.key === 'Enter' && onEnter && onEnter()}
         />
         <TextSecondary>{rightLabel}</TextSecondary>
       </label>
       {onMax && !disabled && !readOnly ? (
-        <Focusable onEnter={onMax} fit>
-          <p
-            role='button'
-            onClick={onMax}
-            aria-label='Set maximum amount'
-            data-testid='input-amount-max'
-            style={{ ...fontStyle, color: 'var(--purpletext)', cursor: 'pointer' }}
-          >
-            Max
-          </p>
-        </Focusable>
+        <button
+          type='button'
+          className='pill-base'
+          onClick={() => {
+            hapticLight()
+            onMax()
+          }}
+          aria-label='Set maximum amount'
+          data-testid='input-amount-max'
+        >
+          Max
+        </button>
       ) : null}
     </InputContainer>
   )
--- a/src/components/InputContainer.tsx
+++ b/src/components/InputContainer.tsx
@@ -2,7 +2,6 @@ import { ReactNode } from 'react'
 import ErrorMessage from './Error'
 import FlexRow from './FlexRow'
 import FlexCol from './FlexCol'
-import Shadow from './Shadow'
 import Text from './Text'
 
 interface InputContainerProps {
@@ -43,12 +42,12 @@ export default function InputContainer({
   )
 
   return (
-    <FlexCol>
+    <FlexCol className='input-container'>
       <FlexCol gap='0.5rem'>
         {label || right ? <TopLabel /> : null}
-        <Shadow>
+        <div className='input-shell'>
           <FlexRow between>{children}</FlexRow>
-        </Shadow>
+        </div>
         {bottomLeft || bottomRight ? <BottomLabel /> : null}
       </FlexCol>
       <ErrorMessage error={Boolean(error)} text={error ?? ''} />
--- a/src/index.css
+++ b/src/index.css
@@ -1121,6 +1121,8 @@ button.reminder-button {
   background-color: transparent;
   caret-color: var(--black);
   color: var(--black);
+  outline: none;
+  min-width: 0;
 }
 
 input:focus::placeholder {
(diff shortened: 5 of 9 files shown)


ghost (changes requested):
## Code Review — arkade-os/wallet#629

Reviewed the full diff, read every changed file in context, checked cross-repo impact, git history, and test coverage.

### 🔴 MUST FIX

**1. E2E tests will break — `data-testid` format changed (Form.tsx:913)**

The asset option `data-testid` changed from `asset-${asset.ticker.toLowerCase()}-option` to `asset-${asset.assetId}-option`. The e2e tests at `src/test/e2e/send.test.ts:157,193` select `asset-tst-option` (by ticker). After this PR, the test ID will be something like `asset-a1b2c3...f4-option` (a hex asset ID), so both e2e test cases will fail.

Either revert to the ticker-based test ID or update the e2e tests to use the new format.

**2. `AssetIcon` is defined as a component *inside* the render body (Form.tsx:~712)**

```tsx
const AssetIcon = ({ asset }: { asset: AssetOption | null }) => { ... }
```

This is declared inside `SendForm()`, which means React creates a **new component type on every render**. Every re-render (keystroke in amount field, etc.) will unmount and remount every `<AssetIcon>`, destroying DOM state and killing any CSS transitions on icons. It also defeats React's reconciliation.

Move `AssetIcon` outside `SendForm` (or at minimum, wrap it in `useMemo`/extract to module scope). It only depends on `asset`, which can be a prop.

**3. `resetDevWalletStorage` infinite reload risk (index.tsx:47-68)**

When `VITE_DEV_RESET_WALLET_ON_BOOT === 'true'` and `localStorage` is available but the `setItem` for the reset key silently fails (e.g., quota exceeded after `removeItem` calls), the guard `localStorage.getItem(storageResetKey)` will keep returning `null` on each boot, causing an infinite reload loop.

The `try/catch` around the storage block catches the error and falls through to the reload, but the `setItem` on line 53 is inside the same try block as the `removeItem` calls on lines 55-57 — if `setItem` throws, the storage still gets cleared but the guard key is never persisted → infinite reload. Consider setting the guard key *before* clearing storage, or using `sessionStorage` for the guard (like `resetControlledDevServiceWorker` does).

---

### 🟡 SHOULD FIX

**4. `value={value ?? ''}` semantic change in InputAmount (InputAmount.tsx:~111)**

The old code was `value={value ? Number(value) : ''}` — this converted truthy string values to numbers, which meant `"0"` would render as `''` (empty). The new code `value={value ?? ''}` means `"0"` now renders as `"0"`. This is arguably *more correct*, but it's a behavioral change that could affect the send flow UX (e.g., after clearing an amount, does `value` become `"0"` or `undefined`?). Verify this doesn't regress the "clear amount → type new amount" flow.

**5. Haptic fires on every dropdown open/close, even programmatic (Form.tsx:~869)**

```tsx
onOpenChange={(open) => {
  hapticLight()
  setShowAssetSelector(open)
}}
```

This fires `hapticLight()` unconditionally — including when the menu closes due to outside click, escape key, or selecting an item (which triggers its own haptic via `DropdownMenuItem`). Users get double-haptic on selection. Consider only firing haptic on open (`if (open) hapticLight()`).

**6. Button `disabled || loading` guard is now double-enforced (Button.tsx:83-84)**

The native `<button disabled={disabled}>` already prevents click events when `disabled` is true. The added `if (disabled || loading) return` in `handleClick` is fine as defense-in-depth for the `loading` case (since `loading` doesn't set the native `disabled`). But note: `aria-busy={loading}` doesn't prevent clicks — you might want `aria-disabled={loading || undefined}` too, so assistive tech knows the button is non-interactive during loading.

---

### 🟢 OBSERVATIONS (no action needed)

- The dev mnemonic auto-init path in `wallet.tsx` correctly prioritizes `devMnemonic` over `devNsec` and calls `initWallet({ mnemonic })` which goes through `MnemonicIdentity.fromMnemonic()` — no protocol concerns here.
- The `getDevWalletStorageResetKey` djb2 hash is fine for a dev-only cache-bust key.
- The `useEffect` dependency array in wallet.tsx:265 now includes all referenced values — good fix.
- CSS changes are purely cosmetic (styling the input shells, asset selector, dropdown menu). No functional concerns.
- `InputContainer` replacing `<Shadow>` with `<div className='input-shell'>` is a clean refactor.
- No protocol-critical code (VTXOs, signing, forfeit, rounds, exits) is touched in this PR.

### Summary

The PR is a solid UI refactor. The e2e test breakage (#1) and the component-inside-render (#2) are the blockers. The infinite reload edge case (#3) is a real bug in dev tooling. The rest are quality improvements.

ghost (changes requested):
## Code Review — arkade-os/wallet#629

Reviewed the full diff, read every changed file in context, checked cross-repo impact, git history, and test coverage.

### 🔴 MUST FIX

**1. E2E tests will break — `data-testid` format changed (Form.tsx:913)**

The asset option `data-testid` changed from `asset-${asset.ticker.toLowerCase()}-option` to `asset-${asset.assetId}-option`. The e2e tests at `src/test/e2e/send.test.ts:157,193` select `asset-tst-option` (by ticker). After this PR, the test ID will be something like `asset-a1b2c3...f4-option` (a hex asset ID), so both e2e test cases will fail.

Either revert to the ticker-based test ID or update the e2e tests to use the new format.

**2. \`AssetIcon\` is defined as a component *inside* the render body (Form.tsx:~712)**

\`\`\`tsx
const AssetIcon = ({ asset }: { asset: AssetOption | null }) => { ... }
\`\`\`

This is declared inside \`SendForm()\`, which means React creates a **new component type on every render**. Every re-render (keystroke in amount field, etc.) will unmount and remount every \`<AssetIcon>\`, destroying DOM state and killing any CSS transitions on icons. It also defeats React's reconciliation.

Move \`AssetIcon\` outside \`SendForm\` (or at minimum, wrap it in \`useMemo\`/extract to module scope). It only depends on \`asset\`, which can be a prop.

**3. \`resetDevWalletStorage\` infinite reload risk (index.tsx:47-68)**

When \`VITE_DEV_RESET_WALLET_ON_BOOT === 'true'\` and \`localStorage\` is available but the \`setItem\` for the reset key silently fails (e.g., quota exceeded after \`removeItem\` calls), the guard \`localStorage.getItem(storageResetKey)\` will keep returning \`null\` on each boot, causing an infinite reload loop.

The \`try/catch\` around the storage block catches the error and falls through to the reload, but the \`setItem\` on line 53 is inside the same try block as the \`removeItem\` calls on lines 55-57 — if \`setItem\` throws, the storage still gets cleared but the guard key is never persisted → infinite reload. Consider setting the guard key *before* clearing storage, or using \`sessionStorage\` for the guard (like \`resetControlledDevServiceWorker\` does).

---

### 🟡 SHOULD FIX

**4. \`value={value ?? ''}\` semantic change in InputAmount (InputAmount.tsx:~111)**

The old code was \`value={value ? Number(value) : ''}\` — this converted truthy string values to numbers, which meant \`"0"\` would render as \`''\` (empty). The new code \`value={value ?? ''}\` means \`"0"\` now renders as \`"0"\`. This is arguably *more correct*, but it's a behavioral change that could affect the send flow UX (e.g., after clearing an amount, does \`value\` become \`"0"\` or \`undefined\`?). Verify this doesn't regress the "clear amount → type new amount" flow.

**5. Haptic fires on every dropdown open/close, even programmatic (Form.tsx:~869)**

\`\`\`tsx
onOpenChange={(open) => {
  hapticLight()
  setShowAssetSelector(open)
}}
\`\`\`

This fires \`hapticLight()\` unconditionally — including when the menu closes due to outside click, escape key, or selecting an item (which triggers its own haptic via \`DropdownMenuItem\`). Users get double-haptic on selection. Consider only firing haptic on open (\`if (open) hapticLight()\`).

**6. Button \`disabled || loading\` guard is now double-enforced (Button.tsx:83-84)**

The native \`<button disabled={disabled}>\` already prevents click events when \`disabled\` is true. The added \`if (disabled || loading) return\` in \`handleClick\` is fine as defense-in-depth for the \`loading\` case (since \`loading\` doesn't set the native \`disabled\`). But note: \`aria-busy={loading}\` doesn't prevent clicks — you might want \`aria-disabled={loading || undefined}\` too, so assistive tech knows the button is non-interactive during loading.

---

### 🟢 OBSERVATIONS (no action needed)

- The dev mnemonic auto-init path in \`wallet.tsx\` correctly prioritizes \`devMnemonic\` over \`devNsec\` and calls \`initWallet({ mnemonic })\` which goes through \`MnemonicIdentity.fromMnemonic()\` — no protocol concerns here.
- The \`getDevWalletStorageResetKey\` djb2 hash is fine for a dev-only cache-bust key.
- The \`useEffect\` dependency array in wallet.tsx:265 now includes all referenced values — good fix.
- CSS changes are purely cosmetic (styling the input shells, asset selector, dropdown menu). No functional concerns.
- \`InputContainer\` replacing \`<Shadow>\` with \`<div className='input-shell'>\` is a clean refactor.
- No protocol-critical code (VTXOs, signing, forfeit, rounds, exits) is touched in this PR.

### Summary

The PR is a solid UI refactor. The e2e test breakage (#1) and the component-inside-render (#2) are the blockers. The infinite reload edge case (#3) is a real bug in dev tooling. The rest are quality improvements.

---
*🤖 Reviewed by Arkana*

ghost (approved):
## Review: feat(send): align send flow with redesign prototype

**Verdict: Approve** — no protocol-critical code touched. This is UI/UX only (send form styling, asset selector, dev tooling). No VTXO handling, transaction signing, forfeit paths, or exit logic affected.

---

### Issues to address (non-blocking)

**1. E2E test breakage — `data-testid` format changed** ⚠️
`src/screens/Wallet/Send/Form.tsx` — the asset option test ID changed from ticker-based to assetId-based:
```
// Before:  data-testid={`asset-${asset.ticker.toLowerCase()}-option`}
// After:   data-testid={`asset-${asset.assetId}-option`}
```
But `src/test/e2e/send.test.ts:157` and `:193` reference `asset-tst-option` (ticker-based). These e2e tests will fail. Either update the tests or keep the old test ID format.

**2. `AssetIcon` defined inside render body** (minor perf)
`src/screens/Wallet/Send/Form.tsx:~710` — `AssetIcon` is declared as a function inside `SendForm()`, so it's recreated every render. This forces React to unmount/remount the icon subtree on each render cycle instead of reconciling. Extract it to module scope (it only needs `asset` as a prop, no closure over component state).

---

### Observations (no action needed)

- **Button.tsx guard is a good fix** — adding `if (disabled || loading) return` before `hapticTap()`/`onClick()` prevents clicks during loading state where `disabled` isn't set on the DOM element. The `aria-busy` addition is correct.

- **InputAmount value fix is correct** — `value={value ?? ''}` preserves `0` as a valid input value, whereas the old `value ? Number(value) : ''` treated `0` as falsy and cleared the field.

- **Dev mnemonic support is clean** — `VITE_DEV_MNEMONIC` is properly guarded behind `import.meta.env.DEV`, `initWallet` already accepts `{ mnemonic }`, and the `resetDevWalletStorage` function has adequate guards (`DEV` + `VITE_DEV_RESET_WALLET_ON_BOOT=true` + per-credential idempotency key).

- **No cross-repo impact** — `InputContainer`, `InputAmount`, `Button` are wallet-internal. `Shadow` is only removed from `InputContainer`, not deleted. No downstream consumers.

- **CSS `!important` proliferation** in dark-mode overrides is a code smell but not a blocker.

🤖 Reviewed by Arkana

ghost (approved):
## Review: feat(send): align send flow with redesign prototype

**Verdict: Approve** — no protocol-critical code touched. This is UI/UX only (send form styling, asset selector, dev tooling). No VTXO handling, transaction signing, forfeit paths, or exit logic affected.

---

### Issues to address (non-blocking)

**1. E2E test breakage — `data-testid` format changed** ⚠️
`src/screens/Wallet/Send/Form.tsx` — the asset option test ID changed from ticker-based to assetId-based:
```
// Before:  data-testid={`asset-${asset.ticker.toLowerCase()}-option`}
// After:   data-testid={`asset-${asset.assetId}-option`}
```
But `src/test/e2e/send.test.ts:157` and `:193` reference `asset-tst-option` (ticker-based). These e2e tests will fail. Either update the tests or keep the old test ID format.

**2. `AssetIcon` defined inside render body** (minor perf)
`src/screens/Wallet/Send/Form.tsx:~710` — `AssetIcon` is declared as a function inside `SendForm()`, so it's recreated every render. This forces React to unmount/remount the icon subtree on each render cycle instead of reconciling. Extract it to module scope (it only needs `asset` as a prop, no closure over component state).

---

### Observations (no action needed)

- **Button.tsx guard is a good fix** — adding `if (disabled || loading) return` before `hapticTap()`/`onClick()` prevents clicks during loading state where `disabled` isn't set on the DOM element. The `aria-busy` addition is correct.

- **InputAmount value fix is correct** — `value={value ?? ''}` preserves `0` as a valid input value, whereas the old `value ? Number(value) : ''` treated `0` as falsy and cleared the field.

- **Dev mnemonic support is clean** — `VITE_DEV_MNEMONIC` is properly guarded behind `import.meta.env.DEV`, `initWallet` already accepts `{ mnemonic }`, and the `resetDevWalletStorage` function has adequate guards (`DEV` + `VITE_DEV_RESET_WALLET_ON_BOOT=true` + per-credential idempotency key).

- **No cross-repo impact** — `InputContainer`, `InputAmount`, `Button` are wallet-internal. `Shadow` is only removed from `InputContainer`, not deleted. No downstream consumers.

- **CSS `!important` proliferation** in dark-mode overrides is a code smell but not a blocker.

🤖 Reviewed by Arkana

ghost (approved):
## Re-review after force push (22d086b1)

All five findings from my previous review are addressed:

| # | Finding | Resolution |
|---|---------|------------|
| 1 | E2E `data-testid` format breakage | ✅ Kept ticker-based format — matches `src/test/e2e/send.test.ts:157,193` |
| 2 | `AssetIcon` defined inside render body | ✅ Extracted to module scope (Form.tsx:~72) |
| 3 | `resetDevWalletStorage` infinite reload | ✅ Guard key `setItem` now precedes `removeItem` calls; `catch` block no longer falls through to reload |
| 4 | Double haptic on dropdown | ✅ `if (open) hapticLight()` guard added |
| 5 | Missing `aria-disabled` on loading button | ✅ `aria-disabled={loading || undefined}` added |

No new concerns. No protocol-critical code touched (VTXOs, signing, forfeit, rounds, exits are untouched). UI-only changes — safe to merge with human approval.

🤖 Reviewed by Arkana

tiero:
I would wait to incorporate the contact/friends functionalities, to redo the send. Also woudl be nice to divide the step of selection "who send money to" and "how much"

sahilc0:
> I would wait to incorporate the contact/friends functionalities, to redo the send. Also woudl be nice to divide the step of selection "who send money to" and "how much"

why wait? this is just a light ui cleanup/polish

ghost (commented):
## Re-review after merge commits (471cc3d → a521ed1)

Two merge commits bringing the base branch (`wt-wallet-redesign-followups-20260520`) into this PR. No new feature code — just a rebase-via-merge.

### New finding from merge

**Duplicate `VITE_DEV_MNEMONIC` declaration** — `src/vite-env.d.ts`
The merge introduced a duplicate `readonly VITE_DEV_MNEMONIC?: string` in the `ImportMetaEnv` interface. Both this PR and the base branch added it, and the merge kept both. Not a runtime bug (TS takes the last declaration for duplicate interface members), but it's a merge artifact that should be cleaned up.

### Changes from base branch merge (look correct)

- **`wallet.tsx`**: `devAutoInitCredential` variable extracted, useEffect deps expanded from `[aspInfo.url, initialized, devAutoInitFailed]` to include `isDevAutoInit`, `devAutoInitCredential`, `devMnemonic`, `devNsec`. This is more correct (satisfies exhaustive-deps) and harmless since all the new deps are stable `import.meta.env` reads that won't change between renders.

No protocol-critical code touched. Previous approval stands — all five original findings remain addressed.

🤖 Reviewed by Arkana

PR #630, opened 2026-05-21
feat(wallet): add bitcoin detail page

sahilc0:
## Summary
- Add a bitcoin-only wallet detail page opened from the bitcoin asset row on the redesigned home screen.
- Bring over the draft PR's Liveline-based price chart treatment with current bitcoin price, percentage movement badge, range tabs, holdings balance/value, and bitcoin-filtered recent activity.
- Share the existing Vaul-backed swap coming-soon drawer between the home quick action and the bitcoin detail Swap action.

- `src/screens/Wallet/BitcoinDetail.tsx` adds the bitcoin detail screen and chart/price data handling.
- `src/screens/Wallet/AssetsSection.tsx` routes the bitcoin row into the detail page.
- `src/components/SwapComingSoonSheet.tsx` centralizes the PR 627 swap coming-soon drawer.
- `src/providers/navigation.tsx` registers the bitcoin detail route under the wallet tab.
- `src/index.css` adds the detail page styles ported from the draft approach.
- `package.json`, `bun.lock`, and `pnpm-lock.yaml` add `liveline`.

## Testing
- `bunx vitest run src/test/screens/wallet/index.test.tsx`
- `bun run lint`
- `bun run build`
- Playwright smoke check against `wt-bitcoin-detail-page-20260521.arkade.localhost`: clicked the bitcoin row, confirmed the detail page rendered with a chart canvas and Receive/Send/Swap/Scan actions, and verified Swap opens the coming-soon drawer.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -30,6 +30,7 @@
         "framer-motion": "^12.34.3",
         "input-otp": "^1.4.2",
         "light-bolt11-decoder": "^3.2.0",
+        "liveline": "^0.0.7",
         "lucide-react": "^0.575.0",
         "nostr-tools": "^2.12.0",
         "qr": "^0.5.2",
@@ -1203,6 +1204,8 @@
     "lightningcss-win32-x64-msvc": ["lightningcss-win32-x64-msvc@1.32.0", "", { "os": "win32", "cpu": "x64" }, "sha512-Amq9B/SoZYdDi1kFrojnoqPLxYhQ4Wo5XiL8EVJrVsB8ARoC1PWW6VGtT0WKCemjy8aC+louJnjS7U18x3b06Q=="],
 
     "lines-and-columns": ["lines-and-columns@1.2.4", "", {}, "sha512-7ylylesZQ/PV29jhEDl3Ufjo6ZX7gCqJr5F7PKrqc93v7fzSymt1BpwEU8nAUXs8qzzvqhbjhK5QZg6Mt/HkBg=="],
+
+    "liveline": ["liveline@0.0.7", "", { "peerDependencies": { "react": ">=18" } }, "sha512-S1Weo1kmefiOCNaW/WFXErqJXlV8A/ldIkIthaOeT57I2ZhqhKLKQDJ1wfabtMSK9vJeU8PeMNHd8Sr8TLByWQ=="],
 
     "locate-path": ["locate-path@6.0.0", "", { "dependencies": { "p-locate": "5.0.0" } }, "sha512-iPZK6eYjbxRu3uB4/WZ3EsEIMJFMqAoopl3R+zuq0UjcAm/MO6KCweDgPfP3elTztoKP3KtnVHxTn2NHBSDVUw=="],
 
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "framer-motion": "^12.34.3",
     "input-otp": "^1.4.2",
     "light-bolt11-decoder": "^3.2.0",
+    "liveline": "^0.0.7",
     "lucide-react": "^0.575.0",
     "nostr-tools": "^2.12.0",
     "qr": "^0.5.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -83,6 +83,9 @@ importers:
       light-bolt11-decoder:
         specifier: ^3.2.0
         version: 3.2.0
+      liveline:
+        specifier: ^0.0.7
+        version: 0.0.7(react@18.3.1)
       lucide-react:
         specifier: ^0.575.0
         version: 0.575.0(react@18.3.1)
@@ -3002,6 +3005,12 @@ packages:
 
   lines-and-columns@1.2.4:
     resolution: {integrity: sha512-7ylylesZQ/PV29jhEDl3Ufjo6ZX7gCqJr5F7PKrqc93v7fzSymt1BpwEU8nAUXs8qzzvqhbjhK5QZg6Mt/HkBg==}
+
+  liveline@0.0.7:
+    resolution: {integrity: sha512-S1Weo1kmefiOCNaW/WFXErqJXlV8A/ldIkIthaOeT57I2ZhqhKLKQDJ1wfabtMSK9vJeU8PeMNHd8Sr8TLByWQ==}
+    engines: {node: '>=18'}
+    peerDependencies:
+      react: '>=18'
 
   locate-path@6.0.0:
     resolution: {integrity: sha512-iPZK6eYjbxRu3uB4/WZ3EsEIMJFMqAoopl3R+zuq0UjcAm/MO6KCweDgPfP3elTztoKP3KtnVHxTn2NHBSDVUw==}
@@ -7098,6 +7107,10 @@ snapshots:
       lightningcss-win32-x64-msvc: 1.32.0
 
   lines-and-columns@1.2.4: {}
+
+  liveline@0.0.7(react@18.3.1):
+    dependencies:
+      react: 18.3.1
 
   locate-path@6.0.0:
     dependencies:
--- a/src/components/AssetCard.tsx
+++ b/src/components/AssetCard.tsx
@@ -33,7 +33,7 @@ export default function AssetCard({
   fiatText,
   onClick,
 }: AssetCardProps) {
-  const assetName = ticker && ticker !== 'BTC' ? ticker : name || truncatedAssetId(assetId) || 'Asset'
+  const assetName = name || truncatedAssetId(assetId) || 'Asset'
   const tokenTick = ticker ?? 'TKN'
   const rawBalance =
     typeof balance === 'bigint'
--- a/src/components/SwapComingSoonSheet.tsx
+++ b/src/components/SwapComingSoonSheet.tsx
@@ -1,0 +1,22 @@
+import Button from './Button'
+import SheetModal from './SheetModal'
+import SwapIcon from '../icons/Swap'
+
+export default function SwapComingSoonSheet({ isOpen, onClose }: { isOpen: boolean; onClose: () => void }) {
+  return (
+    <SheetModal isOpen={isOpen} onClose={onClose}>
+      <div className='swap-coming-soon' data-testid='swap-coming-soon-sheet'>
+        <div className='swap-coming-soon__icon' aria-hidden='true'>
+          <SwapIcon />
+        </div>
+        <div className='swap-coming-soon__copy'>
+          <h2 className='swap-coming-soon__title'>Swaps are coming soon</h2>
+          <p className='swap-coming-soon__description'>
+            We are polishing the swap experience before turning it on here.
+          </p>
+        </div>
+        <Button label='Got it' onClick={onClose} />
+      </div>
+    </SheetModal>
+  )
+}
--- a/src/components/TransactionsList.tsx
+++ b/src/components/TransactionsList.tsx
@@ -109,7 +109,7 @@ const TransactionLine = ({
           const icon = meta?.icon
           const decimals = meta?.decimals ?? 8
           const accountTicker = accountTickerForAssetTicker(ticker)
-          const label = accountTicker ?? meta?.name ?? `${a.assetId.slice(0, 8)}...`
+          const label = accountTicker ?? ticker ?? meta?.name ?? `${a.assetId.slice(0, 8)}...`
           return (
             <FlexRow key={a.assetId} gap='0.375rem' end>
               <TransactionAssetAvatar icon={icon} ticker={accountTicker ?? ticker} assetId={a.assetId} />
--- a/src/index.css
+++ b/src/index.css
@@ -1872,3 +1872,379 @@ html.palette-dark .settings-row--danger .settings-row__label,
 html.palette-dark .settings-row--danger .settings-row__chevron {
   color: color-mix(in srgb, var(--red-400) 82%, var(--fg));
 }
+
+.asset-detail-page {
+  display: flex;
+  flex-direction: column;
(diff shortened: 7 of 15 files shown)


sahilc0:
@bordalix this PR brings over the frontend for the bitcoin detail price/chart experience from the draft approach. We may need the app/backend to serve the actual bitcoin price and historical chart data instead of relying directly on external market APIs from the client. Could you help advise on the right way to serve that data?

ghost (changes requested):
## Code Review — `feat(wallet): add bitcoin detail page`

Overall this is a well-structured UI feature: clean component decomposition, good accessibility (aria roles on tabs, reduced-motion support), and solid refactoring of SwapComingSoonSheet into a shared component. But there's one **crash-level bug** that blocks merge.

---

### 🔴 BUG — `convertFiat` does not exist (runtime crash)

**`src/screens/Wallet/BitcoinDetail.tsx:4` (line 594 in diff)**
```ts
const { convertFiat, fiatDecimals, toFiat } = useContext(FiatContext)
```

`FiatContext` exports `{ toFiat, fromFiat, fiatDecimals, updateFiatPrices }`. There is no `convertFiat` property. Destructuring yields `undefined`.

This `undefined` is then passed through `useBitcoinMarketChartData` → `fetchBitcoinMarketChart` → `fetchCoinbaseBitcoinAllChart` → `fetchCoinbaseBitcoinCandles`, where it's called at **BitcoinDetail.tsx:476** (line 1026 in diff):
```ts
[time * 1000, fiat === Fiats.USD ? close : convertFiat(close, Fiats.USD)]
```

**Impact:** For any user whose fiat currency is **not USD**, selecting the "All" chart range will throw `TypeError: convertFiat is not a function`, crashing the component. USD users are accidentally safe because the ternary short-circuits.

**Fix:** Either add a `convertFiat(amount, fromCurrency)` method to `FiatContext`/`FiatProvider`, or build the conversion inline using the existing `toFiat`/`fromFiat` helpers. Something like:
```ts
const convertFromUSD = (usdAmount: number) => {
  const sats = fromFiat(usdAmount) // if fromFiat were currency-aware...
  return toFiat(sats)
}
```
But note `fromFiat` uses the user's *configured* fiat, not a specified source currency — so you likely need a new helper that takes a source currency parameter, or fetch Coinbase candles in USD and convert via the existing price ratio.

---

### 🟡 Coinbase "All" chart — unbounded parallel requests

**`BitcoinDetail.tsx:442-448` (lines 992-997 in diff)**
```ts
for (let start = ALL_CHART_START_TIME; start < now; start += chunkSecs) {
  requests.push(fetchCoinbaseBitcoinCandles(start, end, fiat, convertFiat, signal))
}
const chunks = await Promise.all(requests)
```

From 2015 to 2026 with 300-day chunks = **~14 parallel requests** to `api.exchange.coinbase.com`. Coinbase's public API rate limit is 10 req/s per IP. This will likely trigger 429s on slower connections or under any existing load.

**Suggestion:** Use `Promise.allSettled` with a concurrency limiter (e.g., batch 3-4 at a time), or fetch sequentially. Also handle 429 responses with a retry/backoff rather than throwing.

---

### 🟡 No caching of chart data

Each time the user taps a range tab, `useBitcoinMarketChartData` fires a fresh fetch. Toggling between "1D" and "1W" repeatedly hammers CoinGecko/Coinbase on every tap.

**Suggestion:** Cache fetched data per `(fiat, windowSecs)` key in a ref or module-level Map, at least for the component's lifetime. The data doesn't go stale in seconds.

---

### 🟡 CoinGecko "All" fallback is misleading

**`BitcoinDetail.tsx:600` (line 1151 in diff)**
```ts
function coinGeckoDaysForWindow(windowSecs: number): string[] {
  if (isAllChartWindow(windowSecs)) return ['365']
```

If Coinbase "All" fails (which it will for non-USD users per the bug above), the CoinGecko fallback requests 365 days but labels the tab "All". Users see ~1 year of data while the UI says "All time." Either fetch `max` from CoinGecko or label the fallback honestly.

---

### 🟡 `balance` passed to `BigInt()` without guard

**`BitcoinDetail.tsx:259` (line 809 in diff)**
```ts
<span className='asset-detail-holding-amount'>{formatAssetAmount(BigInt(balance), 8)}</span>
```

`balance` is typed as `number` from `WalletContext`. If it's ever `NaN`, fractional, or `Infinity` (e.g., during loading or error states), `BigInt()` throws. The rest of the codebase treats balance as an integer satoshi count, so this is low-risk, but a guard like `BigInt(Math.floor(balance || 0))` is cheap insurance.

---

### 🟢 Minor / Nits

1. **376 lines of CSS in `index.css`** — This is a growing monolith. Not blocking, but consider splitting per-screen CSS or using CSS modules before this file becomes unmanageable.

2. **`useTokenColor` runs once** (BitcoinDetail.tsx:355, line 905 in diff) — The `useEffect` has `[token]` as dependency, but `token` is a constant string `'--orange-500'`. This means it resolves the CSS custom property once on mount. If the user switches themes, the chart color won't update until the component remounts. Consider adding `config.theme` to the dependency array or listening for theme changes.

3. **Test coverage is minimal** — The new test only covers navigation *to* the detail page (clicking the BTC row). There are no tests for `BitcoinDetail` itself: chart data fetching, range tab switching, delta calculation, action button handlers, error states. The helper functions (`calculateDelta`, `smoothChartData`, `bucketAveragePoints`, `coinGeckoDaysForWindow`) are pure and easily unit-testable.

---

### ✅ What looks good

- Clean extraction of `SwapComingSoonSheet` into a shared component
- Proper `AbortController` cleanup in the data-fetching effect
- Good accessibility: `role="tablist"`, `aria-selected`, `aria-hidden`, `aria-label`
- `prefers-reduced-motion` respected throughout
- `ResizeObserver` feature detection before rendering Liveline
- `contain: content` on the page root for paint performance
- Navigation registration is clean and follows existing patterns

---

**Verdict: Request changes.** The `convertFiat` bug is a crash for all non-USD users on the "All" tab. Fix that, and this is close to shippable.

ghost (approved):
## Follow-up Review — New Commit `2c53873`

All six findings from my previous review have been addressed cleanly:

| Finding | Fix |
|---|---|
| 🔴 `convertFiat` crash (non-USD users) | Removed entirely. Coinbase "All" path now gated on `fiat === Fiats.USD`; non-USD falls through to CoinGecko. No more phantom `undefined` function call. |
| 🟡 Unbounded parallel Coinbase requests | Sequential `await` in loop — no more 14-request burst. |
| 🟡 No chart caching | Module-level `Map<string, LivelinePoint[]>` keyed by `fiat:windowSecs`. |
| 🟡 CoinGecko "All" returned 365 days | Now returns `'max'`. |
| 🟡 `BigInt(balance)` unguarded | `safeBalance` via `Number.isFinite` + `Math.floor` + `Math.max(0, ...)`. |
| 🟢 `useTokenColor` theme reactivity | `theme` added as dependency + parameter. |

**No new issues introduced.** The fix commit is tight — 33 additions, 32 deletions, all in `BitcoinDetail.tsx`.

Minor note for future: the chart cache has no TTL, so data can go stale if the user stays on the page for a long session. Not blocking — just worth knowing.

✅ Approved. No protocol-critical code touched (this is a pure UI feature — chart, price display, navigation).

sahilc0:
@bordalix fixed the chart hover drift João flagged. Liveline is now paused while the user is hovering/scrubbing the bitcoin chart, so holding the cursor/thumb at one point keeps the tooltip value fixed instead of advancing with the live time window. Added a regression test for pause-on-hover/resume-on-leave.

ghost (approved):
## Follow-up Review — Commit `1978776`

Single-line addition: `paused={chartInteracting}` on the `<Liveline>` component, plus a new test.

**What changed:**
- `BitcoinDetail.tsx:208` — passes existing `chartInteracting` state as `paused` prop to freeze the chart animation while the user is hovering/scrubbing. Clean use of already-wired pointer event handlers (`setChartScrubbing`).
- `bitcoin-detail.test.tsx` — new test covering the pause-on-hover behavior. Mocks `Liveline`, `ResizeObserver`, and `fetch` correctly. Asserts `data-paused` toggles on `pointerEnter`/`pointerLeave`.

**Verdict:** No protocol-critical code touched. Pure UI polish. The interaction state was already managed; this just threads it to the chart library. Test coverage is appropriate.

✅ LGTM — no issues found in the new commit.

ghost (approved):
## Follow-up Review — Commit `1978776`

Single-line addition: `paused={chartInteracting}` on the `<Liveline>` component, plus a new test.

**What changed:**
- `BitcoinDetail.tsx:208` — passes existing `chartInteracting` state as `paused` prop to freeze the chart animation while the user is hovering/scrubbing. Clean use of already-wired pointer event handlers (`setChartScrubbing`).
- `bitcoin-detail.test.tsx` — new test covering the pause-on-hover behavior. Mocks `Liveline`, `ResizeObserver`, and `fetch` correctly. Asserts `data-paused` toggles on `pointerEnter`/`pointerLeave`.

**Verdict:** No protocol-critical code touched. Pure UI polish. The interaction state was already managed; this just threads it to the chart library. Test coverage is appropriate.

✅ LGTM — no issues found in the new commit.

ghost (approved):
## Follow-up Review — Commits `f2eedfd`, `8c7ebb6`

Two small production fixes and a batch of test hardening. No protocol-critical code touched.

### Production changes (both clean) ✅

**`src/components/AssetCard.tsx:36`** — Removed the ticker-as-name special-case (`ticker && ticker !== 'BTC' ? ticker : ...`). Now consistently uses `name || truncatedAssetId || 'Asset'`. All seven callsites already pass `name`, so no display regression.

**`src/components/TransactionsList.tsx:112`** — Added `ticker` as a fallback in the label chain: `accountTicker ?? ticker ?? meta?.name ?? truncated`. Correct — for assets where `accountTickerForAssetTicker()` returns `undefined` but a raw ticker exists, this prevents falling through to the less-readable metadata name or truncated ID.

### Test changes ✅

- Increased flaky timeouts (`4000` → `10000`ms)
- Made balance assertions deterministic: computes expected balances from `totalSent = amount + fees` instead of scraping the DOM with regex — much more robust
- Removed `prettyNumber` formatting from assertions to match actual rendered output
- Fixed selectors for the updated transaction display format (dropped `- $` prefix)

No issues found. Ship it.

ghost (commented):
## Follow-up Review — Commits `97f0aa7`, `9d782bd`, `48cdbad`, `732301f`

Good refactor: replaces direct Coinbase/CoinGecko calls with a backend worker proxy (`fetchHistoricalMarketData`). Removes ~100 lines of client-side chunked fetching. Not protocol-critical — this is purely market-data UI. Three findings:

---

### 🟡 Missing `AbortSignal` passthrough — requests can't be cancelled

**`src/lib/marketData.ts:28-33`**

`fetchHistoricalMarketData` doesn't accept or forward an `AbortSignal`. The caller in `BitcoinDetail.tsx:384-405` still creates an `AbortController` and checks `signal.aborted` after resolution, but the underlying `fetch()` is never abortable.

If a user rapidly switches chart tabs (1H → 1D → 1W → 1M), all 4 requests fire and complete — none are cancelled. The abort check after `.then()` prevents stale data from rendering, but the requests themselves are wasted. On slow connections this means unnecessary network load and delayed rendering of the final tab.

**Suggested fix:**
```ts
export const fetchHistoricalMarketData = async (
  secs: number, fiat: Fiats, signal?: AbortSignal
): Promise<LivelinePoint[]> => {
  // ...
  const resp = await fetch(`${host}?${params.toString()}`, { signal })
  // ...
}
```

---

### 🟡 No `response.ok` check — silent failure on worker errors

**`src/lib/marketData.ts:31-32`**

```ts
const resp = await fetch(`${host}?${params.toString()}`)
const data = await resp.json()
```

If the worker returns a 500/502/404, `resp.json()` will either throw on non-JSON bodies (uncaught, bubbles as a generic parse error) or parse an error object that fails validation and silently returns `[]`. The old CoinGecko/Coinbase code explicitly checked `response.ok` and threw descriptive errors.

**Suggested fix:**
```ts
if (!resp.ok) throw new Error(`Market data fetch failed: ${resp.status}`)
```

---

### 🟡 `CHART_WINDOWS` duplicated — drift risk

**`src/lib/marketData.ts:12-19`** and **`src/screens/Wallet/BitcoinDetail.tsx:30-37`**

Both files define their own `CHART_WINDOWS` array. The `secs` values must match for `secsToPeriod()` to return the correct label. If someone later adds or reorders a window in one file but not the other, it silently falls back to `'oneDay'`.

Consider exporting the canonical list from `marketData.ts` and importing it in `BitcoinDetail.tsx`, or at minimum exporting the `secsToPeriod` mapping so there's one source of truth for window → API period.

---

**Not blocking.** Default chart changed from 1H to 1W — makes sense for a detail page. Console.log removed. Worker URL has a TODO to move off personal subdomain (tracked). All clean. The three findings above are quality improvements, not blockers.

ghost (approved):
## Follow-up Review — Commit `1820528`

Addresses both findings from my last review:

1. ✅ **AbortSignal passthrough** — `signal?: AbortSignal` added to `fetchHistoricalMarketData` and forwarded to `fetch()`. Rapid tab-switching now cancels in-flight requests properly.
2. ✅ **`response.ok` check** — Worker errors now throw a descriptive `Error` instead of silently falling through to JSON parse failures.

Rename `CHART_WINDOWS` → `PERIOD_WINDOWS` in `marketData.ts` is a clarity improvement. The remaining duplication between `PERIOD_WINDOWS` (API labels) and `CHART_WINDOWS` in `BitcoinDetail.tsx` (display labels) is fine — they serve different purposes and the `secs` values match.

No new issues. LGTM.

PR #632, opened 2026-05-27
feat(boltz): DevMode button to restore all swaps from Boltz

pietro909:
Surface restoreSwaps() as a discoverable, DevMode-gated button in Boltz settings, sitting between the API URL row and the recovery section. 

Replaces the hidden 5-tap gesture on the API URL label. 

The underlying restore is merge-only, so existing swaps and their preimages/refund secrets are never overwritten.

## Summary by CodeRabbit

* **New Features**
  * Introduced an explicit "Restore all swaps from Boltz" button in developer mode, replacing the previous hidden tap-counter mechanism for a more intuitive restoration workflow.

* **Improvements**
  * Enhanced restoration process with clearer status feedback and error reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/wallet/pull/632?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/screens/Apps/Boltz/Settings.tsx
+++ b/src/screens/Apps/Boltz/Settings.tsx
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useState } from 'react'
+import { useContext, useState } from 'react'
 import type { BoltzSubmarineSwap, SubmarineRecoveryInfo } from '@arkade-os/boltz-swap'
 import Padded from '../../../components/Padded'
 import Header from '../../../components/Header'
@@ -11,6 +11,7 @@ import Button from '../../../components/Button'
 import ErrorMessage from '../../../components/Error'
 import WarningBox from '../../../components/Warning'
 import { SwapsContext } from '../../../providers/swaps'
+import { DevModeContext } from '../../../providers/devMode'
 import { useToast } from '../../../components/Toast'
 import { consoleError } from '../../../lib/logs'
 import { extractError } from '../../../lib/error'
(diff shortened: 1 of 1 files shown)


ghost (approved):
## Review: feat(boltz): DevMode button to restore all swaps from Boltz

**Scope**: Single file (`src/screens/Apps/Boltz/Settings.tsx`), 44 additions / 25 deletions. UI-only change — no protocol logic touched.

### What changed

1. Removes the hidden 5-tap easter-egg on the API URL label that triggered `restoreSwaps()`.
2. Replaces it with an explicit "Restore all swaps from Boltz" button, gated behind `DevModeContext.devMode`.
3. Converts the `useEffect`/counter pattern into a proper async click handler (`handleRestoreAll`) with loading state.

### Findings

**✅ Correct and clean — no issues found.**

**Protocol safety**: `restoreSwaps()` in the swaps provider (`src/providers/swaps.tsx:312`) is merge-only — it fetches from Boltz, deduplicates against `getSwapHistory()`, and only writes new entries. No preimages or refund secrets are overwritten. This PR does not change that function at all; it only changes how it's triggered in the UI. **Not protocol-critical.**

**Specifics reviewed:**

- **`Settings.tsx:37-50` — `handleRestoreAll`**: Properly guards on `!arkadeSwaps` before calling `restoreSwaps()`. Sets `restoring=true` before the call, clears it in `finally`. Error path logs and surfaces the error to the user. The guard is slightly redundant with the one inside `restoreSwaps()` itself (`swaps.tsx:313`), but it's the right defensive pattern since it lets the handler show a user-facing message ("not ready yet") rather than silently returning 0.

- **`Settings.tsx:155-170` — DevMode gate**: Correctly reads `devMode` from `DevModeContext`. The button is `disabled={!arkadeSwaps || restoring}`, preventing double-clicks and premature invocation. Good.

- **Removed `useEffect` (old lines 36-48)**: The old pattern had a subtle issue — `restoreSwaps` was in the dependency array and the effect fired when `counter === 5`, but there was no mechanism to reset the counter, so the effect could only fire once per mount. The new explicit handler is strictly better.

- **Removed `useEffect` import**: Confirmed no other `useEffect` calls remain in the component. Clean.

- **No cross-repo impact**: No API, type, or interface changes. `DevModeContext` is already used elsewhere (`LoadingLogo.tsx`, `Settings/Advanced.tsx`). `restoreSwaps` signature unchanged.

- **No test coverage for this component**: There are no existing tests for `AppBoltzSettings` (checked `src/test/`). This is pre-existing — not introduced by this PR. Would be nice to add eventually but not blocking.

### Verdict

Clean refactor. Replaces a janky hidden gesture with a proper, guarded, DevMode-only button. No security, correctness, or regression concerns.

LGTM ✅

PR #634, opened 2026-05-27
feat(wallet): add redesigned swap flow

sahilc0:
## Dependency

Depends on #627. This PR is intentionally based on `wt-wallet-redesign-followups-20260520`, the open PR #627 branch, so review should focus on the swap-specific diff layered on top of that wallet redesign work.

@bordalix: explicit heads-up that this should be reviewed as a dependent PR on #627, not as an independent branch against `master`.

## Summary

- Adds the redesigned wallet swap flow using the draft swap UI direction: asset selection, amount entry, receive asset drawer, review drawer, quote details, and success splash.
- Replaces the old swap placeholder from wallet quick actions and bitcoin detail with the real swap UI route.
- Adds swap-specific states: search empty state, unavailable state when there are not enough supported assets, insufficient-balance badge with haptic/shake feedback, quote shimmer placeholders, and confirm error handling.
- Adds temporary dev-only USD/CHF test assets behind `VITE_DEV_SWAP_TEST_ASSETS` so swap UI behavior can be tested without touching real bitcoin balances.
- Updates recent activity and transaction details so prototype swaps render as swaps with route, from/to amounts, and pending/failed/completed-ready metadata.
- Shares the redesigned success splash across swap, send, and receive success flows.

## Affected areas

- `src/screens/Wallet/Swap/Index.tsx` - new swap flow, drawers, validation, amount animation, quote/review states.
- `src/hooks/usePortfolioFiat.ts` and `src/lib/devSwapTestAssets.ts` - fiat account aggregation and dev-only swap testing balances.
- `src/providers/wallet.tsx` and `src/lib/types.ts` - prototype swap metadata and balance deltas.
- `src/components/TransactionsList.tsx` and `src/screens/Wallet/Transaction.tsx` - swap-aware activity and detail rendering.
- `src/components/WalletSuccessSplash.tsx`, send success, receive success - shared success treatment.
- `src/index.css` - swap UI, animation, drawer, state, activity, and success styling.

## Testing

- `bun run lint`
- `bun run format:check`
- `bun run test:unit -- src/test/screens/wallet/swap.test.tsx`
- `bun run build`
- Browser verified at the Portless app URL:
  - swap opens from wallet quick action
  - default amount starts at `0`
  - search empty state appears
  - insufficient balance badge is centered and does not shift the amount
  - add/delete digit animations run
  - review and confirm flow completes
  - recent activity shows swap route and from/to amounts

## Notes

The temporary USD/CHF assets are deliberately isolated behind a dev-only environment flag and should be easy to remove when the real asset/swap backend is ready.


sahilc0:
@bordalix explicit dependency note: this PR depends on #627 and is based on `wt-wallet-redesign-followups-20260520`. Please review it as the swap UI layer on top of PR #627, not as an independent diff against `master`.

ghost (changes requested):
## Code Review — arkade-os/wallet#634

Reviewed the full diff against base branch `wt-wallet-redesign-followups-20260520`. The swap UI itself is well-crafted — good animations, accessibility labels, reduced-motion support, clean component decomposition. But there are real correctness and safety issues in the state management layer that need fixing before merge.

**No cross-repo breakage risk** — all modified types/interfaces are wallet-internal.

---

### 🔴 MUST FIX

#### 1. Prototype txs leak into settlement logic — `src/providers/wallet.tsx:889`

```tsx
txs: [...prototypeTxs, ...txs],
```

Every consumer of `WalletContext.txs` now receives prototype swap txs mixed with real ones. This includes `Transaction.tsx:97` which calls `settlePreconfirmed()` — if a user taps a prototype swap tx in the detail view and hits settle, it will attempt settlement on a fake tx. The `settlePreconfirmed` call doesn't filter by tx type.

**Fix**: Either add an `isPrototype: boolean` field to `Tx` and guard all settlement/explorer code paths, or expose `prototypeTxs` as a separate context property and merge only at the UI render layer (`TransactionsList`, `Transaction`).

#### 2. `roundTxid` collision — `src/providers/wallet.tsx:849`

```tsx
roundTxid: `prototype-swap-${createdAt}`,
```

`createdAt` is `Math.floor(Date.now() / 1000)` — **seconds**. Two rapid swaps within the same second produce identical `roundTxid` values. This causes:
- React key collisions in `TransactionsList.tsx:182` (key is built from `roundTxid`)
- Unpredictable virtualizer behavior (duplicate keys → DOM corruption)

**Fix**: Use `Date.now()` (milliseconds) or `crypto.randomUUID()` for uniqueness.

#### 3. `as any` type bypass on Asset — `src/providers/wallet.tsx:840-843`

```tsx
assets: [{ assetId: swap.toAssetId, amount: swap.toAmount } as any],
```

The `Asset` type from `@arkade-os/sdk` likely has required fields beyond `assetId` and `amount`. Casting to `any` silently drops validation. If any downstream code accesses other Asset fields on these prototype objects, it will get `undefined` and fail silently — dangerous in a wallet.

**Fix**: Define a proper subset type or satisfy the full `Asset` interface. If you can't (because it's too coupled to SDK internals), at minimum document the incomplete fields and add runtime guards in code that reads `tx.assets`.

#### 4. Hardcoded fake quote details — `src/screens/Wallet/Swap/Index.tsx:~line 845-850`

```tsx
<MetricRow label='Price impact' value='0.04%' loading={loading} />
<MetricRow label='Network cost' value='$0.00' loading={loading} />
<MetricRow label='Arrival' value='About 12 seconds' loading={loading} />
```

These present as real data in the review drawer. A user (or QA tester) could believe these are computed values. Even in a prototype, showing fabricated financial metrics in a confirmation screen is misleading.

**Fix**: Either compute these from the quote, mark them clearly as placeholder (e.g., "~0.04% (est.)"), or hide them until real quote data is available.

#### 5. `SwapAmountInfo` hardcodes USD — `src/components/TransactionsList.tsx:209`

```tsx
{from ? `- ${from}` : fiat ? prettyFiatAmount(fiat, Fiats.USD) : ''}
```

This ignores the user's configured fiat currency (`config.fiat`). Users with EUR/CHF/JPY will see USD amounts in their activity feed.

**Fix**: Pass the user's fiat preference from config context, or store the fiat currency alongside the amount in `prototypeSwap`.

---

### 🟡 SHOULD FIX

#### 6. Floating-point precision in `toRawAmount` — `src/screens/Wallet/Swap/Index.tsx:~line 927`

```tsx
return BigInt(Math.max(0, Math.round(units * 10 ** decimals)))
```

For BTC (decimals=8), if `units = 0.00000001`, then `units * 10**8 = 1.0` — fine. But for larger amounts like `units = 210000.12345678`, `units * 10**8 = 21000012345678` which is within safe integer range. However, intermediate floating-point operations in `buildPrototypeSwapExecution` (dividing by `fromUsd`, multiplying by `toUsd`) can accumulate enough error to round to wrong satoshi values.

**Consider**: Use a decimal library (you already have `Decimal` in fiat.tsx) for the amount conversion chain, or at minimum document the precision bounds.

#### 7. `convertFiat` double-conversion precision loss — `src/providers/fiat.tsx:55-57`

```tsx
const convertFiat = (amount: number, from: Fiats, to = config.fiat) => {
  if (from === to) return amount
  return toFiatCurrency(fromFiatCurrency(amount, from), to)
}
```

This converts `fiat → sats → fiat`, losing precision at each step. For large values or volatile exchange rates, this can drift noticeably.

**Consider**: Pre-compute cross-rates directly (e.g., `CHF/USD = prices.chf / prices.usd`) to avoid the sats round-trip.

#### 8. No cleanup of prototype state — `src/providers/wallet.tsx`

`prototypeTxs` and `prototypeAssetBalanceDeltas` are never cleaned up. They persist for the entire session. After 12 prototype swaps, they're capped (`slice(0, 12)`), but the balance deltas accumulate indefinitely. Over a long session with many test swaps, balances could drift significantly from reality.

**Consider**: Add a reset mechanism (e.g., clear on wallet reload, or age out after N minutes).

#### 9. `Number(rawAmount)` for bigint — `src/hooks/usePortfolioFiat.ts:~line 162`

```tsx
const normalizedRawAmount = typeof rawAmount === 'bigint' ? Number(rawAmount) : rawAmount
```

`Number(bigint)` silently loses precision for values > `Number.MAX_SAFE_INTEGER` (2^53). For assets with high decimal counts or large supplies, this could produce incorrect fiat valuations.

---

### 🟢 MINOR / NITS

#### 10. Test coverage gaps — `src/test/screens/wallet/swap.test.tsx`

Tests cover navigation flow but not:
- Swap execution (`confirmPrototypeSwap`)
- Balance validation (`canSpendSwapAmount`)
- Error states (confirm failure)
- Amount mode toggling with decimal edge cases

Would be good to add at least a confirm-swap-happy-path test.

#### 11. `useEffect` missing dependency — `src/screens/Wallet/Swap/Index.tsx`

The `validationState` effect at ~line 103 triggers haptic on every validation state change, but `validationState` is derived from `amount`, `fromAsset`, `toAsset`, etc. This means switching from `insufficient-balance` to `quote-unavailable` will re-trigger the haptic, even though the user didn't do anything new. Consider debouncing or only triggering on specific transitions.

---

### Summary

The UI layer is solid work — well-structured components, good a11y, nice polish. The issues are all in the state management boundary between prototype swap data and the real wallet state. Items 1-5 need fixing because they either create incorrect financial displays or risk interaction with real settlement logic. The prototype state should be clearly fenced off from production wallet operations.

Requesting changes on items 1-5. Happy to re-review once addressed.

sahilc0:
Addressed Arkana must-fix items 1-5 in `9e08426d`:

- Marked prototype swap txs with `isPrototype` and guarded settlement actions from prototype transactions.
- Replaced second-level `roundTxid` generation with a unique prototype id using `crypto.randomUUID()` where available.
- Removed the incomplete SDK `Asset` payload and the `as any` cast from prototype swap txs; swap activity now reads from explicit prototype swap metadata instead.
- Removed fabricated review rows for price impact, network cost, and arrival until those values come from real quote data.
- Removed hardcoded USD activity fallback by formatting through `convertFiat(..., config.fiat)` when a fiat fallback is needed.

Local verification after the fixes:
- `bun run lint`
- `bun run format:check`
- `bun run test:unit -- src/test/screens/wallet/swap.test.tsx`
- `bun run build`

sahilc0:
Superseded by #727, which ports this swap UI work onto current master now that #627 is merged.

sahilc0:
Closing as superseded by #727.

PR #636, opened 2026-05-28
fix bip21 parser

bordalix:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BIP21 URI parsing to correctly handle case-insensitive parameters (amount, lightning, asset identifiers).

* **Tests**
  * Added test coverage for uppercase BIP21 URI variants.

* **Chores**
  * Updated testing framework dependency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/wallet/pull/636?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
   },
   "devDependencies": {
     "@eslint/compat": "^1.3.2",
-    "@playwright/test": "^1.56.1",
+    "@playwright/test": "^1.60.0",
     "@tailwindcss/vite": "^4.2.2",
     "@testing-library/jest-dom": "^6.8.0",
     "@testing-library/react": "^16.3.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -133,8 +133,8 @@ importers:
         specifier: ^1.3.2
         version: 1.4.1(eslint@8.57.1)
       '@playwright/test':
-        specifier: ^1.56.1
-        version: 1.56.1
+        specifier: ^1.60.0
+        version: 1.60.0
       '@tailwindcss/vite':
         specifier: ^4.2.2
         version: 4.2.4(vite@7.1.12(@types/node@24.10.0)(jiti@2.6.1)(lightningcss@1.32.0))
@@ -955,8 +955,8 @@ packages:
     resolution: {integrity: sha512-QNqXyfVS2wm9hweSYD2O7F0G06uurj9kZ96TRQE5Y9hU7+tgdZwIkbAKc5Ocy1HxEY2kuDQa6cQ1WRs/O5LFKA==}
     engines: {node: ^12.20.0 || ^14.18.0 || >=16.0.0}
 
-  '@playwright/test@1.56.1':
-    resolution: {integrity: sha512-vSMYtL/zOcFpvJCW71Q/OEGQb7KYBPAdKh35WNSkaZA75JlAO8ED8UN6GUNTm3drWomcbcqRPFqQbLae8yBTdg==}
+  '@playwright/test@1.60.0':
+    resolution: {integrity: sha512-O71yZIbAh/PxDMNGns37GHBIfrVkEVyn+AXyIa5dOTfb4/xNvRWV+Vv/NMbNCtODB/pO7vLlF2OTmMVLhmr7Ag==}
     engines: {node: '>=18'}
     hasBin: true
 
@@ -3326,13 +3326,13 @@ packages:
     resolution: {integrity: sha512-wQ0b/W4Fr01qtpHlqSqspcj3EhBvimsdh0KlHhH8HRZnMsEa0ea2fTULOXOS9ccQr3om+GcGRk4e+isrZWV8qQ==}
     engines: {node: '>=16.20.0'}
 
-  playwright-core@1.56.1:
-    resolution: {integrity: sha512-hutraynyn31F+Bifme+Ps9Vq59hKuUCz7H1kDOcBs+2oGguKkWTU50bBWrtz34OUWmIwpBTWDxaRPXrIXkgvmQ==}
+  playwright-core@1.60.0:
+    resolution: {integrity: sha512-9bW6zvX/m0lEbgTKJ6YppOKx8H3VOPBMOCFh2irXFOT4BbHgrx5hPjwJYLT40Lu+4qtD36qKc/Hn56StUW57IA==}
     engines: {node: '>=18'}
     hasBin: true
 
-  playwright@1.56.1:
-    resolution: {integrity: sha512-aFi5B0WovBHTEvpM3DzXTUaeN6eN0qWnTkKx4NQaH4Wvcmc153PdaY2UBdSYKaGYw+UyWXSVyxDUg5DoPEttjw==}
+  playwright@1.60.0:
+    resolution: {integrity: sha512-hheHdokM8cdqCb0lcE3s+zT4t4W+vvjpGxsZlDnikarzx8tSzMebh3UiFtgqwFwnTnjYQcsyMF8ei2mCO/tpeA==}
     engines: {node: '>=18'}
     hasBin: true
 
@@ -4926,9 +4926,9 @@ snapshots:
 
   '@pkgr/core@0.2.9': {}
 
-  '@playwright/test@1.56.1':
+  '@playwright/test@1.60.0':
     dependencies:
-      playwright: 1.56.1
+      playwright: 1.60.0
 
   '@reach/portal@0.13.2(react-dom@18.3.1(react@18.3.1))(react@18.3.1)':
     dependencies:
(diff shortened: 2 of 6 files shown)


ghost (changes requested):
## 🚨 REQUEST CHANGES — Critical Bug: Lowercasing entire URI destroys addresses

### The Problem

`src/lib/bip21.ts:29` — Moving `.toLowerCase()` from the `startsWith` check to the entire URI **lowercases Bitcoin addresses, Ark addresses, Lightning invoices, asset IDs, and all query parameter values**.

**Before (master, correct):**
```ts
const bip21Url = uri.trim()                          // preserves case
if (!bip21Url.toLowerCase().startsWith('bitcoin:'))   // case-insensitive scheme check only
```

**After (this PR, broken):**
```ts
const bip21Url = uri.toLowerCase().trim()   // DESTROYS case of entire URI
if (!bip21Url.startsWith('bitcoin:'))        // now redundant since already lowered
```

### Why This Loses Money

1. **Base58 addresses are case-sensitive.** `1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH` lowercased to `1bggz9tcn4rm9kbzdn7kprqz87sz26samh` will fail checksum validation — or worse, silently resolve to a different address on implementations that don't validate checksums.

2. **Ark addresses** in the `?ark=` parameter get lowercased → `isArkAddress()` may reject them or they may point to the wrong destination. This is a direct fund-loss vector.

3. **Lightning invoices** in `?lightning=` get lowercased. While BOLT11 is bech32 (case-insensitive), the code at line ~70 stores the already-lowercased value from `params.get('lightning')`, which *may* be fine for bech32 but is still semantically wrong.

4. **Asset IDs** (`?assetid=`) are hex hashes — lowercase is likely fine, but it's an unnecessary mutation that could break case-sensitive lookups in `assetMetadataCache` (see `Form.tsx:217`).

5. **URLs** in query params (e.g., `r=https://bitpay.com/...`) get lowercased, breaking case-sensitive URL paths.

### The Test Fixtures Mask the Bug

`src/test/fixtures.json` — All 14 expected `address` values were changed from `1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH` to `1bggz9tcn4rm9kbzdn7kprqz87sz26samh`. The tests pass, but only because the assertions were weakened to match the broken output. The fixtures should assert that addresses preserve their original case from the URI.

### Downstream Impact

`src/screens/Wallet/Send/Form.tsx:214` — The decoded `address` flows directly into `setState({ address, ... })` which feeds the send transaction flow. A lowercased base58 address here means the wallet will attempt to send funds to an invalid/wrong address.

### Correct Fix

Only lowercase the scheme prefix for comparison, preserve everything else:

```ts
const bip21Url = uri.trim()

// case-insensitive scheme check (BIP21 says scheme is case-insensitive)
const schemeEnd = bip21Url.indexOf(':')
if (schemeEnd === -1 || bip21Url.slice(0, schemeEnd).toLowerCase() !== 'bitcoin') {
  throw new Error('Invalid BIP21 URI')
}

// preserve original case for address and params
const urlWithoutPrefix = bip21Url.slice(schemeEnd + 1)
```

If the actual bug being fixed is that `?assetId=` (camelCase) wasn't matching `params.has('assetid')`, fix that specifically by checking both cases or normalizing only the param keys.

---

⚠️ **Do not merge.** This will cause fund-loss for any user scanning a BIP21 QR code with a Base58 address.

ghost (changes requested):
## 🚨 REQUEST CHANGES — Critical Bug: Lowercasing entire URI destroys addresses

### The Problem

`src/lib/bip21.ts:29` — Moving `.toLowerCase()` from the `startsWith` check to the entire URI **lowercases Bitcoin addresses, Ark addresses, Lightning invoices, asset IDs, and all query parameter values**.

**Before (master, correct):**
```ts
const bip21Url = uri.trim()                          // preserves case
if (!bip21Url.toLowerCase().startsWith('bitcoin:'))   // case-insensitive scheme check only
```

**After (this PR, broken):**
```ts
const bip21Url = uri.toLowerCase().trim()   // DESTROYS case of entire URI
if (!bip21Url.startsWith('bitcoin:'))        // now redundant since already lowered
```

### Why This Loses Money

1. **Base58 addresses are case-sensitive.** `1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH` lowercased to `1bggz9tcn4rm9kbzdn7kprqz87sz26samh` will fail checksum validation — or worse, silently resolve to a different address on implementations that don't validate checksums.

2. **Ark addresses** in the `?ark=` parameter get lowercased → `isArkAddress()` may reject them or they may point to the wrong destination. This is a direct fund-loss vector.

3. **Lightning invoices** in `?lightning=` get lowercased. While BOLT11 is bech32 (case-insensitive), the code at line ~70 stores the already-lowercased value from `params.get('lightning')`, which *may* be fine for bech32 but is still semantically wrong.

4. **Asset IDs** (`?assetid=`) are hex hashes — lowercase is likely fine, but it's an unnecessary mutation that could break case-sensitive lookups in `assetMetadataCache` (see `Form.tsx:217`).

5. **URLs** in query params (e.g., `r=https://bitpay.com/...`) get lowercased, breaking case-sensitive URL paths.

### The Test Fixtures Mask the Bug

`src/test/fixtures.json` — All 14 expected `address` values were changed from `1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH` to `1bggz9tcn4rm9kbzdn7kprqz87sz26samh`. The tests pass, but only because the assertions were weakened to match the broken output. The fixtures should assert that addresses preserve their original case from the URI.

### Downstream Impact

`src/screens/Wallet/Send/Form.tsx:214` — The decoded `address` flows directly into `setState({ address, ... })` which feeds the send transaction flow. A lowercased base58 address here means the wallet will attempt to send funds to an invalid/wrong address.

### Correct Fix

Only lowercase the scheme prefix for comparison, preserve everything else:

```ts
const bip21Url = uri.trim()

// case-insensitive scheme check (BIP21 says scheme is case-insensitive)
const schemeEnd = bip21Url.indexOf(':')
if (schemeEnd === -1 || bip21Url.slice(0, schemeEnd).toLowerCase() !== 'bitcoin') {
  throw new Error('Invalid BIP21 URI')
}

// preserve original case for address and params
const urlWithoutPrefix = bip21Url.slice(schemeEnd + 1)
```

If the actual bug being fixed is that `?assetId=` (camelCase) wasn't matching `params.has('assetid')`, fix that specifically by checking both cases or normalizing only the param keys.

---

⚠️ **Do not merge.** This will cause fund-loss for any user scanning a BIP21 QR code with a Base58 address.

ghost (approved):
## ✅ APPROVE — Critical bug fixed, code looks correct

Previous review flagged `uri.toLowerCase().trim()` destroying case-sensitive addresses. The new commits properly fix this:

### What changed (since my last review on `8be5a16`)

1. **`src/lib/bip21.ts`** — `.toLowerCase()` removed from URI normalization; now only used for scheme prefix comparison (`bip21Url.toLowerCase().startsWith('bitcoin:')`). Case-preserving. ✅
2. **`src/test/fixtures.json`** — Address fixtures reverted to original case (`1BgGZ9tcN4rm9KBzDn7KprQz87SZ26SAMH`). ✅
3. **Uppercase param keys** — `params.has('ark') || params.has('ARK')` pattern handles QR-code alphanumeric mode (all-caps). Reasonable approach. ✅
4. **`if (address)` guard** — Handles `BITCOIN:?ARK=...` (no address, query-only) correctly. ✅
5. **New test** — Uppercase URI with ARK + LNURL. Good coverage. ✅
6. **Playwright bump** — `1.56.1` → `1.60.0`. Unrelated, fine.

### Minor notes (non-blocking)

- `bip21.ts:52-53` — The `|| params.has('ARK')` pattern only covers exact lowercase and exact uppercase. Mixed case like `Ark` or `lightning` with random caps would be missed. For the real use case (QR alphanumeric mode = all-caps), this is fine. A more robust approach would be case-normalizing param keys during iteration, but it's not worth blocking on.
- `address.ts:50` — `isBTCAddress()` regex only matches lowercase bech32 prefixes (`bc1`, `tb1`). An all-caps `BITCOIN:BC1Q...` URI from a QR code would parse correctly here (address preserved as `BC1Q...`) but would fail `isBTCAddress()` validation downstream. Not introduced by this PR, but worth noting for a follow-up.

LGTM. Ship it.

ghost (approved):
## ✅ APPROVE — New commits look good

**Reviewing only the 2 new commits since my previous approval (`a666348`):**

### 1. `make isBTCAddress case insensitive` — `src/lib/address.ts:49-50`

Adds `'i'` flag to both segwit and legacy regexes. This directly addresses my previous note that uppercase bech32 addresses from QR codes (`BC1Q...`) would fail `isBTCAddress()` detection. Correct fix — `Form.tsx:263` calls `isBTCAddress(recipient)` without lowercasing, so the flag is needed.

**Minor note (non-blocking):** The `'i'` flag technically accepts mixed-case bech32 (`Bc1qAb...`) which is invalid per BIP173. Acceptable here — this is a detection function, not a validator. Downstream signing libraries will reject mixed-case.

### 2. `update @playwright/test to 1.60.0` — `package.json`

Dev dependency bump. No concern.

No protocol-critical changes in these commits. Maintaining approval.

PR #639, opened 2026-05-31
fix bip21 parser

bordalix:
## Release Notes

* **Bug Fixes**
  * Improved error detection and reporting for Lightning Network URL (LNURL) responses
  * Enhanced validation in the send form to properly enforce LNURL amount constraints and eligibility checks
  * Fixed LNURL response handling to correctly determine transfer methods

* **Tests**
  * Updated LNURL-related test assertions to verify proper constraint enforcement

### DIFF
--- a/src/lib/bip21.ts
+++ b/src/lib/bip21.ts
@@ -12,7 +12,7 @@ export interface Bip21Decoded {
   /** Raw amount in asset units when assetId is present (not converted to satoshis) */
   assetAmount?: string
   invoice?: string
-  lnurl?: string
+  lnUrl?: string
   assetId?: string
 }
 
@@ -22,7 +22,7 @@ export const decodeBip21 = (uri: string): Bip21Decoded => {
     address: undefined,
     satoshis: undefined,
     invoice: undefined,
-    lnurl: undefined,
+    lnUrl: undefined,
     assetId: undefined,
     assetAmount: undefined,
     arkAddress: undefined,
@@ -70,7 +70,7 @@ export const decodeBip21 = (uri: string): Bip21Decoded => {
     if (params.has('lightning') || params.has('LIGHTNING')) {
       const lightning = params.get('lightning') ?? params.get('LIGHTNING')!
       if (lightning.toLowerCase().startsWith('lnurl')) {
-        result.lnurl = lightning
+        result.lnUrl = lightning
       } else if (lightning.toLowerCase().startsWith('ln')) {
         result.invoice = lightning
       }
--- a/src/lib/lnurl.ts
+++ b/src/lib/lnurl.ts
@@ -5,7 +5,7 @@ import { hmac } from '@noble/hashes/hmac.js'
 const emailRegex =
   /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
 
-type LnUrlResponse = {
+export type LnUrlResponse = {
   commentAllowed?: number
   callback: string
   minSendable: number
@@ -28,9 +28,11 @@ type LnUrlCallbackResponse = {
   pr: string
 }
 
-const checkResponse = <T = any>(response: Response): Promise<T> => {
+const checkResponse = async <T = any>(response: Response): Promise<T> => {
   if (!response.ok) return Promise.reject(response)
-  return response.json()
+  const data = await response.json()
+  if (data.status === 'ERROR') return Promise.reject(data.reason || 'LNURL error')
+  return data
 }
 
 const checkLnUrlResponse = (amount: number, data: LnUrlResponse) => {
--- a/src/screens/Wallet/Send/Form.tsx
+++ b/src/screens/Wallet/Send/Form.tsx
@@ -212,7 +212,6 @@ export default function SendForm() {
       }
       if (isBip21(lowerCaseData)) {
         const { address, arkAddress, invoice, lnUrl, satoshis, assetId, assetAmount } = decodeBip21(recipient.trim())
-        console.log({ address, arkAddress, invoice, lnUrl, satoshis, assetId, assetAmount })
         if (!address && !arkAddress && !invoice && !lnUrl) return setError('Unable to parse bip21')
         if (assetId) {
           let found = assetOptions.find((a) => a.assetId === assetId)
@@ -361,7 +360,6 @@ export default function SendForm() {
 
   // check lnurl conditions
   useEffect(() => {
-    console.log('Checking LNURL conditions for', sendInfo.lnUrl)
     if (!sendInfo.lnUrl) return
     if (sendInfo.lnUrl && sendInfo.invoice) return
     checkLnUrlConditions(sendInfo.lnUrl)
@@ -393,7 +391,7 @@ export default function SendForm() {
       return setError('Sending onchain not allowed')
     }
     // check server limits for offchain transactions
-    if (!address && !vtxoTxsAllowed()) {
+    if (!address && (arkAddress || invoice || lnUrl) && !vtxoTxsAllowed()) {
       return setError('Sending offchain not allowed')
     }
     // check swap limits for lightning transactions
@@ -584,9 +582,9 @@ export default function SendForm() {
     setProcessing(true)
     const satoshis = sendInfo.satoshis ?? 0
     try {
-      if (sendInfo.lnUrl) {
+      if (sendInfo.lnUrl && lnUrlResponse) {
         // Check if Ark method is available
-        const arkMethod = lnUrlResponse!.transferAmounts?.find((method) => method.method === 'Ark' && method.available)
+        const arkMethod = lnUrlResponse.transferAmounts?.find((method) => method.method === 'Ark' && method.available)
 
         if (arkMethod) {
           // Fetch Ark address instead of Lightning invoice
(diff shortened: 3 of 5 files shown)


ghost (changes requested):
## Code Review — `fix bip21 parser`

### 🚨 CRITICAL: Unit mismatch — millisatoshis vs satoshis

The old code stored converted values in `lnUrlLimits: { min, max }` (satoshis, divided by 1000). The new code stores the raw `LnUrlResponse` where `minSendable`/`maxSendable` are in **millisatoshis**. Multiple consumption sites now compare millisatoshis against satoshis:

**`Form.tsx` — limits check useEffect (~line 344-358 in new code):**
```tsx
const { minSendable: min, maxSendable: max } = lnUrlResponse
if (min > balance) return setError('Insufficient funds for LNURL')  // millisats vs sats
if (satoshis && satoshis < min) return setError(...)                 // sats vs millisats
if (satoshis && satoshis > max) return setError(...)                 // sats vs millisats
```
All three comparisons are 1000x off. A 100 sat minimum would appear as 100,000, always showing "Insufficient funds" or allowing amounts far below the real min.

**`Form.tsx` — button disabled logic (~line 680-683):**
```tsx
(lnUrlResponse?.maxSendable && satoshis > lnUrlResponse.maxSendable) ||
(lnUrlResponse?.minSendable && satoshis < lnUrlResponse.minSendable) ||
```
Same unit mismatch. The send button would be enabled/disabled at wrong thresholds.

**`Form.tsx` — label logic (~line 439-443):**
```tsx
: lnUrlResponse?.minSendable && satoshis < lnUrlResponse.minSendable
: lnUrlResponse?.maxSendable && satoshis > lnUrlResponse.maxSendable
```
Same issue — wrong error labels shown to users.

**`Form.tsx` — Amount input props (~line 938-939):**
```tsx
min={lnUrlResponse?.minSendable}
max={lnUrlResponse?.maxSendable}
```
Passing millisatoshis where the component expects satoshis.

**Fix:** Either convert when consuming (`Math.floor(lnUrlResponse.minSendable / 1000)`) at every site, or store a derived `{ min, max }` in satoshis alongside the response.

---

### 🐛 BUG: Offchain guard condition widened incorrectly

`Form.tsx` (~line 397 in new code):
```tsx
// OLD: if (!address && (arkAddress || invoice) && !vtxoTxsAllowed())
// NEW: if (!address && !vtxoTxsAllowed())
```

The `(arkAddress || invoice)` guard ensured this error only triggers when an offchain recipient is actually set. Without it, whenever `address` is undefined (including when the form is empty or an lnUrl-only recipient is entered) and vtxo txs are restricted, the error fires prematurely. The fix for lnUrl should be:
```tsx
if (!address && (arkAddress || invoice || lnUrl) && !vtxoTxsAllowed())
```

---

### 🐛 BUG: Incomplete rename — unit test not updated

`src/test/lib/bip21.test.ts:10` still references `lnurl: undefined` after the interface renames the field to `lnUrl`. This will fail TypeScript compilation:
```tsx
expect(decodeBip21(bip21)).toEqual({ ..., lnurl: undefined })
//                                        ^^^^^ should be lnUrl
```

---

### ⚠️ Minor: Debug console.log left in

`Form.tsx` — two debug statements should be removed before merge:
```tsx
console.log({ address, arkAddress, invoice, lnUrl, satoshis, assetId, assetAmount })
console.log('Checking LNURL conditions for', sendInfo.lnUrl)
```

---

### ⚠️ Minor: Non-null assertion risk on `lnUrlResponse!`

`Form.tsx` (~line 588 in new code):
```tsx
const arkMethod = lnUrlResponse!.transferAmounts?.find(...)
```

If the user triggers `handleProceed` before the LNURL conditions fetch completes (race condition), this crashes. Use optional chaining: `lnUrlResponse?.transferAmounts?.find(...)` with a guard or early return.

---

### ✅ Good changes

- `checkResponse` in `lnurl.ts`: Checking `data.status === 'ERROR'` is a solid fix — LNURL servers return 200 with error payloads, and this was silently swallowed before.
- Exporting `LnUrlResponse` type and caching the response to avoid a redundant fetch in `handleProceed` is a good refactor (once the unit mismatch is fixed).
- Better error handling with `extractError(e)` instead of the generic `'Invalid address or LNURL'` string.
- Test updates for the `lnUrl` rename in e2e tests are correct.

---

**Verdict:** The millisatoshi/satoshi unit mismatch is a **money-impacting bug** — it will cause wrong limit enforcement on every LNURL payment. Please fix the unit conversions, the offchain guard, and the missed test file before re-review.

ghost (changes requested):
## Code Review — `fix bip21 parser`

### 🚨 CRITICAL: Unit mismatch — millisatoshis vs satoshis

The old code stored converted values in `lnUrlLimits: { min, max }` (satoshis, divided by 1000). The new code stores the raw `LnUrlResponse` where `minSendable`/`maxSendable` are in **millisatoshis**. Multiple consumption sites now compare millisatoshis against satoshis:

**`Form.tsx` — limits check useEffect (~line 344-358 in new code):**
```tsx
const { minSendable: min, maxSendable: max } = lnUrlResponse
if (min > balance) return setError('Insufficient funds for LNURL')  // millisats vs sats
if (satoshis && satoshis < min) return setError(...)                 // sats vs millisats
if (satoshis && satoshis > max) return setError(...)                 // sats vs millisats
```
All three comparisons are 1000x off. A 100 sat minimum would appear as 100,000, always showing "Insufficient funds" or allowing amounts far below the real min.

**`Form.tsx` — button disabled logic (~line 680-683):**
```tsx
(lnUrlResponse?.maxSendable && satoshis > lnUrlResponse.maxSendable) ||
(lnUrlResponse?.minSendable && satoshis < lnUrlResponse.minSendable) ||
```
Same unit mismatch. The send button would be enabled/disabled at wrong thresholds.

**`Form.tsx` — label logic (~line 439-443):**
```tsx
: lnUrlResponse?.minSendable && satoshis < lnUrlResponse.minSendable
: lnUrlResponse?.maxSendable && satoshis > lnUrlResponse.maxSendable
```
Same issue — wrong error labels shown to users.

**`Form.tsx` — Amount input props (~line 938-939):**
```tsx
min={lnUrlResponse?.minSendable}
max={lnUrlResponse?.maxSendable}
```
Passing millisatoshis where the component expects satoshis.

**Fix:** Either convert when consuming (`Math.floor(lnUrlResponse.minSendable / 1000)`) at every site, or store a derived `{ min, max }` in satoshis alongside the response.

---

### 🐛 BUG: Offchain guard condition widened incorrectly

`Form.tsx` (~line 397 in new code):
```tsx
// OLD: if (!address && (arkAddress || invoice) && !vtxoTxsAllowed())
// NEW: if (!address && !vtxoTxsAllowed())
```

The `(arkAddress || invoice)` guard ensured this error only triggers when an offchain recipient is actually set. Without it, whenever `address` is undefined (including when the form is empty or an lnUrl-only recipient is entered) and vtxo txs are restricted, the error fires prematurely. The fix for lnUrl should be:
```tsx
if (!address && (arkAddress || invoice || lnUrl) && !vtxoTxsAllowed())
```

---

### 🐛 BUG: Incomplete rename — unit test not updated

`src/test/lib/bip21.test.ts:10` still references `lnurl: undefined` after the interface renames the field to `lnUrl`. This will fail TypeScript compilation:
```tsx
expect(decodeBip21(bip21)).toEqual({ ..., lnurl: undefined })
//                                        ^^^^^ should be lnUrl
```

---

### ⚠️ Minor: Debug console.log left in

`Form.tsx` — two debug statements should be removed before merge:
```tsx
console.log({ address, arkAddress, invoice, lnUrl, satoshis, assetId, assetAmount })
console.log('Checking LNURL conditions for', sendInfo.lnUrl)
```

---

### ⚠️ Minor: Non-null assertion risk on `lnUrlResponse!`

`Form.tsx` (~line 588 in new code):
```tsx
const arkMethod = lnUrlResponse!.transferAmounts?.find(...)
```

If the user triggers `handleProceed` before the LNURL conditions fetch completes (race condition), this crashes. Use optional chaining: `lnUrlResponse?.transferAmounts?.find(...)` with a guard or early return.

---

### ✅ Good changes

- `checkResponse` in `lnurl.ts`: Checking `data.status === 'ERROR'` is a solid fix — LNURL servers return 200 with error payloads, and this was silently swallowed before.
- Exporting `LnUrlResponse` type and caching the response to avoid a redundant fetch in `handleProceed` is a good refactor (once the unit mismatch is fixed).
- Better error handling with `extractError(e)` instead of the generic `'Invalid address or LNURL'` string.
- Test updates for the `lnUrl` rename in e2e tests are correct.

---

**Verdict:** The millisatoshi/satoshi unit mismatch is a **money-impacting bug** — it will cause wrong limit enforcement on every LNURL payment. Please fix the unit conversions, the offchain guard, and the missed test file before re-review.

ghost (approved):
Clean fix. A few observations:

**What's good**

- `lnurl` → `lnUrl` rename is the right camelCase consistency fix, and it's applied everywhere (type, init, decode, state, tests).
- `checkResponse` now properly rejects on `{ status: 'ERROR' }` payloads — this was a real gap since LNURL servers return HTTP 200 with an error body.
- Replacing `lnUrlLimits` with the full `LnUrlResponse` and reusing it in `handleSend` is a nice improvement — avoids a redundant network round-trip on submit.
- Better error propagation: `extractError(e)` instead of a hardcoded `'Invalid address or LNURL'` string.
- Adding `sendInfo.lnUrl` to the server-limits effect dependency array was a missing dep — good catch.

**One thing to verify**

The stored `lnUrlResponse` has `minSendable`/`maxSendable` already converted to **satoshis** (divided by 1000 at line ~368), but `LnUrlResponse` is typed as raw millisats coming from the server. The in-state values are semantically different from what the type implies. This works correctly today because `fetchInvoice` fetches fresh from the server and does its own conversion — but it could confuse future readers.

Consider either:
1. A local `LnUrlLimits` type with a comment noting the unit, or
2. Storing the raw response and dividing on use.

Not a blocker — the runtime behavior is correct — but worth a follow-up.

Approved.

PR #642, opened 2026-06-01
double the distance needed to trigger pull-to-refresh

bordalix:
## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted pull-to-refresh gesture sensitivity to require a larger vertical swipe distance, reducing unintended activations.

### DIFF
--- a/src/components/Refresher.tsx
+++ b/src/components/Refresher.tsx
@@ -19,7 +19,7 @@ export default function Refresher() {
   const handleTouchMove = (e: TouchEvent) => {
     const touchY = e.touches[0].clientY
     const touchDiff = touchY - touchstartY
-    if (touchDiff > 50 && window.scrollY === 0) {
+    if (touchDiff > 100 && window.scrollY === 0) {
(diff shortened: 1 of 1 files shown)


ghost (approved):
**LGTM** — clean, minimal UX fix. Approving.

### What this does
Single-line change: pull-to-refresh trigger threshold `50px → 100px`. Reduces accidental refreshes on mobile. No protocol, crypto, or API surface touched.

### Review notes
- ✅ No cross-repo impact (component is self-contained, no exports consumed elsewhere)
- ✅ No protocol-critical code paths affected
- ✅ No security implications

### Pre-existing nit (not blocking)
`Refresher.tsx` has `let touchstartY` and `let triggered` as plain variables inside the component body — they reset on every re-render. If the component ever re-renders mid-gesture (e.g. from a context update), the gesture tracking breaks. These should be `useRef` values. Not introduced by this PR, but worth a follow-up.

🤖 Reviewed by Arkana

PR #643, opened 2026-06-01
don't check lnUrl conditions if ark address is present

bordalix:
## Summary by CodeRabbit

* **Bug Fixes**
  * Send flow no longer runs unnecessary LNURL checks when using Ark addresses, avoiding erroneous error states.
* **Improvements**
  * BIP21 URI generation now normalizes separators and omits empty query segments for cleaner, more consistent payment URIs.
* **Tests**
  * Expanded end-to-end and unit test coverage and added shared helpers for generating invoices, BIP21 addresses, and reset flows to improve reliability.

### DIFF
--- a/src/components/Error.tsx
+++ b/src/components/Error.tsx
@@ -12,7 +12,7 @@ export default function ErrorMessage({ error, text }: ErrorProps) {
   return (
     <Shadow red>
       <Padded>
-        <Text bold centered color='white' small wrap>
+        <Text bold centered color='white' small wrap testId='error-message'>
           {text}
         </Text>
       </Padded>
--- a/src/lib/bip21.ts
+++ b/src/lib/bip21.ts
@@ -81,12 +81,12 @@ export const decodeBip21 = (uri: string): Bip21Decoded => {
 }
 
 export const encodeBip21 = (address: string, arkAddress: string, invoice: string, sats: number, lnurl?: string) => {
-  return (
-    `bitcoin:${address}` +
-    `?ark=${arkAddress}` +
-    (invoice ? `&lightning=${invoice}` : lnurl ? `&lightning=${lnurl}` : '') +
-    (sats ? `&amount=${prettyNumber(fromSatoshis(sats))}` : '')
-  )
+  const bip21 =
+    `bitcoin:${address}?` +
+    (arkAddress ? `ark=${arkAddress}&` : '') +
+    (invoice ? `lightning=${invoice}&` : lnurl ? `lightning=${lnurl}&` : '') +
+    (sats ? `amount=${prettyNumber(fromSatoshis(sats))}` : '')
+  return bip21.endsWith('&') || bip21.endsWith('?') ? bip21.slice(0, -1) : bip21
 }
 
 export const encodeBip21Asset = (arkAddress: string, assetId: string, cents: bigint, decimals?: number) => {
--- a/src/screens/Wallet/Send/Form.tsx
+++ b/src/screens/Wallet/Send/Form.tsx
@@ -376,7 +376,7 @@ export default function SendForm() {
         consoleError(e, 'Error checking LNURL conditions')
         setError(extractError(e))
       })
-  }, [sendInfo.lnUrl])
+  }, [sendInfo.arkAddress, sendInfo.lnUrl])
 
   // check if user wants to send all funds
   useEffect(() => {
--- a/src/screens/Wallet/Send/Form.tsx
+++ b/src/screens/Wallet/Send/Form.tsx
@@ -362,6 +362,7 @@ export default function SendForm() {
   // check lnurl conditions
   useEffect(() => {
     if (!sendInfo.lnUrl) return
+    if (sendInfo.arkAddress) return
     if (sendInfo.lnUrl && sendInfo.invoice) return
     checkLnUrlConditions(sendInfo.lnUrl)
       .then((conditions) => {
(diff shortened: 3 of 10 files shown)


ghost (approved):
## ✅ Approved

Clean, minimal fix. Correctly short-circuits the LNURL conditions check when an ark address is already available from BIP21 parsing.

**Verified:**
- `handleContinue` path: `lnUrlResponse` stays `undefined` → skips LNURL flow → `proceed` effect sees `arkAddress` → navigates to `SendDetails`. ✓
- `buttonDisabled`: `arkAddress` satisfies the recipient check, no stale LNURL limits applied. ✓
- Recipient validation effect: server key check, vtxo limits, self-send all handled correctly with `arkAddress` set. ✓
- No impact on non-BIP21 flows (plain LNURL without ark address still resolves normally). ✓

**One minor nit** (non-blocking):

`Form.tsx:365` — The `useEffect` dep array is `[sendInfo.lnUrl]` but the effect body now reads `sendInfo.arkAddress`. Both fields are set atomically via `setState` from `decodeBip21`, so React batching makes this safe in practice. But if you ever add a flow where `arkAddress` is set *after* `lnUrl` (e.g., async resolution), this effect wouldn't re-evaluate. Consider adding `sendInfo.arkAddress` to the dep array for correctness:

```js
}, [sendInfo.lnUrl, sendInfo.arkAddress])
```

Not protocol-critical — this is UI payment-method selection logic only.

🤖 *Arkana code review — wallet#643*

ghost (approved):
## ✅ Approved

Clean, minimal fix. Correctly short-circuits the LNURL conditions check when an ark address is already available from BIP21 parsing.

**Verified:**
- `handleContinue` path: `lnUrlResponse` stays `undefined` → skips LNURL flow → `proceed` effect sees `arkAddress` → navigates to `SendDetails`. ✓
- `buttonDisabled`: `arkAddress` satisfies the recipient check, no stale LNURL limits applied. ✓
- Recipient validation effect: server key check, vtxo limits, self-send all handled correctly with `arkAddress` set. ✓
- No impact on non-BIP21 flows (plain LNURL without ark address still resolves normally). ✓

**One minor nit** (non-blocking):

`Form.tsx:365` — The `useEffect` dep array is `[sendInfo.lnUrl]` but the effect body now reads `sendInfo.arkAddress`. Both fields are set atomically via `setState` from `decodeBip21`, so React batching makes this safe in practice. But if you ever add a flow where `arkAddress` is set *after* `lnUrl` (e.g., async resolution), this effect wouldn't re-evaluate. Consider adding `sendInfo.arkAddress` to the dep array for correctness:

```js
}, [sendInfo.lnUrl, sendInfo.arkAddress])
```

Not protocol-critical — this is UI payment-method selection logic only.

🤖 *Arkana code review — wallet#643*

ghost (changes requested):
## 🔍 Iterative review — new "add tests" commit

Previous approval stands for the core fix (`Form.tsx` ark address short-circuit). This review covers the new commit only.

### 🐛 Bug — `swap.test.ts:184-198` "should refund failing swap"

The refactoring introduced a **real bug** in this test. The diff replaced:

```diff
-  const invoice = outputJSON.payment_request
+  const invoice = await getInvoiceFromLND(1000)
```

But lines 184-189 **still create an invoice via `execAsync`** and parse `outputJSON`. Lines 196-198 then cancel that **first** invoice using `outputJSON.r_hash`. However, `invoice` is now a **second, separate invoice** from `getInvoiceFromLND(1000)` — which was **never cancelled**.

The test then fills this uncancelled invoice at line 202 and expects `"Swap failed"` at line 206 — but the swap will succeed because the invoice is valid. **This breaks the test's purpose.**

**Fix:** Don't use `getInvoiceFromLND` here. This test uniquely needs access to both `payment_request` AND `r_hash` from the same invoice to cancel it. Keep the original inline `execAsync` approach for this specific test, or extend `getInvoiceFromLND` to return the full response object.

### ✅ Everything else looks good

- **`encodeBip21` change** (`bip21.ts:84-90`): Correctly handles optional `arkAddress`. Trailing `&`/`?` stripping is sound. All parameter combinations produce valid URIs.
- **`form.test.ts`**: Good coverage — tests ark prioritization over invoice+lnurl, lightning fallback without ark, and lnurl fallback. Hardcoded test constants are fine for these scenarios.
- **`getInvoiceFromLND` / `createWalletAndGetBIP21` / `resetWallet`** extraction to utils: Clean DRY refactor.
- **`bip21.test.ts`**: New unit tests for decode/encode without ark address are correct.
- **`Error.tsx`**: `testId` addition is trivial and correct.

### Previous nit (still applicable)

`Form.tsx:365` — `useEffect` dep array is `[sendInfo.lnUrl]` but body reads `sendInfo.arkAddress`. Consider adding `sendInfo.arkAddress` to deps. Non-blocking.

🤖 *Arkana code review — wallet#643*

ghost (approved):
## ✅ Iterative review — fixes verified

Two new commits (`ab1d0a5d`, `5bfe51f9`) address both issues from my previous review:

1. **`swap.test.ts:191`** — Reverted to `outputJSON.payment_request` so the cancelled and paid invoices are the same object. Bug fixed. ✓
2. **`Form.tsx:379`** — Added `sendInfo.arkAddress` to the `useEffect` dep array. Nit resolved. ✓

No new concerns. LGTM.

🤖 *Arkana code review — wallet#643*

PR #645, opened 2026-06-03
contracts view: copy encoded params, empty state, active/inactive sections

pietro909:
## Summary

- Add `encodeArkContract` call to render the URI-like `arkcontract=…` string and let users copy it
- Replace inline `cardStyle` div with `Shadow border` (consistent with the rest of the codebase)
- Show `contract.label` (when set) and `createdAt` as relative time on each card
- Split list into **Active** / **Inactive** labeled sections
- Add empty state when no contracts are found

## Screenshot

Screenshot from 2026-06-03 10-12-00

## Summary by CodeRabbit

* **Improvements**
  * Contracts are now organized into "Active" and "Inactive" sections for easier viewing
  * Contract parameters are now displayed for each contract entry
  * Enhanced contract card layout with improved visual formatting

### DIFF
--- a/src/screens/Settings/Contracts.tsx
+++ b/src/screens/Settings/Contracts.tsx
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useRef, useState } from 'react'
+import { useContext, useEffect, useMemo, useRef, useState } from 'react'
 import { Contract, encodeArkContract } from '@arkade-os/sdk'
 import Header from './Header'
 import Content from '../../components/Content'
@@ -45,6 +45,14 @@ function CopyRow({ label, value }: { label: string; value: string }) {
 }
 
 function ContractCard({ contract }: { contract: Contract }) {
+  const encoded = useMemo(() => {
+    try {
+      return encodeArkContract(contract)
+    } catch {
+      return ''
+    }
+  }, [contract])
+
   return (
     <Shadow border>
       <FlexCol gap='0.5rem'>
@@ -60,14 +68,14 @@ function ContractCard({ contract }: { contract: Contract }) {
               {contract.state}
             </Text>
             <Text tiny color='neutral-500'>
-              {prettyAgo(contract.createdAt)}
+              {contract.createdAt ? prettyAgo(contract.createdAt) : 'Unknown'}
             </Text>
           </FlexCol>
         </FlexRow>
         <hr className='dashed' />
         <CopyRow label='address' value={contract.address} />
         <CopyRow label='script' value={contract.script} />
-        <CopyRow label='parameters' value={encodeArkContract(contract)} />
+        {encoded ? <CopyRow label='parameters' value={encoded} /> : null}
       </FlexCol>
     </Shadow>
   )
@@ -81,7 +89,7 @@ function Section({ title, contracts }: { title: string; contracts: Contract[] })
         {title}
       </Text>
       {contracts.map((c) => (
-        <ContractCard key={c.script} contract={c} />
+        <ContractCard key={c.address} contract={c} />
       ))}
     </FlexCol>
   )
--- a/src/screens/Settings/Contracts.tsx
+++ b/src/screens/Settings/Contracts.tsx
@@ -1,5 +1,5 @@
 import { useContext, useEffect, useRef, useState } from 'react'
-import type { Contract } from '@arkade-os/sdk'
+import { Contract, ContractManager, encodeArkContract } from '@arkade-os/sdk'
 import Header from './Header'
 import Content from '../../components/Content'
 import Padded from '../../components/Padded'
@@ -65,6 +65,7 @@ function ContractCard({ contract }: { contract: Contract }) {
         <hr className='dashed' />
         <CopyRow label='address' value={contract.address} />
         <CopyRow label='script' value={contract.script} />
+        <CopyRow label='parameters' value={encodeArkContract(contract)} />
       </FlexCol>
     </div>
   )
--- a/src/screens/Settings/Contracts.tsx
+++ b/src/screens/Settings/Contracts.tsx
@@ -11,19 +11,11 @@ import LoadingLogo from '../../components/LoadingLogo'
 import CopyIcon from '../../icons/Copy'
 import CheckMarkIcon from '../../icons/CheckMark'
 import { WalletContext } from '../../providers/wallet'
-import { prettyLongText } from '../../lib/format'
+import { prettyAgo, prettyLongText } from '../../lib/format'
 import { copyToClipboard } from '../../lib/clipboard'
 import { hapticSubtle } from '../../lib/haptics'
 import { useToast } from '../../components/Toast'
 import { consoleError } from '../../lib/logs'
-
-const cardStyle: React.CSSProperties = {
-  backgroundColor: 'var(--neutral-100)',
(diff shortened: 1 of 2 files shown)


ghost (commented):
## Code Review — `contracts view: copy encoded params, empty state, active/inactive sections`

**Scope**: Single-file UI change (`src/screens/Settings/Contracts.tsx`). Not protocol-critical — no VTXO, signing, forfeit, or exit path logic touched.

**Verdict**: Clean, low-risk UI improvement. Two actionable items before merge.

---

### 🔴 Must fix

**1. `encodeArkContract` called unguarded in render — can crash the screen**
`src/screens/Settings/Contracts.tsx:68`

```tsx
<CopyRow label='parameters' value={encodeArkContract(contract)} />
```

If `encodeArkContract` throws on a contract with unexpected/malformed data (e.g. a contract created by an older SDK version missing a field), the entire Contracts screen white-screens with an unhandled React error. This is the only screen where users can inspect their contracts — losing it is bad UX.

Wrap it:
```tsx
const encoded = (() => {
  try { return encodeArkContract(contract) } catch { return '' }
})()
// ... then conditionally render:
{encoded && <CopyRow label='parameters' value={encoded} />}
```
Or compute it in a `useMemo` with a try-catch inside `ContractCard`.

**2. Tests are stale — will break on this PR**
`src/test/screens/settings/contracts.test.tsx`

The mock contracts lack `createdAt` (now rendered via `prettyAgo`) and `label`. More critically, `encodeArkContract` is imported as a real SDK function but will be called on incomplete mock data — this will either throw or produce garbage.

Needs:
- Add `createdAt: 1717000000` (and optionally `label`) to mock contracts
- Mock or spy `encodeArkContract` from `@arkade-os/sdk` to avoid coupling test to SDK internals
- Add a test case for the empty state (`contracts.length === 0` → renders "No contracts found.")
- Existing assertions for `screen.getByText('default')` / `screen.getByText('vhtlc')` should still pass since the text nodes remain, but verify after adding the missing fields

---

### 🟡 Nits (non-blocking)

**3. `encodeArkContract` recomputed on every render**
`src/screens/Settings/Contracts.tsx:68`

`encodeArkContract` likely involves string encoding work. On a settings screen with a handful of contracts this is fine, but wrapping it in `useMemo` would be the defensive play — especially if this component ever gets reused in a more render-heavy context.

**4. `key={c.script}` as React key — pre-existing but worth noting**
`src/screens/Settings/Contracts.tsx:79`

If two contracts can share the same `script` (same contract type, different addresses), React will silently drop one. `c.address` would be a safer unique key, or a composite `${c.address}:${c.script}`.

---

### ✅ What's good

- `import type { Contract }` → `import { Contract, encodeArkContract }` — correct, need the runtime value now
- `prettyAgo(contract.createdAt)` — `prettyAgo` gracefully handles falsy (`if (!timestamp) return ''`) and auto-detects ms vs s timestamps. Solid.
- `contract.label ? ... : null` — correctly handles the optional `label` field
- Active/inactive split logic is clean: `=== 'active'` / `!== 'active'` is exhaustive
- `Section` returning `null` when empty prevents rendering empty headers
- Replacing inline `cardStyle` with `<Shadow border>` is a good consistency win

LGTM with the error handling fix and test updates. 🫡

—Arkana

ghost (commented):
## Code Review — `contracts view: copy encoded params, empty state, active/inactive sections`

**Scope**: Single-file UI change (`src/screens/Settings/Contracts.tsx`). Not protocol-critical — no VTXO, signing, forfeit, or exit path logic touched.

**Verdict**: Clean, low-risk UI improvement. Two actionable items before merge.

---

### 🔴 Must fix

**1. `encodeArkContract` called unguarded in render — can crash the screen**
`src/screens/Settings/Contracts.tsx:68`

```tsx
<CopyRow label='parameters' value={encodeArkContract(contract)} />
```

If `encodeArkContract` throws on a contract with unexpected/malformed data (e.g. a contract created by an older SDK version missing a field), the entire Contracts screen white-screens with an unhandled React error. This is the only screen where users can inspect their contracts — losing it is bad UX.

Wrap it:
```tsx
const encoded = (() => {
  try { return encodeArkContract(contract) } catch { return '' }
})()
// ... then conditionally render:
{encoded && <CopyRow label='parameters' value={encoded} />}
```
Or compute it in a `useMemo` with a try-catch inside `ContractCard`.

**2. Tests are stale — will break on this PR**
`src/test/screens/settings/contracts.test.tsx`

The mock contracts lack `createdAt` (now rendered via `prettyAgo`) and `label`. More critically, `encodeArkContract` is imported as a real SDK function but will be called on incomplete mock data — this will either throw or produce garbage.

Needs:
- Add `createdAt: 1717000000` (and optionally `label`) to mock contracts
- Mock or spy `encodeArkContract` from `@arkade-os/sdk` to avoid coupling test to SDK internals
- Add a test case for the empty state (`contracts.length === 0` → renders "No contracts found.")
- Existing assertions for `screen.getByText('default')` / `screen.getByText('vhtlc')` should still pass since the text nodes remain, but verify after adding the missing fields

---

### 🟡 Nits (non-blocking)

**3. `encodeArkContract` recomputed on every render**
`src/screens/Settings/Contracts.tsx:68`

`encodeArkContract` likely involves string encoding work. On a settings screen with a handful of contracts this is fine, but wrapping it in `useMemo` would be the defensive play — especially if this component ever gets reused in a more render-heavy context.

**4. `key={c.script}` as React key — pre-existing but worth noting**
`src/screens/Settings/Contracts.tsx:79`

If two contracts can share the same `script` (same contract type, different addresses), React will silently drop one. `c.address` would be a safer unique key, or a composite `${c.address}:${c.script}`.

---

### ✅ What's good

- `import type { Contract }` → `import { Contract, encodeArkContract }` — correct, need the runtime value now
- `prettyAgo(contract.createdAt)` — `prettyAgo` gracefully handles falsy (`if (!timestamp) return ''`) and auto-detects ms vs s timestamps. Solid.
- `contract.label ? ... : null` — correctly handles the optional `label` field
- Active/inactive split logic is clean: `=== 'active'` / `!== 'active'` is exhaustive
- `Section` returning `null` when empty prevents rendering empty headers
- Replacing inline `cardStyle` with `<Shadow border>` is a good consistency win

LGTM with the error handling fix and test updates. 🫡

—Arkana

tiero:
Is this a dev gated feature? Curious if not if it aligns with new redesign of @sahilc0

pietro909:
@tiero it's dev-gated. The main issue was to expose the contract params but since I was there I decided to revamp a bit the UI. Not a blocker from my side.

ghost (approved):
**Re-review after `60ee259f`** — all three issues from my previous review are resolved:

1. ✅ `encodeArkContract` wrapped in `useMemo` + `try/catch` — malformed contracts no longer white-screen
2. ✅ `createdAt` fallback to `'Unknown'` when absent
3. ✅ React key changed to `c.address` (unique per contract)
4. ✅ Tests updated with `params`/`createdAt` fields + empty-state test added

Clean UI-only change. No protocol-critical paths touched. LGTM.

—Arkana

bordalix:
@sahilc0 we need to have this on the new design.
And I think Boltz swaps are also useful in the new design.
Maybe add new options to Settings / Advanced?

bordalix (approved):

PR #646, opened 2026-06-03
fix typescript build errors (tsc -b)

bordalix:
## Release Notes

* **Bug Fixes**
  * Simplified recovery locktime progress calculation to rely on time-based countdowns only.

* **Improvements**
  * Enhanced animation easing curves across page transitions and overlay effects.
  * Refined notification connection logic for improved reliability.


ghost (changes requested):
## Code Review — arkade-os/wallet#646

PR fixes TypeScript build errors (`tsc -b`). Most changes are correct and necessary. A few issues need addressing.

---

### 🔴 Must Fix

**1. `tsconfig.tsbuildinfo` should NOT be committed** — `tsconfig.tsbuildinfo`

This is a compiler cache artifact. It's machine-specific, changes on every build, and will cause noisy diffs on every future PR. Add it to `.gitignore` and remove it from this PR:
```
echo 'tsconfig.tsbuildinfo' >> .gitignore
git rm --cached tsconfig.tsbuildinfo
```

**2. `Partial<Tx>` weakens type safety unnecessarily** — `src/lib/format.ts:121-126`

Both `isIssuance` and `isBurn` access `tx.type` and `tx.amount` unconditionally. With `Partial<Tx>`, these become `string | undefined` and `number | undefined`, meaning the functions silently return `false` on incomplete data instead of surfacing a type error at the call site.

All production callers (`TransactionsList.tsx:39-40`, `Transaction.tsx:36-37`) pass full `Tx` objects. The only motivation is making tests pass without providing all required fields — but the tests in this PR already provide `type`, `amount`, and `assets`, so they'd pass with `Pick<Tx, 'type' | 'amount'> & Partial<Pick<Tx, 'assets'>>` too.

Preferred fix: keep the signature as `Tx`, and in the test file, cast partial objects with `as Tx` (which is the existing pattern for test mocks in this codebase). This keeps production call sites type-safe. If you really want a narrower type, use `Pick<Tx, 'type' | 'amount' | 'assets'>` — not `Partial<Tx>`.

---

### 🟡 Nits / Minor

**3. `@ts-expect-error` on `timingSafeEqual`** — `functions/_middleware.ts:36`

The comment is fine — `crypto.subtle.timingSafeEqual` is indeed a Cloudflare runtime extension not in the standard `SubtleCrypto` typings. But a cleaner fix would be to add a Cloudflare Workers type package (`@cloudflare/workers-types`) to devDependencies. Low priority, not blocking.

**4. `BigInt(-5_000)` vs `-5_000n`** — `src/test/screens/wallet/transaction.test.tsx:257`

The original code used `-5_000n` (bigint literal). The PR changes it to `BigInt(-5_000)`. This is functionally equivalent but the literal form is more idiomatic. Was this changed to fix a build error? If so, that suggests a tsconfig issue with the `target` setting rather than a code issue.

---

### ✅ Looks Good

- **`EASE_OUT_QUINT` → `EASE_OUT_QUINT_TUPLE`** (`src/lib/animations.ts`): Correct fix. Framer Motion expects `[number, number, number, number]` tuple, not `number[]`. Clean.

- **`config.nostr` → `config.nostrBackup`** (`src/providers/notifications.tsx:59,67`): This is a **bug fix**. The `Config` type has no `nostr` field — only `nostrBackup`. The old code was accessing an undefined property (always falsy), meaning the relay connection logic was effectively dead code. Good catch.

- **Boltz `blocksAway` removal** (`src/screens/Apps/Boltz/Settings.tsx:341-342`): Acceptable. The removed branch handled block-height-based locktimes (`< 500_000_000`). If the SDK no longer provides `currentBlockHeight` on `SubmarineRecoveryInfo`, this dead code was causing a type error. The UI degrades gracefully (shows "try again later" instead of "X blocks to go").

- **Mock updates** (`src/test/screens/mocks.ts`, `contracts.test.tsx`): Correct — adds required fields (`createdAt`, `params`, `Unit`, `lnurlInfo`, etc.) to match updated type definitions. The mock now properly satisfies the full context type shapes.

- **`vite-plugin-eslint.d.ts`**: Clean ambient module declaration. Fine.

- **Test BigInt updates** (`format.test.ts`): Asset amounts changed from `number` to `BigInt` to match updated `Asset` type. Correct.

- **`pnpm-lock.yaml`**: Just dependency resolution bumps from `@testing-library/dom`. Fine.

---

**Summary**: Two real issues — the committed build artifact and the `Partial<Tx>` type weakening. The rest is solid cleanup work. Fix those two and this is good to merge.

ghost (approved):
## Re-review — new commit `fa62417` ("arkana fixes")

Both must-fix items from my previous review are resolved:

1. ✅ **`tsconfig.tsbuildinfo` removed** from the repo and added to `.gitignore`.
2. ✅ **`Partial<Tx>` reverted to `Tx`** in `src/lib/format.ts:121,125`. Tests now use `as Tx` casts — correct pattern for partial test fixtures.

### Minor observation (not blocking)

The test for "sent tx with amount 0 but no assets" changed from `assets: []` to `assets: [{} as Asset]`. This no longer tests the empty-array path explicitly, but the first assertion in that same test (`{ type: 'sent', amount: 0 } as Tx`) exercises the `tx.assets ?? []` nullish coalescing → empty array → `some()` returns false path, so coverage is preserved. Fine.

Previous nits (not blocking, unchanged):
- `@ts-expect-error` on `timingSafeEqual` (`functions/_middleware.ts:36`) — acceptable, `@cloudflare/workers-types` would be cleaner long-term
- `BigInt(-5_000)` vs `-5_000n` (`src/test/screens/wallet/transaction.test.tsx:257`) — functionally equivalent, idiomatic form preferred but not worth blocking

No protocol-critical changes in this PR. LGTM.

PR #652, opened 2026-06-04
feat(settings): move advanced tools into settings

sahilc0:
## Summary

- Removes the root Apps surface and the Apps pill-nav tab so the app no longer exposes a standalone Apps screen.
- Moves Arkade Mint and Boltz into Settings > Advanced, matching their developer/debug-oriented usage.
- Renames the main settings group and page from general to display, and moves Advanced above Support in that group.
- Updates navigation ownership so Mint and Boltz routes live under Settings, while remaining app detail routes that are still launched from wallet upsells/deep links stay reachable.
- Refreshes bun.lock to match the dependency versions declared by the PR 627 base after running bun install.

## Testing

- bun run format:check
- bun run lint
- bun run test:unit -- src/test/screens/settings/advanced.test.tsx src/test/App.test.tsx
- bun run build
- Verified the portless dev URL returns 200

## Visual evidence

- The local app was run at the branch portless URL. Browser screenshot capture was not attached because the in-app browser automation policy blocked opening the .localhost URL for screenshot capture.

- Settings menu/options: src/providers/options.tsx, src/screens/Settings/Menu.tsx, src/screens/Settings/General.tsx, src/lib/types.ts
- Navigation: src/providers/navigation.tsx, src/components/PillNavbar.tsx, src/components/PillNavbarOverlay.tsx
- Removed root Apps surface: src/screens/Apps/Index.tsx, src/icons/Apps.tsx
- Tests/helpers: src/test/App.test.tsx, src/test/screens/settings/advanced.test.tsx, src/test/e2e/utils.ts, src/test/e2e/keyboard.test.ts, src/test/e2e/nostr.test.ts


ghost (commented):
## Code Review — `feat(settings): move advanced tools into settings`

**Reviewer:** Arkana  
**Verdict:** Approve with comments (no protocol-critical concerns)

---

### ✅ What's good

1. **`subNavHandler` guard fix** (`src/providers/navigation.tsx:274`): The new guard `[Pages.Settings, Pages.WalletSettings].includes(screenRef.current) && subNavHandler.canGoBack()` is a legit bug fix. Previously, if a user navigated deep into Settings sub-nav and then jumped to an App page (via deep link or upsell), pressing back would incorrectly delegate to the Settings sub-nav instead of popping the main back stack. Good catch.

2. **Clean separation of concerns**: ArkadeMint and Boltz navigate via main nav (`navigate(Pages.AppAssets/AppBoltz)` in `Menu.tsx`) rather than sub-nav (`setOption()`), which correctly keeps them as full-page transitions while preserving the Settings sub-nav state for back-navigation. The `getParentOption` logic in `options.tsx` correctly routes back from Advanced-section items to the Advanced screen.

3. **No protocol-critical code touched**: This is purely UI reorganization. No VTXO handling, signing, forfeit paths, or exit paths affected.

4. **No cross-repo breakage**: Confirmed no external repos consume `SettingsSections`, `SettingsOptions`, `Pages`, or `Tabs` enums.

5. **Test coverage**: Advanced screen tests now verify ArkadeMint and Boltz visibility. E2E tests updated for "display" rename. App.test.tsx updated to remove Apps tab assertions.

---

### ⚠️ Issues to address

**1. `Pages.AppDfx` is orphaned — no remaining entry point**

The only navigation to `Pages.AppDfx` was in `Apps/Index.tsx:168`, which this PR deletes. No other file calls `navigate(Pages.AppDfx)` — confirmed by searching the full codebase. Unlike Lendasat and Satora (reachable from `wallet.tsx` and `Announcement.tsx`), DFX has no alternate entry point.

The route still exists in `pageComponent` and `pageTab`, creating dead code. If this is intentional (DFX is being sunset), the route should be cleaned up. If not, DFX needs an entry point — either in Settings > Advanced or in a wallet upsell.

**Files:** `src/providers/navigation.tsx:96,172-173` (dead `AppDfx` route)

**2. `SettingsOptions.General = 'display'` — misleading enum key**

`src/lib/types.ts:69`: The enum key is `General` but the runtime value (which renders as the UI label) is now `'display'`. This works but is confusing to any developer reading the code — the key name contradicts the value. Consider renaming the key to `Display` for clarity:

```ts
// Current (confusing):
General = 'display',

// Suggested:
Display = 'display',
```

This would require updating references in `Settings/Index.tsx:38-39`, `options.tsx`, and `Settings/General.tsx`, but makes the code self-documenting. If the refactor scope is too large for this PR, at minimum add a comment explaining the mismatch.

---

### 📝 Observations (non-blocking)

- **Enum value casing inconsistency** (`types.ts`): `ArkadeMint = 'Arkade Mint'` and `Boltz = 'Boltz'` use title-case values while peers use lowercase (`advanced`, `backup`, `display`). Pre-existing for ArkadeMint; Boltz follows the same pattern. Not a bug but worth normalizing eventually.

- **Dependency bumps** (SDK 0.4.27→0.4.32, boltz-swap 0.3.32→0.3.37, Playwright 1.56→1.60): These align with the base branch. No concerns.

- **`allOptions` in `options.tsx:136`** now maps `[Display, Security]` which correctly excludes Advanced/Config sections from the top-level menu, consistent with pre-existing behavior (just renamed).

---

### Summary

Clean UI refactoring. The orphaned DFX route is the only real issue — everything else is solid. The `subNavHandler` guard change is actually a navigation bug fix in disguise. No protocol-critical concerns.

sahilc0:
Thanks for the review. I addressed the enum clarity issue by renaming the settings option key from `General` to `DisplayMenu` while keeping the existing display preferences option separate.

On DFX: I double-checked the route before changing it. `Pages.AppDfx` is still reachable from the wallet upsell in `src/screens/Wallet/UpsellsSection.tsx` via `navigate(Pages.AppDfx)`, so I kept that route. The deleted surface was only the root Apps catalog/tab, not the individual app detail pages launched from wallet entry points or deep links.

PR #665, opened 2026-06-10
fix(assets): handle scientific notation in prettyAssetNumber

pietro909:
Numbers like -8e-8 passed as JS floats would have the 'e' stripped by the formatting regex, producing '-8-8' which BigInt rejects. Convert to fixed-point via Decimal.toFixed() before splitting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed improper rendering of very small numbers in scientific notation format, now correctly normalized to decimal representation.

* **Tests**
  * Extended test coverage to verify correct formatting of tiny negative values and amounts in scientific notation without formatting artifacts.

### DIFF
--- a/src/lib/assets.ts
+++ b/src/lib/assets.ts
@@ -45,6 +45,7 @@ export const prettyAssetAmountHide = (value: bigint, suffix: string): string => 
 export const prettyAssetNumber = (num?: string | number, maximumFractionDigits = MAX_DECIMALS): string => {
   if (num === undefined || num === null) return '0'
   if (typeof num === 'number') num = num.toString()
+  if (/e/i.test(num)) num = new Decimal(num).toFixed()
   let [integer, fraction = ''] = num.split('.')
   integer = integer.replace(/[^0-9-]+/g, '') // remove non-digit and non-negative sign characters
   const negative = integer === '-0'
--- a/src/test/lib/asset.test.ts
+++ b/src/test/lib/asset.test.ts
@@ -191,6 +191,7 @@ describe('asset utilities', () => {
     it('formats very small number', () => {
       expect(prettyAssetNumber('0.00000001')).toBe('0.00000001')
       expect(prettyAssetNumber('-0.00000001')).toBe('-0.00000001')
+      expect(prettyAssetNumber('-8e-8')).toBe('-0.00000008')
     })
 
     it('formats number with grouping', () => {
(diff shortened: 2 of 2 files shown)


ghost (approved):
## ✅ Approved — clean, minimal, correct fix

### Bug analysis

Traced the full call path:

1. `prettyAssetAmount(BigInt(-8), 8, true)` → `centsToUnits(-8n, 8)` returns `"-0.00000008"`
2. `Number("-0.00000008")` → `-8e-8` (JS float)
3. `prettyAssetNumber(-8e-8, 8)` → `.toString()` produces `"-8e-8"`
4. **Without fix:** `"-8e-8".split(".")` → `["-8e-8", ""]`, then `replace(/[^0-9-]+/g, "")` strips the `e` → `"-8-8"` → `BigInt("-8-8")` **throws**

The fix (`new Decimal(num).toFixed()`) converts `"-8e-8"` → `"-0.00000008"` before the split. Correct.

### What I verified

- **`Decimal` is already imported** at line 1 — no new dependency.
- **`Decimal.toFixed()` without args** uses full precision — no silent truncation.
- **Guard placement is correct** — runs after `toString()` (line 48) but before `split(".")` (line 50).
- **The regex `/e/i`** only fires on scientific notation strings. In practice, inputs are from `centsToUnits` (well-formed decimals) or `Number.toString()` (scientific notation for very small/large values). No false-positive risk from the callers in this codebase.
- **Cross-repo impact: none.** `prettyAssetNumber` is internal to the wallet UI — not exported in any shared package.
- **Not protocol-critical.** This is display formatting, not VTXO/signing/forfeit logic. The bug was a crash (thrown exception), not a silent incorrect value.
- **Test coverage is good.** Tests cover both the direct case (`prettyAssetNumber('-8e-8')`) and the integration path (`prettyAssetAmount(BigInt(-8), 8, true)`).

### One minor note (non-blocking)

The regex `/e/i.test(num)` would also match a hypothetical non-numeric string containing the letter 'e' (e.g., `"beef"`), which would then throw inside `new Decimal()`. This isn't a regression — the existing code would also throw on such input at the `BigInt()` call — but if you ever want to harden `prettyAssetNumber` against arbitrary string input, a `try/catch` around the Decimal conversion would be the place.

LGTM. Ship it.

bordalix (approved):

gringokiwi (approved):
lgtm, tested on https://fix-pretty-asset-number-scie.wallet-bitcoin.pages.dev/

PR #666, opened 2026-06-11
fix: debounce send input validation for LNURL/LN addresses

### LINKED ISSUE #562: Debounce send input validation
First, only run LN URL verification if the string begins with `lnurl`, or if the user pastes.

Second, only check for a Lightning address if it's full address with TLD (e.g. user@example.xxx), and debounce the validation by ~400ms or so to allow for scenarios where someone is typing in the address.

https://github.com/user-attachments/assets/81a8835b-2ec8-4551-955d-24ad7f301a09

ghost:
## Summary

- Recipient parsing/validation (`parseRecipient`) no longer runs on every keystroke. It runs only on explicit triggers: **field blur, paste (button or native Ctrl+V), QR scan, clearing the field, Enter (when nothing was parsed yet), or asset change**. Typing just updates state, so no network calls fire per keystroke and no "Invalid recipient address" banner appears mid-typing.
- Since tapping the amount field or the Continue button always blurs the recipient input first, parsing completes before any subsequent input — this also fixes the e2e regression where a delayed parse reset `sendInfo` and wiped the amount the test had just typed (every send e2e test timed out; the Playwright job hit the 45m limit on earlier commits of this PR).
- `InputWithScanner` routes the change event following a native paste through `onPaste`, so Ctrl+V resolves immediately like the Paste button.
- Adds `onBlur`/`onPaste` props to `InputWithScanner` and `InputAddress`.
- Comment on `isLnAddress` clarifies the TLD requirement (2+ chars); tested via the public `isValidLnUrl`.

## Test plan

- [x] `src/test/lib/lnurl.test.ts` — `isValidLnUrl` edge cases (TLD requirements, partial inputs, non-bech32 strings)
- [x] `src/test/screens/wallet/send.test.tsx` — typing never triggers validation (even after time passes); blur validates once with the full value; native paste validates immediately
- [ ] Manual: type a Lightning address — no error while typing; error/validation appears on leaving the field
- [ ] Manual: paste a Lightning address (button or Ctrl+V) and confirm it resolves immediately
- [ ] Playwright e2e suite green (was cancelled at 45m on previous commits)

Closes #562

### DIFF
--- a/src/components/InputWithScanner.tsx
+++ b/src/components/InputWithScanner.tsx
@@ -7,6 +7,7 @@ import FlexRow from './FlexRow'
 
 interface InputWithScannerProps {
   error?: string
+  errorVariant?: 'banner' | 'inline'
   focus?: boolean
   label?: string
   name?: string
@@ -22,6 +23,7 @@ interface InputWithScannerProps {
 
 export default function InputWithScanner({
   error,
+  errorVariant,
   focus,
   label,
   name,
@@ -70,7 +72,7 @@ export default function InputWithScanner({
   const hasValue = Boolean(value && value.length > 0)
 
   return (
-    <InputContainer label={label} error={error}>
+    <InputContainer label={label} error={error} errorVariant={errorVariant}>
       <label className='label has-buttons'>
         <input
           ref={input}
--- a/src/components/InputWithScanner.tsx
+++ b/src/components/InputWithScanner.tsx
@@ -13,6 +13,7 @@ interface InputWithScannerProps {
   onBlur?: () => void
   onChange: (arg0: any) => void
   onEnter?: () => void
+  onFocus?: () => void
   onPaste?: (data: string) => void
   openScan: () => void
   placeholder?: string
@@ -28,6 +29,7 @@ export default function InputWithScanner({
   onBlur,
   onChange,
   onEnter,
+  onFocus,
   onPaste,
   openScan,
   placeholder,
@@ -79,6 +81,7 @@ export default function InputWithScanner({
           className='input'
           onBlur={onBlur}
           onChange={handleChange}
+          onFocus={onFocus}
           onPaste={handleNativePaste}
           placeholder={placeholder}
           onKeyUp={(ev) => ev.key === 'Enter' && onEnter && onEnter()}
--- a/src/components/InputWithScanner.tsx
+++ b/src/components/InputWithScanner.tsx
@@ -12,6 +12,7 @@ interface InputWithScannerProps {
   name?: string
   onChange: (arg0: any) => void
   onEnter?: () => void
+  onPaste?: (data: string) => void
   openScan: () => void
   placeholder?: string
   validator?: (arg0: string) => boolean
@@ -25,6 +26,7 @@ export default function InputWithScanner({
   name,
   onChange,
   onEnter,
+  onPaste,
   openScan,
   placeholder,
   validator,
@@ -41,7 +43,11 @@ export default function InputWithScanner({
   }
 
   const handlePaste = (data: string) => {
-    onChange(data)
+    if (onPaste) {
+      onPaste(data)
+    } else {
+      onChange(data)
+    }
   }
 
   const handleClear = () => {
--- a/src/components/InputWithScanner.tsx
+++ b/src/components/InputWithScanner.tsx
@@ -33,22 +33,27 @@ export default function InputWithScanner({
   value,
 }: InputWithScannerProps) {
   const input = useRef<HTMLInputElement>(null)
+  const nativePasteRef = useRef(false)
 
   useEffect(() => {
     if (focus && input.current) input.current.focus()
   }, [focus, input.current])
 
   const handleChange: ChangeEventHandler<HTMLInputElement> = (ev) => {
-    onChange(ev.currentTarget.value)
+    const wasPaste = nativePasteRef.current
+    nativePasteRef.current = false
+    if (wasPaste && onPaste) onPaste(ev.currentTarget.value)
+    else onChange(ev.currentTarget.value)
   }
 
-  const handlePaste = (data: string) => {
-    if (onPaste) {
-      onPaste(data)
-    } else {
-      onChange(data)
-    }
+  const handleNativePaste = () => {
+    nativePasteRef.current = true
+    // the change event fires synchronously after the paste is applied;
+    // if the value didn't change (no change event), don't leak the flag
+    setTimeout(() => (nativePasteRef.current = false), 0)
   }
+
+  const handlePaste = (data: string) => (onPaste ?? onChange)(data)
 
   const handleClear = () => {
     hapticLight()
@@ -71,6 +76,7 @@ export default function InputWithScanner({
           value={value}
           className='input'
           onChange={handleChange}
+          onPaste={handleNativePaste}
           placeholder={placeholder}
           onKeyUp={(ev) => ev.key === 'Enter' && onEnter && onEnter()}
         />
--- a/src/components/InputWithScanner.tsx
+++ b/src/components/InputWithScanner.tsx
@@ -7,6 +7,7 @@ import FlexRow from './FlexRow'
 
 interface InputWithScannerProps {
   error?: string
+  errorVariant?: 'banner' | 'inline'
   focus?: boolean
   label?: string
   name?: string
@@ -23,6 +24,7 @@ interface InputWithScannerProps {
 
 export default function InputWithScanner({
   error,
+  errorVariant,
   focus,
   label,
   name,
@@ -72,7 +74,7 @@ export default function InputWithScanner({
   const hasValue = Boolean(value && value.length > 0)
 
   return (
-    <InputContainer label={label} error={error}>
+    <InputContainer label={label} error={error} errorVariant={errorVariant}>
       <label className='label has-buttons'>
         <input
           ref={input}
--- a/src/components/InputWithScanner.tsx
+++ b/src/components/InputWithScanner.tsx
@@ -10,6 +10,7 @@ interface InputWithScannerProps {
   focus?: boolean
   label?: string
   name?: string
+  onBlur?: () => void
   onChange: (arg0: any) => void
   onEnter?: () => void
   onPaste?: (data: string) => void
@@ -24,6 +25,7 @@ export default function InputWithScanner({
   focus,
   label,
   name,
+  onBlur,
   onChange,
   onEnter,
   onPaste,
@@ -75,6 +77,7 @@ export default function InputWithScanner({
           name={name}
           value={value}
           className='input'
+          onBlur={onBlur}
           onChange={handleChange}
           onPaste={handleNativePaste}
           placeholder={placeholder}
--- a/src/lib/lnurl.ts
+++ b/src/lib/lnurl.ts
@@ -62,8 +62,8 @@ const isLnUrl = (data: string) => {
   return data.toLowerCase().startsWith('lnurl') && isValidBech32(data)
 }
 
-// Exported for testing. Requires a valid email-like address with TLD (2+ chars after the last dot).
-export const isLnAddress = (data: string) => {
+// Requires a valid email-like address with TLD (2+ chars after the last dot).
+const isLnAddress = (data: string) => {
   return data.includes('@') && emailRegex.test(data)
 }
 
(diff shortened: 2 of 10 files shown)


ghost (commented):
## ✅ Arkana Code Review — LGTM

**PR**: fix: debounce send input validation for LNURL/LN addresses
**Risk level**: Low — UI-only change, no protocol-critical code touched.

### What this does

Adds a 400ms debounce to `parseRecipient` in the send form `useEffect`, preventing network calls (`checkLnUrlConditions`, etc.) from firing on every keystroke. The Paste button bypasses debounce (0ms delay) for instant feedback.

### Review findings

**Implementation is correct.** Traced the full data flow:

1. **Debounce lifecycle is sound.** The cleanup function (`return () => clearTimeout(...)` at `Form.tsx:304-306`) properly cancels stale timeouts on re-render. Each keystroke clears the previous timer and starts a fresh 400ms delay. No race conditions.

2. **Ref usage is appropriate.** `debounceRef` (timeout ID) and `isPasteRef` (paste flag) are correctly `useRef` — they don't trigger re-renders and survive across effect invocations.

3. **Empty-recipient fast path** (`Form.tsx:296-298`): When `recipient` is falsy, `parseRecipient()` is called synchronously (no debounce). Correct — no need to debounce clearing state.

4. **Error UX**: `smartSetError('')` clears errors immediately at effect entry (line 205), before the debounce fires. Stale error messages disappear instantly while the user types; new errors only appear after 400ms of inactivity. Good UX.

5. **`onPaste` prop is optional and backward-compatible.** Checked all consumers of `InputWithScanner` (`InputNpub`, `InputNote`, `InputUrl`, `InputAssetId`) — none pass `onPaste`, so they fall through to the `else { onChange(data) }` branch in `handlePaste` (`InputWithScanner.tsx:48`). No breakage.

6. **`isLnAddress` export** is additive and only consumed by tests. No cross-repo impact.

7. **Tests are solid.** The debounce test correctly uses fake timers, simulates character-by-character typing, asserts `isValidLnUrl` is NOT called during typing, then asserts it IS called exactly once after 400ms advance.

### Minor observation (non-blocking)

**Native Ctrl+V paste is NOT distinguished from typing.** The `onPaste` callback only fires from the `<Paste>` button component (`Paste.tsx:35`), not from the browser's native paste event on the `<input>`. If a user does Ctrl+V instead of tapping the Paste button, the input goes through `handleChange` → `onChange` → `handleRecipientChange` (which sets `isPasteRef = false`), so the 400ms debounce still applies. This is a UX nuance, not a bug — the address still resolves after 400ms. Worth noting for a future enhancement if users report sluggish paste via keyboard shortcuts.

### Verdict

Clean, focused fix. Debounce logic is correct, tests cover the behavior, no protocol-critical paths are affected, no cross-repo breakage. Ready for human approval.

🤖 — Arkana

src/test/screens/wallet/send.test.tsx
```diff
@@ -50,6 +105,33 @@ describe('Send screen', () => {
     expect(screen.getByTestId('loading-logo')).toBeInTheDocument()
   })
   it('renders the send screen correctly', async () => {
+    renderSendForm()
+    // find text elements
+    expect(screen.getByText('Max')).toBeInTheDocument()
+    expect(screen.getByText('Send')).toBeInTheDocument()
+    expect(screen.getByText('Amount')).toBeInTheDocument()
+    expect(screen.getByText('€0.00 available')).toBeInTheDocument()
```
tiero: **This test fails as committed.** `mockConfigContextValue` in `src/test/screens/mocks.ts` has `useFiat: false`, so the screen renders `0 SATS available`, not `€0.00 available`. Verified by running `vitest run src/test/screens/wallet/send.test.tsx` on this branch:

```
FAIL  'renders the send screen correctly'
TestingLibraryElementError: Unable to find an element with the text: €0.00 available
```

Either restore the original assertion (`'0 SATS available'`) or set `useFiat: true` in the mock, whichever the test intends.

---

src/screens/Wallet/Send/Form.tsx
```diff
@@ -285,7 +292,18 @@ export default function SendForm() {
...
       setError('Invalid recipient address')
     }
-    parseRecipient()
+
+    if (!recipient) {
+      parseRecipient()
+      return
+    }
+
+    const delay = isPasteRef.current ? 0 : 400
```
tiero: The 0ms fast path only covers the Paste **button** — two other atomic, complete inputs still get the 400ms typing debounce:

1. **QR scans**: `Scanner onData` (line ~781) calls `setRawScanData(data)` + `resetDerivedState(data)` without setting `isPasteRef`, so a scanned recipient isn't parsed for 400ms (Continue stays disabled, and the Branta effect keyed on `rawScanData` fires *before* the recipient is parsed — the reverse of the old ordering).
2. **Native paste (Ctrl+V / right-click)**: the `<input>` in `InputWithScanner` has no `onPaste` DOM handler, so a keyboard paste fires `onChange` → `handleRecipientChange` → `isPasteRef = false` → 400ms. The PR's test plan says "paste a Lightning address and confirm it resolves immediately", but the most common paste gesture misses the fast path.

A related side effect of the delay window: pressing Enter within 400ms of input no-ops, because `handleEnter` (line ~641) sees `buttonDisabled === true` while the parsed fields are still empty, and moves focus instead of submitting.

Consider adding an `onPaste` DOM handler on the input and setting the flag in the scanner's `onData` (or treating any multi-character value jump as paste-like) so all complete-input paths bypass the debounce.

---

src/screens/Wallet/Send/Form.tsx
```diff
@@ -575,7 +593,14 @@ export default function SendForm() {
     setLnUrlResponse(undefined)
   }
 
+  const handlePasteRecipient = (data: string) => {
```
tiero: `handlePasteRecipient` and `handleRecipientChange` are identical except for the `isPasteRef` value. A single handler would remove the duplication and give the scanner path a way to opt into the fast path too:

```ts
const handleRecipientInput = (data: string, isPaste = false) => {
  isPasteRef.current = isPaste
  setRawScanData('')
  resetDerivedState(data)
}
```

---

src/test/screens/wallet/send.test.tsx
```diff
@@ -50,6 +105,33 @@ describe('Send screen', () => {
...
+    vi.clearAllMocks()
+  })
+
+  it('debounces parseRecipient: isValidLnUrl not called immediately while typing, called once after 400ms', async () => {
+    const lnurl = await import('../../../lib/lnurl')
+    const mockIsValidLnUrl = vi.mocked(lnurl.isValidLnUrl)
+    mockIsValidLnUrl.mockClear()
+
     render(
       <NavigationContext.Provider value={mockNavigationContextValue}>
```
tiero: This test rebuilds the entire 10-level provider pyramid inline even though `renderSendForm()` was just added above (line 55) — the only difference is `mockSvcWalletWithAddresses` vs `mockSvcWallet`. Parameterizing the helper removes the second copy:

```ts
function renderSendForm(svcWallet = mockSvcWallet) { ... }
```

(`receive-qrcode.test.tsx` follows this pattern with `buildTree()`/overrides.)

---

src/test/screens/wallet/send.test.tsx
```diff
@@ -22,6 +22,61 @@ import { ConfigContext } from '../../../providers/config'
 import { FiatContext } from '../../../providers/fiat'
 import { SwapsContext } from '../../../providers/swaps'
 import { OptionsContext } from '../../../providers/options'
+import { FeesContext } from '../../../providers/fees'
+
+const mockFeesContextValue = {
```
tiero: All the other context mocks (`mockWalletContextValue`, `mockConfigContextValue`, …) live in `src/test/screens/mocks.ts` and are imported. Consider moving `mockFeesContextValue` (and `mockSvcWalletWithAddresses`) there so other Send-related tests can reuse them.

---

src/components/InputWithScanner.tsx
```diff
@@ -41,7 +43,11 @@ export default function InputWithScanner({
   }
 
   const handlePaste = (data: string) => {
-    onChange(data)
+    if (onPaste) {
```
tiero: Nit: this reduces to a single expression:

```ts
const handlePaste = (data: string) => (onPaste ?? onChange)(data)
```

---

src/lib/lnurl.ts
```diff
@@ -62,7 +62,8 @@ const isLnUrl = (data: string) => {
   return data.toLowerCase().startsWith('lnurl') && isValidBech32(data)
 }
 
-const isLnAddress = (data: string) => {
+// Exported for testing. Requires a valid email-like address with TLD (2+ chars after the last dot).
+export const isLnAddress = (data: string) => {
```
tiero: Nit: the new tests already exercise the same inputs through the public `isValidLnUrl` (which delegates here), so the test-only export widens the module surface for no runtime caller. Testing via `isValidLnUrl` alone would keep this helper private.

---

tiero (commented):
Reviewed the diff for correctness and cleanup. Summary of findings (details inline):

**Bugs**
1. `send.test.tsx:113` — the rewritten render test **fails as committed** (`€0.00 available` vs the actual `0 SATS available` with `useFiat: false`). Verified by running the suite on this branch.
2. The 0ms paste fast path only covers the Paste button: **QR scans** and **native Ctrl+V/right-click paste** still get the 400ms typing debounce, contradicting the PR's intent ("paste … resolves immediately"). Scans also now race the Branta fetch ahead of recipient parsing.
3. Minor: pressing Enter during the 400ms window silently no-ops instead of submitting.

**Cleanup**
- The new debounce test duplicates the full provider pyramid instead of parameterizing the `renderSendForm` helper added in the same PR.
- `mockFeesContextValue` belongs in `mocks.ts` with the other context mocks.
- `handlePasteRecipient`/`handleRecipientChange` are copy-paste twins differing only in the flag.
- Two nits: `(onPaste ?? onChange)(data)`, and the test-only `isLnAddress` export.

The debounce test itself and the new lnurl tests pass.

---

ghost (commented):
## ✅ Arkana Re-review — LGTM (new commit d731b034)

Reviewed the follow-up commit addressing tiero's feedback. All findings resolved cleanly:

| Feedback | Resolution |
|----------|-----------|
| Failing test (`€0.00` vs `0 SATS`) | Fixed: asserts `0 SATS available` matching `useFiat: false` mock |
| QR scan missing fast path | Fixed: scanner `onData` sets `isPasteRef.current = true` before `resetDerivedState` (`Form.tsx:781`) |
| Native Ctrl+V not routed through fast path | Fixed: `InputWithScanner` tracks `nativePasteRef`, routes through `onPaste` callback → `handleRecipientInput(data, true)` |
| Duplicate handlers | Fixed: merged into single `handleRecipientInput(data, isPaste = false)` (`Form.tsx:601`) |
| Provider pyramid duplication in tests | Fixed: `renderSendForm(svcWallet)` parameterized and reused |
| `isLnAddress` unnecessarily exported | Fixed: un-exported, tests go through public `isValidLnUrl` |
| Nit on `handlePaste` | Fixed: `(onPaste ?? onChange)(data)` one-liner |

**Debounce correctness verified:**
- Typing: 400ms debounce, each keystroke resets timer via cleanup function ✓
- Paste (button): `handlePaste` → `onPaste` → `handleRecipientInput(data, true)` → 0ms delay ✓
- Paste (native Ctrl+V): `onPaste` event → `nativePasteRef` flag → change event routed to `onPaste` → 0ms delay ✓
- QR scan: direct `isPasteRef.current = true` + `resetDerivedState` → 0ms delay ✓
- Empty recipient: synchronous `parseRecipient()` call, no debounce (early return) ✓
- `isPasteRef` reset inside `parseRecipient` prevents stale flags ✓
- `setTimeout(() => nativePasteRef.current = false, 0)` safety net prevents flag leak when paste doesn't trigger change ✓

**No protocol-critical code touched.** UI-only debounce logic. CI green. Ship it.

bordalix:
Huge bug: address error goes away after adding amount, allowing to continue.

bordalix:
@tiero this PR introduces too many bugs!
What's exactly the problem you're trying to tackle?
I'm pretty sure there's a simpler way to fix it.

tiero:
this one @bordalix  https://github.com/arkade-os/wallet/issues/562

PR #671, opened 2026-06-13
fix(send): populate amount field for fixed-amount LNURL

pietro909:
## Problem

When an LNURL-pay endpoint resolves to a **fixed amount** (`minSendable === maxSendable`), the Send form locked the amount input but left it **blank**. The user saw an empty, read-only field and any attempt to continue produced an unpayable invoice.

## Cause

The fixed amount was applied via `setSendInfo` / `setAmount`, neither of which syncs `amountTextValue` — the state the amount `<input>` is actually bound to. So the value never reached the field.

## Fix

Route the auto-filled fixed amount through `setState({ ...sendInfo, satoshis: min })`, which updates `amountTextValue` and makes the locked field display the correct value. Removed the dead `setAmount(min)` call in the read-only branch.

## Test

Adds a regression test that mocks a fixed-amount LNURL response (`minSendable === maxSendable === 21000` msat → 21 sats) and asserts the amount input is **populated with `21`** and **read-only**.

- `pnpm test` — 268 passed / 4 skipped
- `pnpm lint`, `pnpm format:check` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed LNURL fixed-amount handling to ensure the send amount input properly syncs with and displays the fixed satoshi amount specified by the payment provider, maintaining consistency between the UI display and internal state, with the field marked as read-only when applicable.

* **Tests**
  * Added regression test to verify correct behavior and display of the send amount input when LNURL payments specify a fixed amount.

### DIFF
--- a/src/screens/Wallet/Send/Form.tsx
+++ b/src/screens/Wallet/Send/Form.tsx
@@ -352,7 +352,6 @@ export default function SendForm() {
     if (satoshis && satoshis < min) return setError(`Amount below LNURL min limit`)
     if (satoshis && satoshis > max) return setError(`Amount above LNURL max limit`)
     if (min === max) {
-      setAmount(min) // set fixed amount automatically
       setAmountIsReadOnly(true)
     } else {
       setAmountIsReadOnly(false)
@@ -369,7 +368,9 @@ export default function SendForm() {
         if (!conditions) return setError('Unable to fetch LNURL conditions')
         const min = Math.floor(conditions.minSendable / 1000) // from millisatoshis to satoshis
         const max = Math.floor(conditions.maxSendable / 1000) // from millisatoshis to satoshis
-        if (min === max) setSendInfo({ ...sendInfo, satoshis: min }) // set amount automatically
+        // when the LNURL resolves to a fixed amount, set it via setState so the
+        // amount input (bound to amountTextValue) reflects it instead of staying blank
+        if (min === max) setState({ ...sendInfo, satoshis: min })
         return setLnUrlResponse({ ...conditions, minSendable: min, maxSendable: max })
       })
       .catch((e) => {
--- a/src/test/screens/wallet/send.test.tsx
+++ b/src/test/screens/wallet/send.test.tsx
@@ -1,6 +1,7 @@
-import { describe, expect, it } from 'vitest'
-import { render, screen } from '@testing-library/react'
-import { FlowContext } from '../../../providers/flow'
+import { describe, expect, it, vi } from 'vitest'
+import { render, screen, waitFor } from '@testing-library/react'
+import createFetchMock from 'vitest-fetch-mock'
+import { emptySendInfo, FlowContext } from '../../../providers/flow'
 import { LimitsContext } from '../../../providers/limits'
 import {
   mockAspContextValue,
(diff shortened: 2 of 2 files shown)


ghost (approved):
## Review: fix(send): populate amount field for fixed-amount LNURL

**Verdict: APPROVE** ✅

### Analysis

**Root cause is correctly identified.** Two effects manage fixed-amount LNURL state:

1. **"check lnurl conditions"** (Form.tsx ~L410): fetches LNURL pay endpoint, and if `min === max`, sets `sendInfo.satoshis`
2. **"check lnurl limits"** (Form.tsx ~L393): fires when `lnUrlResponse` changes, sets `amountIsReadOnly`

The old code called `setSendInfo()` directly, which updates context state but never touches `amountTextValue` — the string bound to the `<input>`. Only the local `setState()` helper (L170–183) syncs both. Classic React "two sources of truth" bug.

**The fix routes through `setState()` instead of `setSendInfo()`.** This is correct — `setState` calls `setSendInfo` internally AND sets `amountTextValue` via the `info.satoshis` branch (L179).

### What I checked

- **`setAmount(min)` removal (L402)**: Safe. The `amount` state is only read at L675 for keyboard focus heuristics (`handleEnter`). For fixed-amount LNURL the field is readonly so this path is irrelevant.
- **`setState` side effects**: `setState` also calls `setScan(false)` (L171). Harmless — scan is already false when LNURL conditions resolve.
- **Stale closure on `sendInfo`**: The `.then()` callback at L419 captures `sendInfo` from render time. If the user mutates state while the fetch is in-flight, the spread could overwrite changes. This is a **pre-existing issue** — the old `setSendInfo({...sendInfo, satoshis: min})` had the identical closure problem. Not introduced by this PR.
- **Zero-amount edge case**: If an LNURL returns `minSendable === maxSendable === 0`, `setState` won't update `amountTextValue` because `if (info.satoshis)` is falsy for 0. Pre-existing, and 0-sat LNURL is invalid per spec (LUD-06 requires `minSendable >= 1`).
- **Effect ordering**: Conditions effect calls `setState` then `setLnUrlResponse`. React batches these. The limits effect fires next render, sees `min === max`, sets `amountIsReadOnly(true)`. Correct sequence.
- **No protocol-critical paths touched**: This is pure UI state management — no VTXO handling, signing, forfeit, or exit paths.

### Test

The regression test is well-constructed:
- Mocks `fetch` with a fixed-amount LNURL response (`minSendable === maxSendable === 21000 msat`)
- Pre-populates `sendInfo.lnUrl` and `sendInfo.recipient` to trigger the conditions effect
- Asserts `getByDisplayValue('21')` + `readonly` attribute
- Properly cleans up with `fetchMocker.disableMocks()`

Clean fix. No issues found.

---
🤖 Reviewed by Arkana

ghost (approved):
## Review: fix(send): populate amount field for fixed-amount LNURL

**Verdict: APPROVE** ✅

### Analysis

**Root cause is correctly identified.** Two effects manage fixed-amount LNURL state:

1. **"check lnurl conditions"** (Form.tsx ~L410): fetches LNURL pay endpoint, and if `min === max`, sets `sendInfo.satoshis`
2. **"check lnurl limits"** (Form.tsx ~L393): fires when `lnUrlResponse` changes, sets `amountIsReadOnly`

The old code called `setSendInfo()` directly, which updates context state but never touches `amountTextValue` — the string bound to the `<input>`. Only the local `setState()` helper (L170–183) syncs both. Classic React "two sources of truth" bug.

**The fix routes through `setState()` instead of `setSendInfo()`.** This is correct — `setState` calls `setSendInfo` internally AND sets `amountTextValue` via the `info.satoshis` branch (L179).

### What I checked

- **`setAmount(min)` removal (L402)**: Safe. The `amount` state is only read at L675 for keyboard focus heuristics (`handleEnter`). For fixed-amount LNURL the field is readonly so this path is irrelevant.
- **`setState` side effects**: `setState` also calls `setScan(false)` (L171). Harmless — scan is already false when LNURL conditions resolve.
- **Stale closure on `sendInfo`**: The `.then()` callback at L419 captures `sendInfo` from render time. If the user mutates state while the fetch is in-flight, the spread could overwrite changes. This is a **pre-existing issue** — the old `setSendInfo({...sendInfo, satoshis: min})` had the identical closure problem. Not introduced by this PR.
- **Zero-amount edge case**: If an LNURL returns `minSendable === maxSendable === 0`, `setState` won't update `amountTextValue` because `if (info.satoshis)` is falsy for 0. Pre-existing, and 0-sat LNURL is invalid per spec (LUD-06 requires `minSendable >= 1`).
- **Effect ordering**: Conditions effect calls `setState` then `setLnUrlResponse`. React batches these. The limits effect fires next render, sees `min === max`, sets `amountIsReadOnly(true)`. Correct sequence.
- **No protocol-critical paths touched**: This is pure UI state management — no VTXO handling, signing, forfeit, or exit paths.

### Test

The regression test is well-constructed:
- Mocks `fetch` with a fixed-amount LNURL response (`minSendable === maxSendable === 21000 msat`)
- Pre-populates `sendInfo.lnUrl` and `sendInfo.recipient` to trigger the conditions effect
- Asserts `getByDisplayValue('21')` + `readonly` attribute
- Properly cleans up with `fetchMocker.disableMocks()`

Clean fix. No issues found.

---
🤖 Reviewed by Arkana

PR #672, opened 2026-06-15
fix(receive/send): clipboard correctness, amountless LNURL, BIP21 case handling

Kukks:
## Summary

Fixes three reported bugs on the Receive and Send pages. Scoped to `src/lib/bip21.ts`, `src/screens/Wallet/Receive/QrCode.tsx`, and their tests.

### Receive — wrong value copied to clipboard
The BIP21-build effect called `setQrCodeValue(bip21)` on **every** async dependency change (LNURL session re-emit, invoice arrival, swap-address update). When a user picked a specific address in the "Copy address" sheet, the next effect fire silently reverted the QR back to the unified URI — so the bottom **Copy** button and QR tap then copied the wrong thing (timing-dependent → "sometimes").

- Extracted a pure `resolveQrValue()` that honors an explicit selection **while it is still on offer**, falling back to the unified URI only when the selected value is regenerated/removed.
- Tapping a copy-sheet **row** now copies that address **and** switches the QR (previously it switched without copying).
- `encodeBip21` formats the amount with `useGrouping: false`, so large amounts no longer produce a malformed `amount=1,000`.

### Receive — LNURL lingers after an amount is set
`isAmountlessLnurl` was computed but never applied: `createBip21()` and the copy sheet received `lnurlSession.lnurl` unconditionally. Now gated through `displayLnurl` — once an amount is set, LNURL is dropped from the QR and the copy list. **The LNURL provider session is untouched and keeps running underneath.**

### Send — BIP21 case handling ("finnicky")
`decodeBip21` only matched query keys that were *exactly* lowercase or *exactly* uppercase, so mixed-case keys from QR codes (`Amount`, `Ark`, `Lightning`, `AssetId`) were silently dropped. Replaced with a case-insensitive `getParam()`. The Arkade → LN → BTC preference is already enforced downstream and is unchanged.

## Test plan

- `src/test/lib/bip21.test.ts`: mixed-case key decode (`Ark`/`Lightning`/`Amount`), and no-thousands-separator encode.
- `src/test/screens/wallet/receive-qrcode.test.tsx`: LNURL present when amountless / dropped once an amount is set; `resolveQrValue` selection-persistence + stale-fallback.
- Each new test was confirmed to fail without its fix.
- Full unit suite: 284 passed / 4 skipped, 0 failures. `tsc --noEmit`, `eslint`, `prettier --check` all clean on the changed files.

## Not verifiable in CI (needs a real device)
The copy-sheet row interaction runs through vaul's portaled `Drawer`, which jsdom cannot click (documented in the existing test file). The sheet behavior is covered via the pure `resolveQrValue` logic; please confirm on-device that tapping a row copies + switches the QR, and that a selected address no longer reverts after an LNURL reconnect.

## Summary by CodeRabbit

* **Bug Fixes**
  * BIP21 URI parameters now parse case-insensitively for improved compatibility
  * LNURL display corrected for amountless payment sessions
  * QR code address selection now persists correctly across rebuilds
  * Amount formatting improved for consistency in BIP21 URIs

### DIFF
--- a/src/lib/bip21.ts
+++ b/src/lib/bip21.ts
@@ -45,30 +45,35 @@ export const decodeBip21 = (uri: string): Bip21Decoded => {
   if (queryString) {
     const params = new URLSearchParams(queryString)
 
-    if (params.has('ark') || params.has('ARK')) {
-      const arkAddress = params.get('ark') ?? params.get('ARK')!
-      if (isArkAddress(arkAddress)) result.arkAddress = arkAddress
+    // BIP21 query keys are case-insensitive — match them regardless of case
+    // (e.g. 'ark', 'ARK', 'Ark'), so QR codes and mixed-case URIs all parse.
+    const getParam = (name: string): string | null => {
+      for (const [key, value] of params) {
+        if (key.toLowerCase() === name) return value
+      }
+      return null
     }
 
-    if (params.has('assetid') || params.has('ASSETID')) {
-      const param = params.get('assetid') ?? params.get('ASSETID')!
-      if (param) result.assetId = param
-    }
+    const arkAddress = getParam('ark')
+    if (arkAddress && isArkAddress(arkAddress)) result.arkAddress = arkAddress
 
-    if (params.has('amount') || params.has('AMOUNT')) {
-      const param = params.get('amount') ?? params.get('AMOUNT')!
+    const assetId = getParam('assetid')
+    if (assetId) result.assetId = assetId
+
+    const amountParam = getParam('amount')
+    if (amountParam != null) {
       if (result.assetId != null) {
-        if (!param.match(/^\d+(\.\d+)?$/)) throw new Error('Invalid asset amount')
-        result.assetAmount = param
+        if (!amountParam.match(/^\d+(\.\d+)?$/)) throw new Error('Invalid asset amount')
+        result.assetAmount = amountParam
       } else {
-        const amount = Number(param)
+        const amount = Number(amountParam)
         if (isNaN(amount) || amount < 0 || !isFinite(amount)) throw new Error('Invalid amount')
         result.satoshis = toSatoshis(Number(amount))
       }
     }
 
-    if (params.has('lightning') || params.has('LIGHTNING')) {
-      const lightning = params.get('lightning') ?? params.get('LIGHTNING')!
+    const lightning = getParam('lightning')
+    if (lightning) {
       if (lightning.toLowerCase().startsWith('lnurl')) {
         result.lnUrl = lightning
       } else if (lightning.toLowerCase().startsWith('ln')) {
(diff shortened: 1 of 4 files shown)


ghost (approved):
## Code Review — arkana

**Verdict: Approve** ✅

Clean, well-scoped bugfix. Three distinct bugs, each with a clear root cause, a minimal fix, and test coverage. No protocol-critical paths touched.

---

### bip21.ts — case-insensitive `getParam` ✅

The `getParam` helper (`src/lib/bip21.ts:51-55`) is the right fix. `URLSearchParams` preserves case, and BIP21 says nothing about case-sensitivity of extension keys. The old `has('ark') || has('ARK')` pattern missed `Ark`, `aRk`, etc. — the new linear scan over entries with `.toLowerCase()` is correct and no worse in performance (old code also did O(n) per `has()`+`get()` pair).

One note: `getParam` requires the `name` argument be lowercase. All four call sites pass string literals (`'ark'`, `'assetid'`, `'amount'`, `'lightning'`), so this is fine, but a brief JSDoc on that contract would help future callers.

The `Bip21Decoded` interface is unchanged, so all consumers (`Send/Form.tsx:262`, `InputAddress.tsx`, e2e tests) are unaffected.

### bip21.ts — `useGrouping: false` in `encodeBip21` ✅

`src/lib/bip21.ts:93`: Passing `useGrouping = false` to `prettyNumber` is correct. BIP21 amounts are parsed by `Number()` downstream — a comma would break parsing. The `prettyNumber` signature (`format.ts:209`) confirms the third positional param is `useGrouping`. Test at `bip21.test.ts:74` covers this with a 1000 BTC edge case.

### QrCode.tsx — `resolveQrValue` ✅

`src/screens/Wallet/Receive/QrCode.tsx:56-67`: Pure function, exported for testing, clean contract. The `.filter(Boolean)` + `.includes(selected)` pattern correctly invalidates stale selections (e.g. when an invoice is regenerated, the old one drops out of candidates, triggering fallback to the unified URI). This is the core fix for the "clipboard copies wrong thing" bug.

### QrCode.tsx — `displayLnurl` gating ✅

`src/screens/Wallet/Receive/QrCode.tsx:131-133`: Gating LNURL display via `displayLnurl = isAmountlessLnurl ? lnurlSession.lnurl : ''` is correct. The LNURL session stays alive underneath (no teardown), just hidden from QR and copy list when an amount is set. The `encodeBip21` call at line 173 and the `AddressList` at line 512 both use `displayLnurl` consistently.

### QrCode.tsx — `onSelect` now copies + switches ✅

The `onSelect` callback (`src/screens/Wallet/Receive/QrCode.tsx:511-514`) now does `setSelectedValue(v)` + `setQrCodeValue(v)` + `handleCopy(v)`. Since `handleCopy` already calls `setShowCopySheet(false)`, the sheet-close behavior is preserved.

The `Focusable.onEnter` at line 601 was changed to only call `onSelect(value)` (removing the separate `onCopy` call) to avoid double-copying. Correct, since `onSelect` now includes the copy.

**Minor note** (not blocking): If a keyboard Enter event triggers both `Focusable.onEnter` and the `FlexRow.onClick`, `handleCopy` fires twice (double toast). This is a pre-existing issue with the same component structure, not introduced by this PR.

### Test coverage ✅

- Mixed-case key decode: covered (`bip21.test.ts:41-53`)
- Mixed-case amount: covered (`bip21.test.ts:55-58`)
- No thousands separator: covered (`bip21.test.ts:74-78`)
- `resolveQrValue` unit tests: default, persist, stale-fallback all covered (`receive-qrcode.test.tsx:369-381`)
- LNURL present when amountless / absent with amount: covered (`receive-qrcode.test.tsx:369-405`)
- On-device copy-sheet interaction correctly noted as not CI-verifiable

### Cross-repo impact: None

`decodeBip21`/`encodeBip21` are wallet-internal (`src/lib/bip21.ts`). No SDK or external consumers.

---

LGTM. Ship it.

PR #674, opened 2026-06-15
feat(dev): support VITE_DEV_MNEMONIC for dev auto-init

Kukks:
## Summary

Extends the dev-only auto-init (previously `VITE_DEV_NSEC` only) to also accept a BIP39 mnemonic via `VITE_DEV_MNEMONIC`, so a dev wallet can be auto-initialized from a seed phrase. When both vars are set, the mnemonic takes precedence.

- **`App.tsx`** — the startup routing keeps the wallet on the loading screen, and the boot animation is skipped, while *either* dev var auto-initializes the wallet (previously only `VITE_DEV_NSEC`).
- **`providers/wallet.tsx`** — `WalletProvider` auto-init now initializes from the mnemonic (`initWallet({ mnemonic })`, which builds a `MnemonicIdentity`) or falls back to the nsec/private-key path.
- **`vite-env.d.ts`** — declares `VITE_DEV_MNEMONIC` in `ImportMetaEnv` (parity with `VITE_DEV_NSEC`).

All of this is gated on `import.meta.env.DEV`, so it has no effect on production builds.

## Test plan

- New unit test in `App.test.tsx` asserts that with `VITE_DEV_MNEMONIC` set and the wallet not yet initialized, the app holds on the loading screen instead of redirecting to unlock/init. Verified it fails without the `App.tsx` guard change (genuine regression test).
- `tsc`, `eslint`, prettier, and the existing suite pass locally.

## Notes / follow-up

- The `WalletProvider` credential-selection branch (mnemonic vs. nsec in `wallet.tsx`) is **not** covered by a unit test: `WalletProvider` currently has no test harness, and the pre-existing `VITE_DEV_NSEC` init path was untested too. The new test covers the observable App-level routing behavior. Standing up a full `WalletProvider` test harness (mocking the SDK `MnemonicIdentity`, IndexedDB adapter, boltz-swap repos, ASP context) is a larger, separate effort — happy to do it as a follow-up if wanted.
- Neither `VITE_DEV_NSEC` nor `VITE_DEV_MNEMONIC` is documented in an `.env.example` (none exists in the repo), so this PR doesn't add one to stay consistent with the current convention.

## Summary by CodeRabbit

* **New Features**
  * Wallet auto-initialization now supports both mnemonic seed phrases and nsec private keys during development, with preference for mnemonic when available.

* **Tests**
  * Added test coverage verifying the loading screen remains visible during wallet initialization in development mode.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -100,8 +100,9 @@ export default function App() {
     // avoid redirect if the user is still setting up the wallet
     if (initInfo.password || initInfo.privateKey) return
     if (!walletLoaded) return navigate(Pages.Loading)
-    // dev auto-init: stay on loading screen while VITE_DEV_NSEC initializes the wallet
-    if (import.meta.env.DEV && import.meta.env.VITE_DEV_NSEC && !initialized) return
+    // dev auto-init: stay on loading screen while VITE_DEV_NSEC/MNEMONIC initializes the wallet
+    if (import.meta.env.DEV && (import.meta.env.VITE_DEV_NSEC || import.meta.env.VITE_DEV_MNEMONIC) && !initialized)
+      return
     if (!wallet.pubkey) return navigate(Pages.Init)
     if (authState === 'locked') return navigate(Pages.Unlock)
   }, [walletLoaded, wallet.pubkey, authState, initInfo, aspInfo.unreachable, jsCapabilitiesChecked, isCapable])
@@ -178,9 +179,9 @@ export default function App() {
 
   // Boot animation: persists on Loading, then flies to the LogoIcon position when
   // Wallet is reached. For any other destination (Unlock, Init, etc.), exits with fly-up.
-  // Skip in dev with VITE_DEV_NSEC — the fast auto-init races with the animation.
+  // Skip in dev with VITE_DEV_NSEC/MNEMONIC — the fast auto-init races with the animation.
   useEffect(() => {
-    if (import.meta.env.DEV && import.meta.env.VITE_DEV_NSEC) return
+    if (import.meta.env.DEV && (import.meta.env.VITE_DEV_NSEC || import.meta.env.VITE_DEV_MNEMONIC)) return
 
     if (page === Pages.Loading && !bootAnimActive) {
       setBootAnimDone(false)
--- a/src/providers/wallet.tsx
+++ b/src/providers/wallet.tsx
@@ -197,20 +197,21 @@ export const WalletProvider = ({ children }: { children: ReactNode }) => {
 
   // wallet is read synchronously in useState initializer above
 
+  const devMnemonic = import.meta.env.VITE_DEV_MNEMONIC as string | undefined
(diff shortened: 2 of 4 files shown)


ghost (approved):
## Review: APPROVE ✅

**Scope:** Dev-only auto-init extension — adds `VITE_DEV_MNEMONIC` alongside existing `VITE_DEV_NSEC`. 4 files touched, all changes are mechanical extensions of the existing pattern.

### Security

- **Dev-gating is correct.** All three call sites (`App.tsx:104`, `App.tsx:184`, `wallet.tsx:202`) are behind `import.meta.env.DEV`. Vite tree-shakes these in production builds — the mnemonic string never ships to users.
- **No new key exposure surface.** The mnemonic flows through the same `initWallet({ mnemonic })` path that already exists for the onboarding flow (`wallet.tsx:615-621`). `MnemonicIdentity.fromMnemonic()` is already imported and used in production code.
- **Mnemonic precedence is clean.** `wallet.tsx:213-214`: `devMnemonic` is checked first, `devNsec` second. No ambiguity when both are set (matches PR description).

### Correctness

- **`App.tsx:104-105`** — Routing guard correctly extended with `||`. The `!initialized` check prevents the guard from firing after init completes. Same pattern as the pre-existing `VITE_DEV_NSEC` guard.
- **`App.tsx:184`** — Boot animation skip correctly extended. No behavioral change for production.
- **`wallet.tsx:206-224`** — The `useEffect` deps `[aspInfo.url, initialized, devAutoInitFailed]` are unchanged and correct. `devMnemonic`/`devNsec` are module-level constants (env vars don't change at runtime), so omitting them from deps is fine.
- **`wallet.tsx:207`** — Removing the `!devNsec` early-return and replacing with `!isDevAutoInit` is correct: `isDevAutoInit` already encodes the "has at least one credential" check.

### Test coverage

- New test (`App.test.tsx:133-141`) covers the observable behavior: with `VITE_DEV_MNEMONIC` set, the app holds on loading instead of redirecting to Unlock/Init. This mirrors the existing NSEC test pattern.
- The test uses a truncated mnemonic (`abandon abandon abandon abandon abandon about` — 6 words, not a valid BIP39 mnemonic), which is fine because the test only exercises `App.tsx` routing logic, not `WalletProvider`'s `initWallet` call (the context is mocked).
- **No `WalletProvider` integration test** for the mnemonic path — acknowledged in the PR description as a separate effort. The pre-existing `VITE_DEV_NSEC` path was equally untested, so this doesn't regress.

### Cross-repo impact

- None. No public API, type, or interface changes. `VITE_DEV_MNEMONIC` is a local env var consumed only by the wallet app. `index.tsx` already handles it (pre-existing code).

### Not protocol-critical

This PR does not touch VTXOs, transaction signing, forfeit paths, round lifecycle, connector trees, or exit paths. The `initWallet` function it calls is unchanged. No human sign-off escalation needed.

Clean, well-scoped change. Ship it.

PR #678, opened 2026-06-17
Remove onboarding button

bordalix:


ghost (approved):
## Code Review — arkade-os/wallet#678

**Verdict: Approve** ✅

Clean PR. Mostly housekeeping + dead code removal + a deliberate UX change to remove the manual boarding settlement button from the Transaction screen.

### Changes reviewed

**1. Storage key centralization (`src/lib/storageKeys.ts`)** — Good.  
Two constants extracted from inline strings. All consumers (`mnemonic.ts`, `privateKey.ts`, tests) updated. No stale hardcoded strings remain in the codebase.

**2. Mutual exclusion of mnemonic / private key** — Correct.  
- `setMnemonic()` now removes `NSEC_STORAGE_KEY` after storing the mnemonic.  
- `setPrivateKey()` now removes `MNEMONIC_STORAGE_KEY` after storing the key.  
- Ordering is right in both: remove the old key only after the new key is persisted, so a failure mid-write doesn't leave the user with zero keys.  
- `isValidPassword()` still checks mnemonic-first then private-key — works correctly since only one should exist at a time.  
- Tests cover the new mutual-exclusion behavior in both directions. ✓

**3. SVG attribute fixes (`Copy.tsx`, `ExternalLink.tsx`)** — Correct.  
`stroke-width` → `strokeWidth`, `stroke-linecap` → `strokeLinecap`, etc. Standard React JSX camelCase for SVG props. Fixes console warnings.

**4. Transaction.tsx — behavioral changes** — This is the substantive part.

- **Removed dead code**: `handleResend` was a no-op stub (`disabled={resending || true}` — always disabled). The TODO comment and the "Resend" button for expired boarding txs are gone. No functionality lost.
- **Removed `resending` state**: Follows from above.
- **`defaultButtonLabel`**: Changed from `boardingTx ? 'Complete boarding' : 'Settle transaction'` → always `'Settle transaction'`. UX-only.
- **Status computation refactored**: Extracted to a `const status` before the `details` object. Logic is equivalent to the old inline version except for the reordering of `settleSuccess` vs `expiredBoardingTx` — but since `expiredBoardingTx` requires `!tx.settled`, and expired txs can't be settled (old `showSettleButtons` excluded them), this reordering is safe.
- **`showSettleButtons` simplified**: Uses `status === 'Preconfirmed'` instead of explicit exclusion conditions. This is equivalent for all cases *except* boarding txs with `preconfirmed=true` (status `'Pending boarding'`), where the old code **would** show the settle button and the new code **won't**. This is the intended removal of the manual "Complete boarding" flow.

> **Note for author**: `src/screens/Settings/Vtxos.tsx:67-71` still references "Complete boarding" labels — intentional to keep that path, or should it be updated in a follow-up?

**5. Package bumps** — SDK 0.4.35→0.4.36, boltz-swap 0.3.40→0.3.41. Lockfile is consistent.

**6. Test coverage** — New tests cover:
- `setMnemonic` removes private key (mnemonic.test.ts)
- `setPrivateKey` / `getPrivateKey` round-trip, wrong password, missing key, and removes mnemonic (privatekey.test.ts)

No protocol-critical concerns (no VTXO handling, signing, forfeit, or exit path changes). UI-only changes to the transaction screen.

LGTM.

ghost (approved):
## Code Review — arkade-os/wallet#678

**Verdict: Approve** ✅

Clean PR. Mostly housekeeping + dead code removal + a deliberate UX change to remove the manual boarding settlement button from the Transaction screen.

### Changes reviewed

**1. Storage key centralization (`src/lib/storageKeys.ts`)** — Good.  
Two constants extracted from inline strings. All consumers (`mnemonic.ts`, `privateKey.ts`, tests) updated. No stale hardcoded strings remain in the codebase.

**2. Mutual exclusion of mnemonic / private key** — Correct.  
- `setMnemonic()` now removes `NSEC_STORAGE_KEY` after storing the mnemonic.  
- `setPrivateKey()` now removes `MNEMONIC_STORAGE_KEY` after storing the key.  
- Ordering is right in both: remove the old key only after the new key is persisted, so a failure mid-write doesn't leave the user with zero keys.  
- `isValidPassword()` checks mnemonic-first then private-key — works correctly since only one should exist at a time.  
- Tests cover the new mutual-exclusion behavior in both directions. ✓

**3. SVG attribute fixes (`Copy.tsx`, `ExternalLink.tsx`)** — Correct.  
`stroke-width` → `strokeWidth`, etc. Standard React JSX camelCase. Fixes console warnings.

**4. Transaction.tsx — behavioral changes** — The substantive part.

- **Removed dead code**: `handleResend` was a no-op stub (`disabled={resending || true}` — always disabled). The TODO comment and "Resend" button for expired boarding txs are gone. No functionality lost.
- **`defaultButtonLabel`**: Always `'Settle transaction'` now. UX-only.
- **Status computation refactored**: Extracted to `const status`. Reordering of `settleSuccess` vs `expiredBoardingTx` is safe because `expiredBoardingTx` requires `!tx.settled` and expired txs couldn't be settled anyway.
- **`showSettleButtons` simplified via `status === 'Preconfirmed'`**: Equivalent for all cases *except* boarding txs with `preconfirmed=true` (status `'Pending boarding'`), where the old code would show the settle button and the new code won't. This is the intended removal of the manual "Complete boarding" flow.

> **Note**: `src/screens/Settings/Vtxos.tsx:67-71` still references "Complete boarding" labels — intentional to keep that path, or should it be updated in a follow-up?

**5. Package bumps** — SDK 0.4.35→0.4.36, boltz-swap 0.3.40→0.3.41. Lockfile consistent.

**6. Test coverage** — New tests cover mutual exclusion in both directions. ✓

No protocol-critical concerns — no VTXO handling, signing, forfeit, or exit path changes. UI-only changes to the transaction screen.

LGTM.

PR #681, opened 2026-06-17
Clean qrcode

bordalix:


ghost (approved):
## Review: Clean qrcode

**Verdict: Approve** — UI cleanup, no protocol-critical code.

### Summary
Refactors the QR code receive screen to use existing `FlexCol`/`Text` components instead of raw `div`s with inline styles. Also tweaks `WarningBox` to be max-width constrained and centered, and removes a stray blank line in `swaps.tsx`.

### Findings

**Minor — `WarningBox` style change is global** (`src/components/Warning.tsx:19-20`)
Changing from `width: 100%` to `maxWidth: 400px` + `margin: 0 auto` affects all 8 consumers (Settings/Server, Settings/Backup, Settings/Vtxos, Settings/Delegates, Boltz/Settings, QrCode). In narrow containers (<400px) this is a no-op; in wider containers the box will now be centered and capped at 400px instead of stretching full width. This is likely the intended behavior everywhere, but worth a quick visual check on the Settings screens to confirm nothing looks awkward.

**Nit — hardcoded `marginTop: '5rem'`** (`src/screens/Wallet/Receive/QrCode.tsx:~434`)
The old code used flex centering (`flex: 1` + `alignItems: center` + `justifyContent: center`) to vertically position the QR code. The new code replaces that with a fixed `marginTop: '5rem'` on the button. This works but is fragile on very short viewports — the QR code could overflow or get cut off. The old approach was more robust. Not a blocker, just something to watch if users report clipping on small screens.

**OK — swaps.tsx blank line removal** (`src/providers/swaps.tsx:110`)
Purely cosmetic, no logic change.

**OK — `<div>` → `<p>` for "No valid payment methods"** (`QrCode.tsx:417`)
Semantically correct, no behavioral change.

### Cross-repo impact
None. All changes are internal UI components with no public API surface.

### Protocol-critical
No. None of these files touch VTXOs, signing, forfeit paths, round lifecycle, or exit paths.

LGTM ✅

PR #683, opened 2026-06-17
clean Reminder

bordalix:


ghost (commented):
## Code Review — `clean Reminder`

**Verdict: Looks good overall** — clean refactor replacing inline styles and custom CSS with shared `Button` and `Text` components. One minor issue to flag.

### 🟡 Minor: Subtitle color lost
**`src/components/Reminder.tsx`** — The old code applied `color: 'var(--neutral-500)'` to the date subtitle text. The new `<Text small>` (line ~52 in the new code) doesn't pass a `color` prop, so `prettyDate(startTime)` will render in the default text color instead of the muted grey. Likely should be:
```tsx
<Text small color='neutral-500'>{prettyDate(startTime)}</Text>
```

### ℹ️ Note: SheetModal padding is global
**`src/components/SheetModal.tsx:21`** — Adding `pt-4` to the drawer content wrapper affects **all** SheetModal consumers (Init.tsx, QrCode.tsx ×2, Send/Form.tsx), not just Reminder. Those screens already have their own inner padding so it's probably fine, but worth a quick visual check on each.

### ✅ Everything else
- `.reminder-button` CSS removal is safe — only used in `Reminder.tsx`
- `Button outline` and `Text bold` correctly map to the design system variants
- No protocol-critical code touched

LGTM pending the color fix above. 🤙

---
*Reviewed by Arkana 🔍*

ghost (commented):
## Code Review — `clean Reminder`

**Verdict: Looks good overall** — clean refactor replacing inline styles and custom CSS with shared `Button` and `Text` components. One minor issue to flag.

### 🟡 Minor: Subtitle color lost
**`src/components/Reminder.tsx`** — The old code applied `color: 'var(--neutral-500)'` to the date subtitle text. The new `<Text small>` (line ~52 in the new code) doesn't pass a `color` prop, so `prettyDate(startTime)` will render in the default text color instead of the muted grey. Likely should be:
```tsx
<Text small color='neutral-500'>{prettyDate(startTime)}</Text>
```

### ℹ️ Note: SheetModal padding is global
**`src/components/SheetModal.tsx:21`** — Adding `pt-4` to the drawer content wrapper affects **all** SheetModal consumers (Init.tsx, QrCode.tsx ×2, Send/Form.tsx), not just Reminder. Those screens already have their own inner padding so it's probably fine, but worth a quick visual check on each.

### ✅ Everything else
- `.reminder-button` CSS removal is safe — only used in `Reminder.tsx`
- `Button outline` and `Text bold` correctly map to the design system variants
- No protocol-critical code touched

LGTM pending the color fix above. 🤙

---
*Reviewed by Arkana 🔍*

ghost (approved):
## Re-review — new commit `f2d7fba`

Color fix looks good — `<Text small color='neutral-500'>` correctly restores the muted subtitle style from the old inline CSS. Previous feedback addressed. ✅

No further issues. LGTM.

---
*Reviewed by Arkana 🔍*

PR #685, opened 2026-06-18
fix: serve .mjs as JS so the service worker registers (+ runtime-configurable LNURL URL)

Kukks:
## Problem

Running the published `ghcr.io/arkade-os/wallet:latest` image locally (e.g. against a regtest Ark stack), the wallet never advances past the welcome screen. Clicking **Create wallet** logs:

> Error during connection: Failed to register a ServiceWorker for scope ('http://localhost/') with script ('http://localhost/wallet-service-worker.mjs'): The script has an unsupported MIME type ('application/octet-stream').

### Root cause

nginx's bundled `mime.types` has no mapping for `.mjs`, so `wallet-service-worker.mjs` is served with the default `application/octet-stream`. Browsers refuse to register a service worker (or load an ES-module script) with that MIME type, so the app's connection routine throws and the wallet can't be created. The repo's `nginx.conf` cache `location` regex (`\.(?:css|js|…|map)$`) also omits `mjs` and adds no `types` mapping, so nothing corrects it.

## Fix

1. **`.mjs` MIME type** — add an nginx `location ~ \.mjs$` that forces `application/javascript`. It uses `Cache-Control: no-cache` rather than the `immutable` 1y policy applied to hashed assets, because the service-worker filename isn't content-hashed and must be allowed to update.

2. **Runtime-configurable LNURL server URL** — `lnurlServerUrl` was read from `import.meta.env.VITE_LNURL_SERVER_URL` at build time only, with no build arg, placeholder, or entrypoint substitution — so it was always `undefined` in the published image and couldn't be set per-deployment. Wire it through the same `ARG`/`ENV` + `__VITE_LNURL_SERVER_URL__` placeholder + `docker-entrypoint.sh` substitution path as `VITE_ARK_SERVER`, `VITE_BOLTZ_URL`, etc.

   Because `lnurlServerUrl` is consumed by truthiness (`!!lnurlServerUrl`, `rawLnurlServerUrl?.replace(...)`) with no fallback, the unsubstituted placeholder is explicitly treated as "unset" — otherwise a deployment that doesn't set `VITE_LNURL_SERVER_URL` would enable the LNURL flow with the literal `__VITE_LNURL_SERVER_URL__` string as its URL.

## Verification

Built the image from this branch and ran it:

- `.mjs` is now served as `Content-Type: application/javascript` with `Cache-Control: no-cache`; `navigator.serviceWorker.register('/wallet-service-worker.mjs', { type: 'module' })` succeeds.
- Default (no `VITE_LNURL_SERVER_URL`): the placeholder remains in the bundle and the `startsWith('__VITE_')` guard resolves `lnurlServerUrl` to `undefined` — LNURL stays disabled, matching prior behavior.
- With `-e VITE_LNURL_SERVER_URL=http://localhost:9090`: the placeholder is fully substituted (zero literal placeholders remain in the served bundle).

## Summary by CodeRabbit

* **New Features**
  * Added support for a configurable LNURL server URL via `VITE_LNURL_SERVER_URL`, supporting both build-time and runtime configuration.
* **Improvements**
  * Enhanced runtime placeholder substitution so only real `VITE_*` values are applied, preventing unresolved placeholders from leaking into the app.
  * Updated static asset serving so `.mjs` files use the correct JavaScript MIME type and are not cached, while other assets keep immutable caching.
* **Bug Fixes**
  * Refined Boltz base URL resolution to correctly respect placeholder-aware runtime values.

3. **Generic entrypoint** — `docker-entrypoint.sh` now substitutes placeholders by looping over the `VITE_*` environment at startup instead of a hardcoded per-var list, so adding a runtime-configurable var only requires its Dockerfile `ARG` (prevents the entrypoint/Dockerfile drift that caused the LNURL gap).


ghost (approved):
## Review: ✅ Approve

Clean, well-documented fix. Not protocol-critical (no VTXO/signing/forfeit/exit path changes).

### nginx `.mjs` MIME fix
**Correct.** The `location ~ \.mjs$` block is properly placed before the static assets regex. Since nginx uses first-matching-regex semantics, `.mjs` files will hit this block and get `application/javascript` + `no-cache`. The existing `\.js$` pattern in the assets block does *not* match `.mjs`, so there's no conflict even without ordering. `no-cache` is the right choice — the service worker filename isn't content-hashed, so it must be revalidated.

### Runtime-configurable LNURL URL
**Correct.** Follows the established `ARG`→`ENV`→placeholder→entrypoint-substitution pattern used by all other `VITE_*` vars. The `__VITE_` prefix guard in `constants.ts:17-18` is a good defensive addition — without it, a deployment that omits `VITE_LNURL_SERVER_URL` would pass a truthy `"__VITE_LNURL_SERVER_URL__"` string into the LNURL provider, triggering SSE connections to a bogus URL.

### Verified
- `lnurl.tsx` imports `lnurlServerUrl` (already sanitized) and aliases it as `rawLnurlServerUrl` — the `.replace(/\/+$/, '')` trailing-slash trim works correctly on `string | undefined`.
- `docker-entrypoint.sh` `substitute()` uses `find "$JS_DIR" -name '*.js'` — the LNURL placeholder lives in `constants.ts` which bundles into `.js` chunks in `assets/`, not the `.mjs` service worker, so substitution works.
- The service worker (`vite.worker.config.ts`) builds separately into `public/` and doesn't import `constants.ts`, confirming no placeholder leakage into `.mjs`.
- Type declaration in `vite-env.d.ts` correctly added as optional.

### Minor note (not blocking)
The `__VITE_` prefix guard is only applied to `lnurlServerUrl`. Other env vars (e.g., `defaultArkServer`) don't have this guard. If a non-Docker deployment forgets to set `VITE_ARK_SERVER`, the literal placeholder `__VITE_ARK_SERVER__` would be used as a server URL. Consider applying this pattern uniformly in a follow-up.

LGTM 🚀

_Reviewed by Arkana_

ghost (approved):
## Review: ✅ Approve

Clean, well-documented fix. Not protocol-critical (no VTXO/signing/forfeit/exit path changes).

### nginx `.mjs` MIME fix
**Correct.** The `location ~ \.mjs$` block is properly placed before the static assets regex. Since nginx uses first-matching-regex semantics, `.mjs` files will hit this block and get `application/javascript` + `no-cache`. The existing `\.js$` pattern in the assets block does *not* match `.mjs`, so there's no conflict even without ordering. `no-cache` is the right choice — the service worker filename isn't content-hashed, so it must be revalidated.

### Runtime-configurable LNURL URL
**Correct.** Follows the established `ARG`→`ENV`→placeholder→entrypoint-substitution pattern used by all other `VITE_*` vars. The `__VITE_` prefix guard in `constants.ts:17-18` is a good defensive addition — without it, a deployment that omits `VITE_LNURL_SERVER_URL` would pass a truthy `"__VITE_LNURL_SERVER_URL__"` string into the LNURL provider, triggering SSE connections to a bogus URL.

### Verified
- `lnurl.tsx` imports `lnurlServerUrl` (already sanitized) and aliases it as `rawLnurlServerUrl` — the `.replace(/\/+$/, '')` trailing-slash trim works correctly on `string | undefined`.
- `docker-entrypoint.sh` `substitute()` uses `find "$JS_DIR" -name '*.js'` — the LNURL placeholder lives in `constants.ts` which bundles into `.js` chunks in `assets/`, not the `.mjs` service worker, so substitution works.
- The service worker (`vite.worker.config.ts`) builds separately into `public/` and doesn't import `constants.ts`, confirming no placeholder leakage into `.mjs`.
- Type declaration in `vite-env.d.ts` correctly added as optional.

### Minor note (not blocking)
The `__VITE_` prefix guard is only applied to `lnurlServerUrl`. Other env vars (e.g., `defaultArkServer`) don't have this guard. If a non-Docker deployment forgets to set `VITE_ARK_SERVER`, the literal placeholder `__VITE_ARK_SERVER__` would be used as a server URL. Consider applying this pattern uniformly in a follow-up.

LGTM 🚀

_Reviewed by Arkana_

bordalix:
Should we add the same guard (`constants.ts:17-18`) to VITE_ARK_SERVER and VITE_BOLTZ_URL?

Kukks:
Done in 0999b88f. Extracted the guard into a shared `fromRuntimeEnv()` helper and applied it to both:

- **`VITE_ARK_SERVER`** in `defaultArkServer()` — an unsubstituted placeholder now falls through to the domain-based defaults instead of being returned as the server URL.
- **`VITE_BOLTZ_URL`** in `swaps.tsx` `BASE_URLS.bitcoin` — falls back to `null` (boltz disabled for mainnet, same as when unset) instead of the literal placeholder.

Verified against a build from this branch: with the vars set, both placeholders are fully substituted (`localhost:7070` / `localhost:9069` land in the bundle); with them unset, the placeholders remain and the compiled guard resolves them to `undefined`.

Scope note: the other `VITE_ARK_SERVER` reads in `config.tsx` are gated behind `import.meta.env.DEV`, which is `false` in the `vite build` Docker image, so they never see a baked placeholder — left as-is. The remaining optional `VITE_*` vars (sentry / chatwoot / nostr / etc.) can reuse `fromRuntimeEnv()` if you'd like them guarded too — happy to extend here or as a follow-up.

ghost (approved):
## Iterative Review: new commit `0999b88` ✅ Approve

Good follow-up. Extracting `fromRuntimeEnv` is the right call — single guard function, reused everywhere a Docker placeholder could leak through as a truthy string.

### What changed
- `fromRuntimeEnv(value)` replaces the inline `startsWith('__VITE_')` guard → cleaner, DRY.
- Applied to `VITE_ARK_SERVER` in `defaultArkServer()` and `VITE_BOLTZ_URL` in `swaps.tsx` — both were previously unguarded against unsubstituted placeholders.
- `swaps.tsx` imports `fromRuntimeEnv` from constants.

### Verified correct
- **`fromRuntimeEnv` logic** (`constants.ts:20-21`): Short-circuits on falsy, rejects `__VITE_*` prefixes, returns real values. Correct.
- **`defaultArkServer`** (`constants.ts:25-26`): Falls through to domain-based detection when placeholder is unsubstituted. Correct — same behavior as before for Docker images without `VITE_ARK_SERVER` set.
- **`VITE_BOLTZ_URL`** (`swaps.tsx:30`): `fromRuntimeEnv(...) ?? null` — unsubstituted placeholder → `undefined` → `null`. Correct — Bitcoin mainnet Boltz stays disabled unless explicitly configured.

### Minor note (non-blocking)
Other Docker-substituted vars (`VITE_NOSTR_RELAY_URL` in `nostr.ts:8`, `VITE_VERIFIED_ASSETS_URL` in `wallet.tsx:291`, `VITE_SENTRY_DSN` in `index.tsx:27`) have the same unsubstituted-placeholder exposure but are outside this PR's scope. Worth a follow-up but not a blocker — none are protocol-critical and most fail gracefully (bad relay URL → connection error, bad Sentry DSN → no telemetry, bad verified-assets URL → fetch 404).

Not protocol-critical. No VTXO/signing/forfeit/exit path changes.

_Reviewed by Arkana_

ghost (approved):
## Iterative Review: new commit `0999b88` ✅ Approve

Good follow-up. Extracting `fromRuntimeEnv` is the right call — single guard function, reused everywhere a Docker placeholder could leak through as a truthy string.

### What changed
- `fromRuntimeEnv(value)` replaces the inline `startsWith('__VITE_')` guard → cleaner, DRY.
- Applied to `VITE_ARK_SERVER` in `defaultArkServer()` and `VITE_BOLTZ_URL` in `swaps.tsx` — both were previously unguarded against unsubstituted placeholders.
- `swaps.tsx` imports `fromRuntimeEnv` from constants.

### Verified correct
- **`fromRuntimeEnv` logic** (`constants.ts:20-21`): Short-circuits on falsy, rejects `__VITE_*` prefixes, returns real values. Correct.
- **`defaultArkServer`** (`constants.ts:25-26`): Falls through to domain-based detection when placeholder is unsubstituted. Correct — same behavior as before for Docker images without `VITE_ARK_SERVER` set.
- **`VITE_BOLTZ_URL`** (`swaps.tsx:30`): `fromRuntimeEnv(...) ?? null` — unsubstituted placeholder → `undefined` → `null`. Correct — Bitcoin mainnet Boltz stays disabled unless explicitly configured.

### Minor note (non-blocking)
Other Docker-substituted vars (`VITE_NOSTR_RELAY_URL` in `nostr.ts:8`, `VITE_VERIFIED_ASSETS_URL` in `wallet.tsx:291`, `VITE_SENTRY_DSN` in `index.tsx:27`) have the same unsubstituted-placeholder exposure but are outside this PR's scope. Worth a follow-up but not a blocker — none are protocol-critical and most fail gracefully (bad relay URL → connection error, bad Sentry DSN → no telemetry, bad verified-assets URL → fetch 404).

Not protocol-critical. No VTXO/signing/forfeit/exit path changes.

_Reviewed by Arkana_

Kukks:
Made the entrypoint generic in 575b4a06: replaced the hardcoded per-var `substitute` list with a loop over the `VITE_*` env vars present at runtime. A new runtime-configurable var now only needs its `ARG` in the Dockerfile — no parallel line in the entrypoint, which removes the exact drift that left `VITE_LNURL_SERVER_URL` unsubstituted to begin with.

Verified against a build: setting `VITE_ARK_SERVER`/`VITE_BOLTZ_URL`/`VITE_LNURL_SERVER_URL`/`VITE_NOSTR_RELAY_URL` substitutes all of them (0 placeholders left); unset vars keep their placeholder for the `fromRuntimeEnv` guard to treat as unset. Empty-skip and sed-escaping behavior unchanged.

ghost (approved):
## Iterative Review: new commit `575b4a06` ✅ Approve

Clean refactor. The hardcoded `substitute` call list is replaced with a dynamic `VITE_*` env-var loop — eliminates the exact category of drift that caused the original bug (`VITE_LNURL_SERVER_URL` had an ARG but no entrypoint line).

### What changed
- `docker-entrypoint.sh`: Removed `substitute()` function and 11 explicit calls. Replaced with a `for` loop over `printenv | sed -n 's/^\(VITE_[A-Za-z0-9_]*\)=.*/\1/p'`.

### Verified correct
- **Name extraction regex** `[A-Za-z0-9_]*` — only alphanumeric + underscore can appear in `$name`, so `s|__${name}__|...|g` cannot inject sed metacharacters. Safe.
- **Value escaping** `sed 's/[&|\\]/\\&/g'` — escapes the three characters special in a sed replacement with `|` delimiter (`&`, `|`, `\`). Unchanged from the old `substitute()`. Correct.
- **Empty-skip** `[ -n "$value" ] || continue` — preserves behavior: unset/empty vars leave `__VITE_*__` placeholders for `fromRuntimeEnv` to reject. Correct.
- **Scope expansion**: Any `VITE_*` env var now gets substituted, not just the hardcoded 10. This is intentional and the right tradeoff — adding a new var only requires a Dockerfile ARG. The security surface is effectively unchanged: an attacker who can inject container env vars already has container-level access.
- **No `.mjs` files processed**: `find "$JS_DIR" -name '*.js'` — the glob doesn't match `.mjs`, so the service worker is untouched by substitution. Correct (it contains no placeholders anyway).

### Edge case (non-blocking)
If a `VITE_*` value contains a newline, `sed` will choke. Extremely unlikely for URL/token/DSN values and was equally broken in the old code, so not a regression.

Not protocol-critical. No VTXO/signing/forfeit/exit path changes.

_Reviewed by Arkana_

ghost (approved):
## Iterative Review: new commit `575b4a06` ✅ Approve

Clean refactor. The hardcoded `substitute` call list is replaced with a dynamic `VITE_*` env-var loop — eliminates the exact category of drift that caused the original bug (`VITE_LNURL_SERVER_URL` had an ARG but no entrypoint line).

### What changed
- `docker-entrypoint.sh`: Removed `substitute()` function and 11 explicit calls. Replaced with a `for` loop over `printenv | sed -n 's/^\(VITE_[A-Za-z0-9_]*\)=.*/\1/p'`.

### Verified correct
- **Name extraction regex** `[A-Za-z0-9_]*` — only alphanumeric + underscore can appear in `$name`, so `s|__${name}__|...|g` cannot inject sed metacharacters. Safe.
- **Value escaping** `sed 's/[&|\\]/\\&/g'` — escapes the three characters special in a sed replacement with `|` delimiter (`&`, `|`, `\`). Unchanged from the old `substitute()`. Correct.
- **Empty-skip** `[ -n "$value" ] || continue` — preserves behavior: unset/empty vars leave `__VITE_*__` placeholders for `fromRuntimeEnv` to reject. Correct.
- **Scope expansion**: Any `VITE_*` env var now gets substituted, not just the hardcoded 10. This is intentional and the right tradeoff — adding a new var only requires a Dockerfile ARG. The security surface is effectively unchanged: an attacker who can inject container env vars already has container-level access.
- **No `.mjs` files processed**: `find "$JS_DIR" -name '*.js'` — the glob doesn't match `.mjs`, so the service worker is untouched by substitution. Correct (it contains no placeholders anyway).

### Edge case (non-blocking)
If a `VITE_*` value contains a newline, `sed` will choke. Extremely unlikely for URL/token/DSN values and was equally broken in the old code, so not a regression.

Not protocol-critical. No VTXO/signing/forfeit/exit path changes.

_Reviewed by Arkana_

PR #687, opened 2026-06-18
Debounce send validation

bordalix:
Replaces #666 ✝️ 

Reverts all changes to make message error inline

@sahilc0 let's target all design changes to branch wt-wallet-redesign-followups-20260520 or else it will be a pain to merge

## Summary by CodeRabbit

* **New Features**
  * Address input now supports focus, blur, and paste events to improve recipient handling responsiveness.

* **Bug Fixes**
  * Improved recipient validation timing: errors are suppressed while typing and shown after blur, and during paste/scan/Enter-driven updates.
  * Recipient parsing was refined to preserve satoshi context across common recipient formats.
  * Lightning query validation is now stricter by properly validating LNURL values.

* **Tests**
  * Expanded coverage for LNURL edge cases and lightning query validation, plus new debounced vs immediate paste/scan validation tests.


ghost (approved):
## ✅ Approved — solid UX fix, no protocol-level concerns

Clean implementation of debounced recipient validation. The core idea is sound: defer `parseRecipient` while the user is typing (400ms), fire immediately on paste/scan/blur/Enter via `parseTick` + `immediateParseRef`. Suppressing "Invalid recipient address" while the field has focus is the right call. Good test coverage for the timing behavior.

**Not protocol-critical** — no changes to VTXO handling, signing, forfeit paths, round lifecycle, or exit paths. The send flow (`handleContinue`) is untouched.

---

### Findings (minor, non-blocking)

**1. `aspErrorText` dropped from `smartSetError` — Form.tsx:127**

Old code used `aspErrorText(aspInfo, 'Arkade server unreachable')` which shows the "wallet outdated" message when `aspInfo.outdated` is true. New code hardcodes `'Ark server unreachable'`, losing that distinction. When `aspInfo.unreachable && aspInfo.outdated`, users now get a generic "server unreachable" instead of the actionable "your wallet is outdated" message. Probably unintentional — consider restoring `aspErrorText(aspInfo, 'Ark server unreachable')`.

**2. `||` vs `??` on bip21 satoshis fallback — Form.tsx:271**

```ts
satoshis: satoshis || satoshisRef.current,
```

If `decodeBip21` returns `satoshis: 0` (unusual but valid bip21 with `amount=0`), `||` falls through to `satoshisRef.current`. Should be `??` if you want to treat explicit zero as intentional. Low risk since `amount=0` bip21s are rare, but `??` is semantically correct.

**3. `nativePasteRef` setTimeout safety — InputWithScanner.tsx:44-47**

The `setTimeout(() => (nativePasteRef.current = false), 0)` cleanup is a reasonable safety net. Worth noting: if a change event fires *asynchronously* after paste (e.g., some mobile IME quirks), the timeout could clear the flag before `handleChange` reads it. In practice this seems fine on all target platforms, just flagging for awareness.

**4. Missing exhaustive deps (pre-existing, not introduced by this PR)**

The `parseRecipient` useEffect references `sendInfo.assets`, `svcWallet`, `assetMetadataCache`, `assetOptions`, `setCacheEntry`, `balance` etc. without listing them in deps `[recipient, parseTick, isAssetSend]`. This is pre-existing tech debt and not introduced by this PR — just noting it since the debounce refactor touched this effect.

---

### What looks good

- `immediateParseRef` + `parseTick` pattern cleanly separates "should this be immediate?" (ref) from "trigger a re-run" (state). Refs avoid unnecessary re-renders.
- `satoshisRef` correctly lets the async `parseRecipient` read the latest amount without stale closures.
- `smartSetError(err: unknown)` properly extracting `.message` from Error objects — fixes the old bug where `.catch(smartSetError)` would stringify the whole Error.
- `isLnAddress` TLD comment + new test cases for edge cases (single-char TLD, partial addresses).
- Test helper `renderSendForm()` reduces duplication nicely.

LGTM. Ship it. 🚢

ghost (commented):
## 🔄 Re-review after new commits

Only change since my last review (`b922587`) is merge commit `7f0c72f4` pulling `master` into the branch.

The merge brings in `f85ff5c1` ("fix branta errors #675") which replaces removed `BrantaServerBaseUrl`/`PrivacyMode` enum imports with string literals in `Form.tsx`. This is a clean merge — no conflict resolution, no changes to the debounce/validation logic.

**Original approval stands.** No new concerns.

ghost (commented):
## 🔄 Re-review — new commit `144a5bff` ("more robust isURLWithLightningQueryString")

Good change overall — validating the `lightning` query param value instead of just checking for key existence closes a real gap (arbitrary garbage in the param would previously pass validation).

### One issue to flag

**`address.ts:59` — lowercasing the full URL creates a case-sensitivity mismatch with the `Form.tsx` consumer**

The function now lowercases the entire input (`data.toLowerCase()`) before extracting query params. This means `isURLWithLightningQueryString` would return `true` for a URL like `http://example.com?LIGHTNING=lnbc1...` (uppercase key).

But in `Form.tsx:209-211`:
```ts
const url = new URL(recipient)           // original case
setRecipient(url.searchParams.get('lightning')!)  // case-sensitive lookup
```

The consumer parses the **original** URL, so `searchParams.get('lightning')` returns `null` when the key is uppercase → `setRecipient(null!)`.

**Risk: low** — payment processors always emit lowercase `lightning`. And `InputAddress.tsx:44` already passes pre-lowered data, so that call site is fine. But the inconsistency could bite if an edge-case URL surfaces.

**Suggestion:** Either (a) lowercase only the protocol portion for the scheme check, not the whole URL, or (b) also lowercase in the `Form.tsx` consumer when extracting the param. Option (a) is simpler and avoids changing any other file:

```ts
export const isURLWithLightningQueryString = (data: string): boolean => {
  try {
    if (!data.match(/^https?:\/\//i)) return false
    const url = new URL(data)
    const val = url.searchParams.get('lightning') ?? ''
    return isValidInvoice(val) || isValidLnUrl(val)
  } catch {
    return false
  }
}
```

This keeps param key matching case-sensitive (consistent with the consumer) while still being case-insensitive on the scheme. The `new URL()` constructor already handles scheme case-insensitivity natively, so the regex just needs the `/i` flag.

### Test coverage

New negative tests are good. Missing a positive test for `?lightning=<valid-lnurl>` — currently only the bolt11 invoice path is tested for the happy case.

### Verdict

Not blocking — original approval stands. The mismatch is low-risk but worth a quick fix while you're in this code.

ghost (approved):
## 🔄 Re-review — commit `f31911e5` ("fix unit tests")

Both items from my last review are addressed:

1. **Case-sensitivity fix (`address.ts:60`)** — `.toLowerCase()` on the full URL removed, `/i` flag added to the scheme regex. `new URL(data)` now parses original-case input, so `searchParams.get('lightning')` in `Form.tsx` stays consistent. ✅
2. **Missing lnurl positive test (`address.test.ts`)** — Added test iterating `fixtures.lib.lnurl` entries through `isURLWithLightningQueryString`. ✅
3. **Dead mock cleanup (`send.test.tsx`)** — Removed unused `checkLnUrlConditions` mock. ✅

Minor nit: two `it` blocks share the identical description `"should return true for a valid URL with lightning query string"` — one tests bolt11, the other lnurl. Differentiating them (e.g., `"...with bolt11 invoice"` / `"...with lnurl"`) would make test output clearer. Non-blocking.

Original approval stands. Ship it. 🚢

ghost (approved):
## 🔄 Re-review — commit `f31911e5` ("fix unit tests")

Both items from my last review are addressed:

1. **Case-sensitivity fix (`address.ts:60`)** — `.toLowerCase()` on the full URL removed, `/i` flag added to the scheme regex. `new URL(data)` now parses original-case input, so `searchParams.get('lightning')` in `Form.tsx` stays consistent. ✅
2. **Missing lnurl positive test (`address.test.ts`)** — Added test iterating `fixtures.lib.lnurl` entries through `isURLWithLightningQueryString`. ✅
3. **Dead mock cleanup (`send.test.tsx`)** — Removed unused `checkLnUrlConditions` mock. ✅

Minor nit: two `it` blocks share the identical description `"should return true for a valid URL with lightning query string"` — one tests bolt11, the other lnurl. Differentiating them (e.g., `"...with bolt11 invoice"` / `"...with lnurl"`) would make test output clearer. Non-blocking.

Original approval stands. Ship it. 🚢

src/screens/Wallet/Send/Form.tsx
```diff
@@ -105,17 +105,27 @@ export default function SendForm() {
...
   const hasAssets = assetBalances.length > 0
   const reserveApplied = !isAssetSend && hasAssets
   const liquidBalance = availableBalance - (reserveApplied ? DUST_AMOUNT : 0)
 
-  const smartSetError = (str: string) => {
-    setError(str === '' ? (aspInfo.unreachable ? aspErrorText(aspInfo, 'Arkade server unreachable') : '') : str)
+  // also receives Error objects from promise .catch handlers
+  const smartSetError = (err: unknown) => {
+    const str = err instanceof Error ? err.message : String(err ?? '')
+    setError(str === '' ? (aspInfo.unreachable ? 'Ark server unreachable' : '') : str)
```
Dunsin-cyber, line 128:
```suggestion
    setError(str === '' ? (aspInfo.unreachable ? aspErrorText(aspInfo, 'Arkade server unreachable') : '') : str)
```

this looks solid, but this PR changed the "server unreachable" message to the literal "Ark server unreachable", while the code that clears it still looks for the "Arkade" wording from aspErrorText. Once that banner shows it never goes away and the Send button stays stuck.

It is also not only the unreachable case. By hardcoding the string instead of going through aspErrorText, the outdated wallet path loses its "please update your wallet" message and hits the same stuck banner, since the clear check is looking for that outdated text too.

routing it back through the helper keeps both cases matched.

bordalix:
replaced by https://github.com/arkade-os/wallet/pull/696

PR #697, opened 2026-06-24
fix(receive): show 'Edit amount' after setting an asset amount

### LINKED ISSUE #694: Receive: amount button shows "Add amount" instead of "Edit amount" for asset receives
on the Receive screen the amount button still says "Add amount" even after you set an amount for **an asset**, because the label only checks satoshis (which stays 0 for assets) 

`hasAmount` helper in PR #693 is a simple fix for this, a  one liner, 
```ts
const amountLabel = hasAmount ? 'Edit amount' : 'Add amount`
```
@Kukks happy to open this as a follow up once #693 merges, or do you mind if we fold it straight into #693?

Dunsin-cyber:
closes #694

on the Receive screen the amount button kept reading "Add amount" even after you set an amount for an asset, because the label only looked at the Bitcoin amount (`satoshis`), which is always 0 for asset receives.

this points the label at the existing `hasAmount` helper, which checks `assetAmount` for assets and `satoshis` for Bitcoin, so it correctly shows "Edit amount" in both cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * The receive QR screen now correctly updates the “Add/Edit amount” button label when an amount is already set, including asset-based receive flows.
  * Improved consistency so the button reflects both asset amounts and satoshi amounts instead of only checking satoshi values.

### DIFF
--- a/src/screens/Wallet/Receive/QrCode.tsx
+++ b/src/screens/Wallet/Receive/QrCode.tsx
@@ -407,7 +407,7 @@ export default function ReceiveQRCode() {
     )
   }
 
-  const amountLabel = satoshis ? 'Edit amount' : 'Add amount'
+  const amountLabel = hasAmount ? 'Edit amount' : 'Add amount'
   const unitLabel = assetMeta?.metadata?.ticker ?? 'sats'
 
   return (


bordalix:
@Dunsin-cyber thank you for your contribution.

Dunsin-cyber:
> @Dunsin-cyber thank you for your contribution.

my pleasure :)

PR #700, opened 2026-06-26
always show Paste button

### LINKED ISSUE #635: ios: always ask permission for clipboard
Now somehow the system forgets permission were given, and the Paste doesnt appear in the send flow anymore

bordalix:
Closes https://github.com/arkade-os/wallet/issues/635

@tiero @sahilc0 please review

Current behaviour is:

- Button checks for data in clipboard (requires user auth at least one time)
- The button only appears If this data matches what's expected for that input

New behaviour would be:
- Kill all validations, always show the Paste button

It was a nice feature, but this iOS 16+ changes really kills it.

## Summary by CodeRabbit

* **New Features**
  * Clipboard paste now appears consistently and works with a simpler permission flow.
* **Bug Fixes**
  * Improved clipboard behavior in browsers with limited permission support, allowing paste to proceed more reliably.
  * Removed several field-specific validation checks from scanner-enabled inputs, reducing cases where valid entries were incorrectly blocked.

### DIFF
--- a/src/components/InputAddress.tsx
+++ b/src/components/InputAddress.tsx
@@ -1,13 +1,3 @@
-import { isValidLnUrl } from '../lib/lnurl'
-import {
-  isArkAddress,
-  isBTCAddress,
-  isEmailAddress,
-  isLightningInvoice,
-  isURLWithLightningQueryString,
-} from '../lib/address'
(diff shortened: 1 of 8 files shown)


tiero:
utACK

PR #712, opened 2026-07-02
fix(send): respect bitcoin unit in amount fields

sahilc0:
## Summary
- Fix the redesigned Send amount input so bitcoin balances, placeholders, secondary fiat conversions, and max-send values respect the selected bitcoin unit instead of always showing sats.
- Add shared bitcoin-unit conversion helpers for converting display values to satoshis and satoshis back to the selected display unit.
- Cover the regression with focused unit/component tests for BTC display units and typed BTC amount conversion.

## Why
PR 627's redesigned Display settings can select `Currency BTC` and `bitcoin unit BTC`, or `Currency USD` with `bitcoin unit BTC`, but the Send form still rendered sats in the amount field. This made the Send screen inconsistent with the user's display preference and could interpret BTC-denominated input incorrectly.

## Affected files
- `src/components/InputAmount.tsx`
- `src/screens/Wallet/Send/Form.tsx`
- `src/lib/format.ts`
- `src/test/screens/wallet/send.test.tsx`
- `src/test/lib/format.test.ts`

## Testing
- `bun run vitest run src/test/screens/wallet/send.test.tsx src/test/lib/format.test.ts`
- `bunx tsc -b --pretty false`
- `bunx eslint src/components/InputAmount.tsx src/screens/Wallet/Send/Form.tsx src/test/screens/wallet/send.test.tsx src/lib/format.ts src/test/lib/format.test.ts`
- Manual browser verification on the PR 627 branch with a seeded funded wallet:
  - `Currency USD` + `bitcoin unit BTC`: Send secondary amount now displays BTC instead of sats.
  - `Currency BTC` + `bitcoin unit BTC`: Send available amount and amount placeholder now display BTC instead of sats.

## Visual evidence
Manual before/after screenshots were captured locally during verification. The post-fix states show BTC units in the same flows that previously showed sats.

### DIFF
--- a/src/lib/format.ts
+++ b/src/lib/format.ts
@@ -13,6 +13,16 @@ export const toSatoshis = (num: number): number => {
   return Decimal.mul(num, 100_000_000).floor().toNumber()
 }
 
+export const bitcoinUnitToSats = (amount: number, unit: Unit): number => {
+  if (unit === Unit.BTC) return toSatoshis(amount)
+  return Math.floor(amount)
+}
+
+export const satsToBitcoinUnit = (sats: number, unit: Unit): number => {
+  if (unit === Unit.BTC) return fromSatoshis(sats)
+  return sats
+}
+
 export const prettyAgo = (timestamp: number | string, long = false): string => {
   if (!timestamp) return ''
   const now = Math.floor(Date.now() / 1000)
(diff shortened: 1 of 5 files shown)


src/lib/format.ts
```diff
@@ -13,6 +13,16 @@ export const toSatoshis = (num: number): number => {
   return Decimal.mul(num, 100_000_000).floor().toNumber()
 }
 
+export const bitcoinUnitToSats = (amount: number, unit: Unit): number => {
```
tiero, line 16: These seems redundant with toSatoshi and fromSatoshi?

The unit is simply BTC or Sats right? The B thing is just an alias to sats

tiero (changes requested):

PR #723, opened 2026-07-07
fix(wallet): smooth home action load animation

sahilc0:
## Summary
- Replaces the nested Framer Motion animation on wallet home quick action buttons with a single CSS keyframe entrance.
- Keeps the parent wallet load-in stagger, but removes the per-button Framer wrappers that caused the actions to flicker after the initial entrance.
- Adds containment to the wallet stagger wrappers so the load-in is isolated to transform/opacity work.
- Updates `bun.lock` to match the dependency versions already declared in `package.json` on current `master`; without this, the local Vite app can resolve stale package versions.

## Visual evidence
- Verified before/after on desktop and mobile-sized browser sessions.
- Phone comparison links were used during review:
  - Before timing tune: `https://vivid-bee.slim.show/?compare=before-phone`
  - After timing tune: `https://teal-crab.slim.show/?compare=after-phone`
- Both links rendered the wallet home actions during verification; Slim links are temporary 30-minute tunnels.

## Testing
- `bunx eslint src/screens/Wallet/HomeQuickActions.tsx src/components/WalletLoadIn.tsx src/lib/animations.ts`
- `git diff --check`
- `bun scripts/git-commit-info.js && bunx vite build -c vite.worker.config.ts && bunx vite build`
- Playwright mobile viewport verification against the local app and Slim phone links confirmed the wallet actions render and the computed animations are applied.

## Notes
- `bun run build` and the commit hook both attempted the package script path that shells through `pnpm`; in this local environment pnpm aborted with `ERR_PNPM_IGNORED_BUILDS` for `esbuild` and `msw`. The equivalent worker and app production builds were run directly with `bunx vite` and passed.

## Affected files
- `src/screens/Wallet/HomeQuickActions.tsx`
- `src/index.css`
- `src/components/WalletLoadIn.tsx`
- `src/lib/animations.ts`
- `bun.lock`

## Summary by CodeRabbit

* **Improvements**
  * Updated wallet and quick-action loading animations for a smoother, more consistent entrance experience.
  * Quick actions now use a simpler, static interaction layout while preserving tap feedback and disabled-state handling.
  * Reduced-motion settings are now handled more explicitly, making motion behavior easier to predict for users who prefer less animation.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,10 +5,10 @@
     "": {
       "name": "arkade",
       "dependencies": {
-        "@arkade-os/boltz-swap": "0.3.37",
-        "@arkade-os/sdk": "0.4.32",
+        "@arkade-os/boltz-swap": "0.3.47",
+        "@arkade-os/sdk": "0.4.42",
         "@base-ui/react": "^1.4.1",
-        "@branta-ops/branta": "0.0.9",
+        "@branta-ops/branta": "3.1.3",
         "@fontsource-variable/geist": "^5.2.8",
         "@lendasat/lendasat-wallet-bridge": "^0.0.90",
         "@noble/curves": "^2.0.1",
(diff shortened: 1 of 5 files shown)


sahilc0:
Addressed CodeRabbit's transform-fill comment in 15747a5 by changing the quick action entrance animation from `both` to `backwards`. Verified in Playwright that the button stays hidden during the stagger delay, animates opacity/transform once, and ends with computed `transform: none` so press feedback is no longer blocked.

PR #744, opened 2026-07-09
increase buttons-on-bottom bottom padding to 2rem

### LINKED ISSUE #725: Bottom primary actions need better iPhone/PWA safe-area handling
Primary action buttons at the bottom of mobile screens can sit too low on rounded iPhone screens / PWA installs. On real mobile web, the bottom CTA area can feel clipped by the screen/home-indicator area, which makes the app look unpolished.

I tried a shared safe-area approach in draft PR #724 using `env(safe-area-inset-bottom)` on the shared bottom action footer:

https://github.com/arkade-os/wallet/pull/724

The challenge is that this still does not feel right visually. It avoids the clipping problem in principle, but in practice there is still too much white space at the bottom on mobile, so the solution feels heavy and not polished.

@bordalix could you help think through the right approach here? I am not confident the current PR is the right solution, but it captures the area of the code and the problem we are trying to solve.

bordalix:
Closes https://github.com/arkade-os/wallet/issues/725

@sahilc0 can you please try it with your phone? Thanks.

## Summary by CodeRabbit

* **Style**
  * Updated the bottom spacing for button sections to keep controls aligned and add consistent padding at the bottom.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -237,6 +237,7 @@ iframe {
 
 .buttons-on-bottom {
   padding: 1rem;
+  padding-bottom: 2rem;
 }
 
 @media (prefers-reduced-motion: reduce) {


sahilc0:
@bordalix hmm it looks strange now, too much padding / safe area at top and bottom 
Screenshot 2026-07-13 at 11 10 17 AM

bordalix:
@sahilc0 it can't be, you must be using a different branch or something: this PR has only 1 line, and it touches only on buttons-on-bottom component.

PR #747, opened 2026-07-09
fix(wallet): preserve home card borders on first load

### LINKED ISSUE #740: borders not showing on light mode

sahilc0:
## Summary
- Preserve the wallet home initial-load state when duplicate wallet navigations happen during startup, so the home sections still use the staggered motion wrappers.
- Remove only `paint` from the wallet stagger wrapper containment so the existing outer card shadows/borders are not clipped on the first render.
- Add a regression test covering duplicate wallet navigations before React renders the wallet page.

Fixes #740

## Why
On first load, startup can dispatch `navigate(Pages.Wallet)` twice in the same tick. The second same-page navigation overwrote `previousPage` to `Wallet`, which made `isInitialLoad` false before the wallet screen mounted. That skipped the stagger wrappers entirely. Separately, when the stagger wrappers did render, `contain: paint` clipped card shadow overflow at the wrapper edge, hiding the left/right border-like shadows on first load.

## Testing
- `bun run test src/test/providers/navigation.test.tsx`
- `bun run lint`
- `bun run git-info && bun run build:worker`
- `bun ./node_modules/vite/bin/vite.js build`
- Fresh browser boot verification: home sections start staggered again at roughly 90ms intervals, and card side borders/shadows are visible on first render.

## Affected files
- `src/providers/navigation.tsx`
- `src/components/WalletLoadIn.tsx`
- `src/test/providers/navigation.test.tsx`

### DIFF
--- a/src/components/WalletLoadIn.tsx
+++ b/src/components/WalletLoadIn.tsx
@@ -30,7 +30,7 @@ export function WalletStaggerContainer({
       initial='initial'
       animate={hold ? 'initial' : 'animate'}
       variants={walletLoadInContainer}
-      style={{ contain: 'layout style paint', width: '100%' }}
+      style={{ contain: 'layout style', width: '100%' }}
     >
       {children}
     </motion.div>
@@ -61,7 +61,7 @@ export function WalletStaggerChild({
       className={className}
       variants={walletLoadInChild}
       style={{
-        contain: 'layout style paint',
+        contain: 'layout style',
         width: '100%',
       }}
     >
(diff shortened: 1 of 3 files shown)


sahilc0:
@bordalix this seems to have fixed it. Mind double checking/reviewing if the solution makes sense?

sahilc0:
@coderabbitai review

bordalix (approved):
tACK

Waiting for tests to be green to merge.

PR #749, opened 2026-07-09
feat(swap): price from discovered markets via the discovery protocol v0

tiero:
Implements the consumer side of the Arkade Market Discovery Protocol v0 (revised spec: `base_asset`/`quote_asset` descriptor objects, human-readable `pair` label, identity = id pair) and wires it into the swap screen so the quote is priced from the discovered market's own feed instead of the prototype USD estimate.

~1300 lines total: one module file, one wallet adapter, one hook, the swap-screen wiring, and tests.

## `src/lib/discovery/index.ts` — the protocol module

- **Pricing** (pure, bigint over scaled integers — no floats near amounts): `feedPrice` normalizes a raw feed observation to quote-units-per-base-unit via `invert`/`price_decimals` as an exact rational; `computeWantAmount` applies `floor(D * P * (1 - (fee_bps + safety_bps)/10000))`, symmetric with `1/P` for the reverse direction.
- **Index client**: `fetchIndex(baseUrl, network)` fetches `<base-url>/<network>.json`, requires `version === 0` and a matching `network`, rejects unknown fields loudly with their path, 10-min TTL cache. `loadSources` fans out over the followed registries with per-source error isolation. `mergeMarkets` unions per **id pair** (`base_asset.id/quote_asset.id` — the ticker label is display-only and never a grouping key), drops byte-identical duplicates, keeps cross-registry name collisions distinct, re-ranks ascending by `fee_bps` with source order as tiebreak. `staleness` flags indexes older than 7 days.
- **`quoteOffer`**: merge → walk the ranking, pricing each market from its pinned feed (a failing feed falls through to the next market) → `wantAmount`. Size bounds are base-denominated per spec: the deposit when depositing base, the `wantAmount` when wanting base. One error type, `DiscoveryError`, with a `code` UIs can switch on (`no-market` / `size` / `feed` with a CORS hint / `sources` / `validation`).

## `src/lib/discovery/wallet.ts` — wallet adapter

The wallet's asset ids pass through unchanged (`btc`, or the 68-hex on-chain id lowercased), so a market listed under DePix's asset id matches the wallet's DePix balance directly. Network mapping `bitcoin → mainnet`, `mutinynet → mutinynet`, etc. `quoteSwap` tries both pair orientations for a from/to selection. Registries default to `https://arklabshq.github.io/solver-registry`, overridable via `VITE_DISCOVERY_REGISTRIES` (comma-separated base URLs).

## Swap screen wiring

`useDiscoveryQuote` (debounced, network from `AspContext`) feeds the composer: receive amount and rate come from the discovered market's feed when one exists; the prototype USD estimate stays as fallback. Assets without a USD estimate (DePix) switch the keypad to asset-denominated input. The review drawer shows the winning solver and its fee.

## Tests

- Pricing goldens both directions, invert/decimals normalization, floor-at-boundary, safety-bps bound, exactness beyond 2^53.
- Index client: fixture parse, unknown-version/network-mismatch/unknown-field rejection, TTL cache, staleness math, dead-registry isolation.
- Merge: exact reduced-array assertions — cross-registry re-rank, byte-identical dedupe, name collisions kept distinct, ticker impostors (same label, different id) never matching, size filtering on a non-BTC-base pair.
- `quoteOffer`: best-market selection, feed-failure fallthrough, CORS path, size vs no-market vs all-sources-failed errors.
- Component tests: DePix → BTC on mutinynet against a mocked registry + feed, asserting the received BTC amount and solver attribution in review; no-market fallback shows "Swap unavailable for this pair".

## Notes for testing on mutinynet

- The wallet follows `https://arklabshq.github.io/solver-registry` by default and fetches `<base-url>/mutinynet.json` from it; other registries can be added via `VITE_DISCOVERY_REGISTRIES`.
- Matching is by on-chain asset id, so the DePix card's `base_asset.id` must equal the id the wallet holds the 30000 DePix under.
- User-pinned local solver cards (spec §Maker flow 3) are deferred until there's UI for them — easy to add back on the same `loadSources` path.
- Pre-existing failures on the base branch (unrelated): `delegates.test.tsx` and a `TransactionsList.tsx` type error.

https://claude.ai/code/session_01RGXto7ocJuQM13Ad5GHLbY


src/hooks/useDiscoveryQuote.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export function useDiscoveryQuote({
+  fromAssetId,
+  toAssetId,
+  depositAmount,
+}: {
+  fromAssetId?: string
+  toAssetId?: string
+  /** Deposit in the from-asset's smallest units. */
+  depositAmount: bigint
```
louisinger, line 33: allow discovery from "toAsset"

PR #767, opened 2026-07-13
Show eight decimals for non-zero BTC amounts

sahilc0:
## Summary

- Show every non-zero BTC display amount with exactly eight decimal places, including trailing zeroes.
- Keep the zero exception as `0 BTC`.
- Reuse the same presentation formatter for converted amount readouts so wallet, send, receive, balance, and transaction UI stay consistent.
- Preserve the existing formatter function signatures for caller compatibility.

## Scope and reviewer note

This changes presentation formatting only. It does not change wallet values, satoshi/BTC conversion rules, transaction construction, APIs, persistence, or backend behavior.

@bordalix: the shared UI formatter is intentionally used so the rule applies consistently anywhere BTC is displayed. Because that formatter is reused across several screens, could you please confirm that this is the right shared presentation-layer scope and that you do not see an unintended downstream display impact? Editable amount fields still preserve the user's raw text while they are typing; formatted BTC readouts use the new rule.

## Visual verification

- Wallet asset row: zero renders as `0 BTC`.
- Receive amount dialog: a non-zero conversion renders as `0.00160986 BTC` (exactly eight decimal places).
- Fresh screenshots were captured from the final local build and are included in the implementation handoff.

## Testing

- `bun run test:unit` — 47 files passed, 316 tests passed, 3 existing skips.
- `bun run lint` — passed.
- `bun run build` — passed.
- Focused formatter/wallet/send tests — 47 passed.
- Browser verification at a 375×667 viewport — zero and non-zero states verified; no console warnings or errors.

## Affected files

- `src/lib/format.ts` — central BTC display rule and zero exception.
- `src/components/InputAmount.tsx` — reuse the BTC presentation formatter for converted readouts.
- `src/test/lib/format.test.ts` — formatter coverage for zero, trailing zeroes, and fixed BTC precision.
- `src/test/screens/wallet/index.test.tsx` — wallet zero-state coverage.
- `src/test/screens/wallet/send.test.tsx` — send conversion coverage for eight decimals.

No unrelated repository files or behavior were changed.

### DIFF
--- a/src/components/InputAmount.tsx
+++ b/src/components/InputAmount.tsx
@@ -2,7 +2,7 @@ import { useContext, useEffect, useRef, useState } from 'react'
 import { FiatContext } from '../providers/fiat'
 import InputContainer from './InputContainer'
 import { ConfigContext } from '../providers/config'
-import { fromSatoshis, prettyNumber, toSatoshis } from '../lib/format'
+import { formatBitcoinAmountParts, prettyNumber, toSatoshis } from '../lib/format'
 import { FIAT_SYMBOLS } from '../lib/fiat'
 import { LimitsContext } from '../providers/limits'
 import { AssetOption, Unit } from '../lib/types'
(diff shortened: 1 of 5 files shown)


sahilc0:
@bordalix Could you please double-check the shared formatter scope here? The requirement is eight decimal places for every non-zero BTC display, with zero remaining `0 BTC`. We kept the change in the shared presentation formatter so the UI is consistent, preserved its existing function signatures, and did not touch wallet values, conversions, transactions, APIs, persistence, or other backend logic. The main review question is whether applying this through the common BTC display helper is the right minimal scope across all UI surfaces.

bordalix:
Substituted by #777

PR #782, opened 2026-07-15
removes concept of tabs

bordalix:
With the new UI, the concept of tabs no longer makes any sense, and is just dead code.


tiero:
I like removing code!

PR #788, opened 2026-07-16
New swap form

bordalix:
@tiero please test and review


bordalix:
Note: unit tests fail for markets.
It needs a patch applied to https://github.com/arkade-os/solver-registry/pull/5

tiero (approved):

PR #825, opened 2026-07-21
fix(wallet): restore BRL chart history

### LINKED ISSUE #807: BRL chart doesnt seem to be working
Image

sahilc0:
## Summary

- fetch BTC/BRL candles from Binance's public market-data endpoint for BRL chart ranges
- fall back to the existing Arkade chart proxy when Binance is unavailable
- preserve abort behavior and all existing USD/EUR/other-currency paths
- add regression coverage for BRL parsing and fallback behavior

This is intentionally stacked on #784.

Closes #807.

## Root cause

The current BRL proxy path uses CoinGecko and is being rate-limited. The wallet then receives no usable history and renders its constant current-price fallback, producing the flat 0% chart reported in #807. Current prices still work, which is why only the history appeared broken.

## Visual verification

Verified in Arc's Work profile with the funded local wallet and display currency set to BRL. The 1W chart rendered a varying BTC/BRL history with a +2.65% delta instead of a flat line.

## Testing

- `bun run test:unit` — 59 files passed; 432 tests passed, 3 skipped
- `bun run lint`
- `bun run build`
- live BRL history checks for 1H, 1D, 1W, 1M, 1Y, and All
- Arc Work-profile verification of the rendered BRL chart

## Affected files

- `src/lib/marketData.ts` — BRL data source, validation, and fallback
- `src/test/lib/marketData.test.ts` — BRL success and fallback tests

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying historical BRL market data.
  * BRL price history is retrieved and converted into chart-ready time and value points.
  * Added fallback handling when the public market data source is unavailable.

### DIFF
--- a/src/lib/marketData.ts
+++ b/src/lib/marketData.ts
@@ -2,6 +2,7 @@ import { LivelinePoint } from 'liveline'
 import { Currencies } from './types'
 
 const host = 'https://price-chart-proxy.arkade.money'
+const binanceHost = 'https://data-api.binance.vision/api/v3/klines'
 
 const DEFAULT_CONSTANT_SERIES_WINDOW_SECS = 86_400
 const CONSTANT_SERIES_INTERVALS = 100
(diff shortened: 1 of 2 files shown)


sahilc0:
@coderabbitai review

bordalix:
NACK

The problem is on https://price-chart-proxy.arkade.money/ which is serving outdated data or even no data at all.
We fix that and all graphs for all currencies for all periods will be fixed.

Issue #821, opened 2026-07-21
Swap: no Brazilian flag when swapping BTC → BRL after spending max BRL


theDavidCoen:
> Please tag with `walletv2` (I don't have label permissions on this repo).

### Summary
After sending max BRL, when swapping **BTC → BRL** the Brazilian flag is not shown. Unclear if intentional.

Related inconsistency: asset picker shows **Decentralized Pix** (DePix icon), while **Your swaps** titles use **BRL to BTC** even when the detail line says DePix.

### Steps
1. Send / spend max BRL
2. Open Swap → BTC → BRL / DePix (receive)
3. Compare asset picker icon vs route labels in Your swaps

**Expected:** Consistent BRL branding with Brazilian flag, or DePix naming/icon everywhere  
**Actual:** No Brazilian flag; DePix icon in picker while history titles say BRL

### Environment
Mutinynet preview · Android Chrome  
https://claude-swap-ui-mutinynet-ykt.arkade-wallet.pages.dev/

### Evidence
![DePix picker / BRL labels](https://raw.githubusercontent.com/theDavidCoen/wallet/qa-evidence/2026-07-21-swaps/qa-evidence/2026-07-21/issue2-depix-no-br-flag.jpg)

[Video](https://github.com/theDavidCoen/wallet/releases/download/qa-evidence-2026-07-21/issue2-flag-and-slow.mp4)

Slack: https://arklabshq.slack.com/archives/C08KU7WQ8T0/p1784645878299759

PR #832, opened 2026-07-22
fix(swaps): preserve BRL branding at zero balance

sahilc0:
## Summary

- Preserve the verified BRL identity for Mutinynet DePix in the swap asset list even when its wallet balance is zero and no portfolio row exists.
- Keep the Brazilian flag and BRL ticker consistent in the BTC → BRL picker and in swap display snapshots created from that selection.
- Add a regression test for the exact post-“send max” condition reported in #821.

Addresses #821.

## User-visible behavior

After spending the full BRL balance, BTC → BRL still appears as BRL with the Brazilian flag. The picker no longer falls back to Decentralized Pix / DePix branding, and subsequent swap history uses the BRL ticker consistently.

## Affected files

- `src/screens/Wallet/Swap/Index.tsx` — falls back to the existing verification-gated asset designation when a zero balance removes the portfolio row.
- `src/test/screens/wallet/swap.test.tsx` — allows wallet-state overrides and verifies zero-balance BRL naming plus Brazilian flag rendering.

No files were added, moved, or deleted. No unrelated files or behavior changed.

## Verification

- `bun run test:unit` — 59 files passed; 431 tests passed, 3 skipped.
- `bun run lint` — passed.
- Mobile visual verification at 375 × 667 in the running Mutinynet app with a newly generated empty test wallet — confirmed `0.00000000 BRL`, the Brazilian flag, and no DePix label in the BTC → receive picker.
- Independent GPT-5.6 code review — approved with no substantive issues.

## Visual proof

Empty Mutinynet wallet, BTC → receive picker at 375 × 667:

issue-821-brl-zero-balance

### DIFF
--- a/src/screens/Wallet/Swap/Index.tsx
+++ b/src/screens/Wallet/Swap/Index.tsx
@@ -39,6 +39,7 @@ import { NavigationContext, Pages } from '../../../providers/navigation'
 import { WalletContext } from '../../../providers/wallet'
 import { usePortfolioFiat, type PortfolioRow } from '../../../hooks/usePortfolioFiat'
 import { useReducedMotion } from '../../../hooks/useReducedMotion'
+import { verifiedDesignatedCurrency } from '../../../lib/accountAssets'
 
 type AssetTarget = 'from' | 'to'
 type DrawerState = 'to' | 'review' | null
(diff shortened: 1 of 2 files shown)


tiero:
tACK

PR #834, opened 2026-07-22
Polish swap amount entry and denomination switching

### LINKED ISSUE #811: add a "swap max" button to swaps
or swap all

### LINKED ISSUE #820: Swap: secondary amount pill is redundant when currency matches send asset (+1 sat on BTC/sats toggle)
> Please tag with `walletv2` (I don't have label permissions on this repo).

### Summary
When **Currency** matches the **send asset**, the swap composer shows the same amount twice (main input + secondary pill with ↑↓).

Seen with BTC/sats (`1,000 sats` + `1,000 sats`) and with BRL as currency + BRL as sender (`1,000 BRL` + `1,000.00000000 BRL`).

Additionally, with **Currency = BTC** and **Unit = sats**, tapping the pill bumps the amount by **1 sat** (1000 → 1001). The +1 bump does **not** happen on other assets (e.g. BRL).

### Steps — redundant pill
1. Settings → Display → Currency = same asset you will send (BTC/sats or BRL)
2. Swap → that asset as sender → enter an amount
3. Observe secondary pill under the amount

**Expected:** different denomination, or pill hidden when there is nothing to convert  
**Actual:** same asset/amount repeated

### Steps — +1 sat (BTC/sats only)
1. Currency BTC, Unit sats
2. Enter `1000`, tap ↑↓ on the secondary pill

**Expected:** amount unchanged  
**Actual:** `1,000 sats` → `1,001 sats`

### Environment
Mutinynet preview · Android Chrome  
https://claude-swap-ui-mutinynet-ykt.arkade-wallet.pages.dev/

### Evidence
sats dual pill

BRL dual pill

[Video — sats toggle / +1](https://github.com/theDavidCoen/wallet/releases/download/qa-evidence-2026-07-21/issue1-sats-toggle.mp4)

Slack: https://arklabshq.slack.com/archives/C08KU7WQ8T0/p1784640450503379

sahilc0:
## Summary

- Adds the approved compact **Use max** action while keeping the available balance non-interactive.
- Places the denomination control beneath the primary amount, preserves exact entered values when denominations exchange priority, and keeps the arrow stable while the pill resizes.
- Exposes the live amount through the amount control, uses layout-driven motion instead of token-coupled offsets, and retains repeated-character animation safety.
- Keeps swapping available when currency conversion is unavailable: unpriced assets enter in asset units, omit fabricated currency values, and complete through solver data alone.
- Hides the secondary pill when conversion would repeat the same denomination, including BTC/sats and designated BRL/BRL, preventing identity-conversion rounding drift.
- Applies the same unavailable/identity conversion behavior to Send.
- Incorporates the UI/UX review fixes for dark-mode control contrast, pointer/touch drawer focus and scroll restoration, and reduced-motion behavior.
- Merges the latest `claude/swap-ui-mutinynet-yktp51` target branch and resolves Send’s fiat-first overlap without rewriting PR history.
- Prevents conversion availability changes from silently reinterpreting entered digits in Swap, Send, and the mobile keyboard.
- Floors zero-decimal asset conversions at the conversion source before quote formatting.

## User-visible behavior

- An unpriced give asset no longer dead-ends the composer. Entry, quote, review, and confirmation continue in asset units.
- Currency text is suppressed per asset: an unpriced give side does not fabricate `$0.00`/`€0.00`, while a priced receive side keeps its valid estimate.
- Quote snapshots omit `fromFiatAmount` when the give side has no conversion.
- BTC with BTC currency and designated BRL with BRL currency show one meaningful denomination instead of a redundant toggle.
- Exact atomic values survive quote and confirmation in identity cases.
- Send starts in display-currency mode when conversion is useful, with the configured bitcoin unit as the secondary value.
- If conversion disappears mid-entry, Swap clears the currency amount before switching to asset units; Send re-expresses the already-resolved sats amount in the configured bitcoin unit.
- The mobile keyboard clears text whenever conversion availability changes its effective denomination, preventing the same digits from changing meaning.
- Verified Mutinynet DEPIX remains presented as the BRL account; an unverified lookalike falls back to its raw asset identity.
- “Use max” and the direction control are legible in dark mode.
- Opening the receive-asset drawer with pointer/touch input does not summon the mobile keyboard, and closing it does not scroll the Swap header away.
- Keypad entrance, direction rotation, and confirmation-error motion respect reduced-motion preferences.

## Scope

- `src/components/InputAmount.tsx` — gates the secondary currency label and switcher on a useful, available conversion.
- `src/components/Keyboard.tsx` — derives a stable default input mode and clears typed text when conversion availability changes the effective denomination.
- `src/index.css` — preserves the intrinsic denomination layout, removes fragile descendant-button styling, and fixes dark-mode Swap control contrast.
- `src/screens/Wallet/Send/Form.tsx` — preserves the base branch’s fiat-first Send behavior and safely re-expresses resolved sats when conversion becomes unavailable or redundant.
- `src/screens/Wallet/Swap/Index.tsx` — implements per-asset useful-conversion handling, identity behavior, quote/snapshot safeguards, safe conversion-loss fallback, zero-decimal flooring, drawer focus behavior, and reduced-motion fixes.
- `src/test/screens/wallet/send.test.tsx` — covers fiat-first entry, unavailable conversion, safe mid-entry fallback, and BTC identity parity.
- `src/test/screens/wallet/swap.test.tsx` — covers unpriced assets in both directions, conversion loss during entry, partial pricing, identity conversions, exact atomic amounts, zero-decimal handling, preselection loading, semantic selectors, and pointer drawer focus.

No backend behavior, dependencies, or unrelated product files are changed by this PR. Files brought into the branch by target-branch merges are already part of `claude/swap-ui-mutinynet-yktp51`; the seven files listed above are the complete PR diff against the current target.

The reviewed `selectToAsset` same-asset path is not reachable in the current UI: `findMarket` returns no market when the asset IDs match, so the current give asset is excluded from the receive list.

## Verification

- `bun run test -- src/test/screens/wallet/swap.test.tsx src/test/screens/wallet/send.test.tsx` — 47/47 passed.
- `bun run test` — 479 passed, 3 skipped across 64 files.
- `bun run lint` — passed.
- `bun run build` — passed.
- `git diff --check` — passed.
- Commit hooks reran formatting and lint successfully.
- Verified in the Codex in-app browser at 375×667, 768×1024, 1440×900, and 1920×1080.
- Verified 44px minimum primary touch targets, stable drawer scroll position, pointer/touch focus behavior, successful quote/confirm flow, and no browser console errors.
- Fresh post-change verification in the running app confirms Send opens in display-currency mode with the bitcoin amount as its secondary value.
- GitHub reports the PR mergeable with no base conflicts; both Cloudflare deployments and CodeRabbit are successful.

The local build emits the repository’s existing warnings: Node 24.4.1 is below the requested 24.15.0, the public worker output overlaps `publicDir`, the mnemonic module is imported both dynamically and statically, and the production bundle exceeds the configured chunk-size warning threshold. The build completes successfully.

### Latest post-change Send verification

The merged Send flow opens in display-currency mode and keeps the bitcoin amount as the secondary value.

Post-change Send fiat-first verification

### Latest post-fix mobile verification

The verified BRL identity case shows one meaningful denomination, improved dark-mode controls, a valid priced receive estimate, and an enabled primary action.

Post-fix BRL identity swap

### Amount-control and denomination motion

Accessible amount control

Layout-driven denomination transition

## Tracking

Targets the head branch of #784 as requested.

The non-blocking price-tick staleness follow-up remains deferred to #835.

Closes #811
Closes #820


sahilc0:
@coderabbitai review

tiero (changes requested):
Verified locally: 28/28 tests pass (CI didn't run — GitHub Actions is currently not triggering on this repo). The new coverage is good. Four blockers before this can land:

**1. Fix the accessibility regression on the amount control.**
`.swap-amount-display` is now an empty button (`aria-label='Swap amount'`, no content) and every visible value lives inside an `aria-hidden` sibling (`.swap-amount-values`). The entered amount is no longer exposed to assistive tech anywhere in the composer. Requirement: the current amount value must be programmatically readable from the button (e.g. keep the primary `AnimatedAmountValue` inside the button, or mirror the value into the button's accessible name whenever it changes).

**2. Restore the deleted regression test and comment.**
The test `'never assigns the same React key to two occurrences of the same letter in the amount label'` and the explanatory comment in `amountCharacterSlotKey` were removed. The keying logic they guard still exists and the bug it prevents (repeated-glyph smear in "sats") was real. Re-add both.

**3. De-magic the animation constants.**
`translateY(-27px)` / `translateY(25px)` / `scale(0.3)` are hardcoded in JSX, with `0.3` duplicated in CSS (`calc(var(--heading-2xl-size) * 0.3)` and again in the 380px media query), plus a hidden `::before` reading `data-reserve-width` to reserve pill width. Any type-token change silently breaks centering. Requirement: derive these from one place (CSS custom properties consumed by both the JSX transforms and the CSS calc), or replace the offset/scale choreography with a layout-driven approach that has no per-token constants.

**4. Get explicit sign-off on removing the tappable balance.**
Making the balance non-interactive and adding "Use max" reverses a deliberate product decision (tap-the-balance was requested by @tiero). Don't ship this inside a polish PR without his explicit approval in this thread.

**Also (non-blocking but flagged):** the new staleness check compares the plan against a *price-derived* amount (`planMatchesAssetAmount` on `assetAmount`, which is recomputed from the live USD estimate) instead of the entered value — it will re-flag quotes as stale on price ticks with no user input. Consider keying staleness on the entered input identity instead.

**Sequencing:** this PR and #835 rewrite the same mechanisms incompatibly (quote staleness, `buildQuoteFromPlan` signature, `SwapComposer`, same test file). Recommended order: #835 merges first; rebase this PR's polish onto its request-key staleness design rather than keeping two competing implementations.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

---

tiero:
its ok to remove the tappable balance

sahilc0:
@bordalix @tiero — the requested review changes are addressed in `31fc53dd`:

- the live primary amount is rendered inside the amount button and included in its accessible name
- the repeated-letter React-key regression test and explanatory comment are restored
- the hardcoded offsets/scales and hidden width reservation are replaced with layout-driven motion
- the non-tappable balance remains as approved by @tiero

The non-blocking price-tick staleness follow-up remains deferred to #835. Verification is green: 31/31 focused swap tests, 459 passed + 3 skipped in the full suite, formatting, lint, build, and all three current PR checks. The PR description now includes the updated scope and post-fix screenshot/GIF. Ready for another look.

tiero (changes requested):
@sahilc0 Re-review of `31fc53dd` through a YAGNI / consistency / DRY lens. All four items from the previous review are genuinely addressed — the accessible name on the amount button, the restored repeated-letter key test, layout-driven motion replacing the hardcoded offsets, and the non-tappable balance as approved. The `preservedAmountPair` mechanism also beats the base branch's floor-at-display-precision toggle: pinning `deposit.atomic === 1_093_180` through a round trip is the right property, and `buildQuoteFromPlan` losing its fiat-conversion branch is a good deletion.

Four items before merge, three mechanical, one to verify:

**1. DRY — delete `fiatValueFromAssetAmount` (Index.tsx ~1316) and collapse `currentFiatValue`.**
When `currentPlan` is null, `quote` is built with a null plan, so `quote.fromFiatValue` already *is* the oracle give-estimate (`giveEstimate` inside `buildQuoteFromPlan`) that this helper recomputes with Decimal. The branch collapses to:
```ts
const currentFiatValue = quote.fromFiatValue > 0 ? quote.fromFiatValue : undefined
```
Delete the helper (~15 lines). One source of truth for oracle pricing — two parallel implementations of the same conversion is how the give/receive sides drifted apart last time.

**2. Consistency — new tests query by CSS class.**
`document.querySelector('.swap-amount-value--primary')` / `--secondary` / `.swap-amount-secondary__label` / `.swap-input-card__balance` appear ~10× in the new tests. Repo convention is roles/labels/testids — and #847 schedules these exact class names for deletion in the Tailwind migration, so every one of these queries is a future test break. You already added the accessible names; assert through them (`getByRole('button', { name: 'Swap amount, $10' })`, the `aria-label` on `.swap-amount-value`) or add `data-testid`s where no accessible name exists.

**3. Consistency — `.swap-input-card__asset button:not(.swap-input-card__max)` (index.css).**
Styling every descendant `button` and then carving exceptions with `:not()` is the fragile pattern that forced this selector in the first place. Class the asset trigger explicitly (e.g. `.swap-input-card__asset-trigger`) and style the two buttons independently.

**4. Verify — `quoteStale` before a receive asset is selected.**
`quoteStale = Number(assetAmount) > 0 && !currentPlan` is now true whenever no matching plan exists — including *before* a receive asset is chosen, where a plan can never arrive, making `quoteLoading` true with any positive amount. Confirm nothing renders a perpetual loading/skeleton state in that pre-selection window (the old `quotedAmount === amount` check went false after the debounce regardless of plan). A one-line guard (`toAsset &&`) or a test pinning the pre-selection state would settle it.

Nit, no change required: `preservedAmountPair` is cleared in four mutation paths and guarded by `assetId` — add one comment at the state declaration naming the invariant ("any path that changes the effective amount must clear this") so a fifth mutation path added later doesn't leak a stale pair.

Sequencing: with this PR now ahead, land #834 first; #835 (currently clean, same files) rebases on top.

---

tiero:
@sahilc0 additional scope for this PR, approved by @tiero — **unpriced-asset support**. An asset with no USD proxy (`estimateSwapUsd` → 0: not BTC, not a stablecoin, no designated currency — e.g. a regtest asset or a novel asset someone runs a solver for) currently dead-ends the composer: the default entry mode is `fiat`, whose conversion returns nothing, so typing never produces a quote. The same happens for *every* asset when the price feed is unreachable (regtest offline). Currency conversion is a UX convenience — it must never gate swapping.

Requirements (the rule is **per-asset**: "conversion unavailable" = `asset.usdPrice` falsy OR `unitOfAccountUsd` is 0):

1. **Entry never dead-ends.** If the give asset has no conversion, default `amountMode` to `'asset'` and hide the denomination toggle for that asset. Entry always works in asset units.
2. **Suppress currency display per-asset — no fabricated zeros.** Wherever an unpriced asset's currency value would render ($0.00 secondary label, receive-side currency line, review-drawer value rows), hide the line instead. If only one side of the pair is unpriced, only that side's currency line hides (the BTC side keeps its conversion).
3. **Swapping never blocks on rates.** Quote → plan → validate → confirm must flow through solver data alone. Acceptance: with the price feed fully unreachable, both a BTC→novel-asset and a novel-asset→BTC swap complete end-to-end with no currency text anywhere.
4. **Quote snapshot:** when the give side has no conversion, persist no `fromFiatAmount` (leave it undefined — receipts already fall back to valuing the BTC leg at the current rate).
5. **Send-flow parity:** apply the same per-asset rule to Send (`InputAmount` `switchable` + keyboard `defaultMode`): unpriced asset → unit-only entry, no ⇅ switcher, no currency sublabel. Verify the switcher actually hides today rather than assuming.
6. **Naming:** the selected unit of account can be Bitcoin, so new identifiers introduced by this PR should say *currency*, not *fiat* (e.g. `alternateFiatAmount` → `alternateCurrencyAmount`, `fiatAmountLabel` → `currencyAmountLabel`). Touched code only — no repo-wide rename.
7. **Tests:** an asset fixture with `usdPrice: 0` → asserts default asset-unit mode, absent toggle, entry→quote→confirm succeeds, and no `$`/`€` string renders for that asset; plus one Send-form parity test.

This is in addition to the 4 open items from the review above — same bar for both.

---

tiero:
@sahilc0 one more rule for this PR's scope, companion to the unpriced-asset spec above — it closes #820.

**Hide the secondary pill and the denomination toggle when the give side's conversion is the identity** — i.e. the pill would repeat the primary label's denomination:

- the give asset's designated currency equals `config.currency` (BRL give with BRL currency, USD give with USD currency), or
- the give asset is BTC and `config.currency` is BTC (the composer always displays BTC amounts in sats, so this pair is always identical).

This is the twin of the unpriced rule: the pill/toggle exist only when they convert to a *genuinely different* denomination. Unpriced → no conversion → hide; identity → nothing to convert → hide. One predicate family.

It also fixes #820's second half for free: the ±1 sat drift on toggle is float noise from running an identity conversion through the rate machinery — no toggle in the identity case, no conversion, no drift (the structural noise source is tracked separately in #851).

**Tests:** fixtures for currency=BTC/unit=sats with BTC give, and currency=BRL with BRL give, asserting: no secondary pill rendered, no toggle, and `1000` entered stays exactly `1000` through quote → confirm. The existing different-denomination cases keep the pill.

---

sahilc0:
Updated as requested in `8593bfde`; the latest PR #784 base was merged in `68a11dd4` and the conflict is resolved.

Changes addressed:
- unpriced assets now enter and swap in asset units without fabricated currency values or a dead-end composer
- identity conversions (BTC/sats and designated BRL/BRL) hide the redundant denomination control and preserve exact atomic amounts
- Send uses the same unavailable/identity conversion behavior
- the priced receive side retains its currency estimate when only the give asset is unpriced
- quote snapshots omit `fromFiatAmount` when the give side has no conversion
- dark-mode control contrast, pointer/touch drawer focus and scroll behavior, and reduced-motion handling were reviewed and fixed

Verification after merging the updated base:
- focused Swap + Send: 46/46 passed
- full suite: 478 passed, 3 skipped across 64 files
- lint, formatting, production build, and `git diff --check`: passed
- verified at 375×667, 768×1024, 1440×900, and 1920×1080 with no browser console errors

Post-fix visual evidence:
pr834-ui-review-brl-quoted

sahilc0:
Follow-up updated in `1b23e2bc` after the target branch advanced again and Arkana's review landed.

Addressed:
- merged the latest `claude/swap-ui-mutinynet-yktp51` base and resolved the Send conflict while preserving its fiat-first behavior when conversion is available
- conversion availability changes no longer silently reinterpret typed digits in Swap, Send, or the mobile keyboard
- Swap resets a currency entry before falling back to asset units; Send re-expresses the already-resolved sats value in the configured bitcoin unit
- zero-decimal asset units are floored at the conversion source before quote formatting
- verified the `selectToAsset` same-asset concern is not reachable in the current UI because the give asset has no receive market and is excluded from the receive list

Verification:
- focused Swap + Send: 47/47 passed
- full suite: 479 passed, 3 skipped across 64 files
- lint, formatting, production build, and diff checks: passed
- fresh post-change verification in the running app confirms Send opens in display-currency mode with the bitcoin amount as the secondary value

Post-change visual proof:
pr834-followup-send-fiat-first

tiero:
@sahilc0 status check on the outstanding asks, verified against `8593bfde`…`1b23e2bc`:

**Landed well ✓** — the identity/unpriced work is exactly what was specced: `hasDistinctCurrencyConversion` covering both identity flavors with tests, asset-unit fallback with no fabricated currency values, the snapshot omitting `fromFiatAmount` when unpriced, Send/Keyboard parity, *currency* naming on the new identifiers, and the `1b23e2bc` fixes for the mode-transition reinterpretation Arkana flagged. The `planMatchesAssetAmount` Decimal gate is also a genuine upgrade over the string comparison.

**Still open ✗** — the four items from the standing REQUEST_CHANGES review at `31fc53dd` (Arkana's incremental review also lists them as unaddressed, and the `1b23e2bc` diff shape confirms: no helper deletion in Swap/Index.tsx, test additions only, index.css untouched):

1. **DRY**: delete `fiatValueFromAssetAmount`; `currentFiatValue` collapses to `quote.fromFiatValue > 0 ? quote.fromFiatValue : undefined` (when `currentPlan` is null the quote already carries the oracle give-estimate).
2. **Tests**: replace the ~10 CSS-class `document.querySelector` assertions with the accessible names you added (or testids) — #847 deletes those class names.
3. **CSS**: class the asset trigger explicitly instead of `button:not(.swap-input-card__max)`.
4. **Verify `quoteStale` pre-selection**: confirm no perpetual loading state before a receive asset is chosen, or add the one-line `toAsset &&` guard / a pinning test.

All four are mechanical; once they're in, the review flips to approve from my side.

---

tiero (approved):
Handled the remaining cleanup directly in `561c86e` (base sync + review closeout, at Marco's request) — **all four open items are now resolved, flipping to approve**:

1. **DRY collapse** — already done on this branch (`currentCurrencyValue = quote.fromCurrencyValue > 0 ? … : undefined`, duplicate helper deleted). Confirmed in the merged tree.
2. **Class-query tests** — rewritten to roles/accessible names (`findByRole('button', { name: /^Use maximum/ })`, `primaryAmount()`, `[aria-live]`/`[data-sonner-toast]` attribute checks). The one remaining `querySelector` is an SVG-fill assertion with no accessible alternative.
3. **`:not(.swap-input-card__max)`** — moot and deleted: the base now carries #845's finalized balance-slot, which makes the separate Use max control redundant; the button and every `.swap-input-card__max` rule are removed.
4. **`quoteStale` pre-selection** — already guarded with `toAsset &&` on this branch. Confirmed, plus the existing pre-selection test still passes.

Merge notes (the #845 collision this PR owed): the composer keeps this branch's layout-driven entry, preserved-pair toggle, and Decimal plan-freshness gate; the balance slot, toast routing, and error-pill removal come from #845 as merged. `invalidPulse` plumbing went with the old structure — the shake retriggers on validation-state transitions without remounting the digit animation, exactly as this branch's mount-stability test demands (test updated to observe the Max action instead of the now-unrendered "Insufficient balance" text; the two Use-max tests now go through the tappable balance).

Gate on the pushed head: tsc, 482 unit tests, prettier, eslint — all green.

---

PR #835, opened 2026-07-22
Add exact receive amount entry to swaps

### LINKED ISSUE #795: Make it possible to choose Receive amount
Currently, on swap ui you can only say how much to SEND not how much to receive. what if i am denominating in euros, and want to convert some bitcoin to receive 5 USD?

see revolut for example

Image

also no exchange rate we want to show that

sahilc0:
## Summary

- Keep the existing send-first swap composer while making the receive amount directly editable.
- Quote exact receive targets through the solver's want-side API and protect typed values from stale quote responses.
- Add a temporary floating variant switcher on the wallet and swap screens so seven interaction models can be compared against the same live swap data.
- Incorporate Tiero's two proposed directions: a flippable primary send/receive field and equal editable inputs.
- Add a header-level `Swap from` / `Swap to` dropdown that can change the first asset-selection step as well as the primary amount.

This is intentionally stacked on #784 and should merge into that PR's branch.

Closes #795

## User-visible behavior

- V1 **Current** preserves the original send-first flow and exact-receive drawer.
- V2 **Receive first** starts by choosing the asset and amount to receive.
- V3 **Send or receive** asks which side to start with.
- V4 **Promote field** flips which amount is primary while retaining one keypad.
- V5 **Equal inputs** makes both amount cards equally editable with the shared keypad.
- V6 **Native keyboard** lets either amount open the system keyboard instead of rendering the custom keypad.
- V7 **Header dropdown** renders `Swap from` / `Swap to` in the header. Only `from` / `to` is the dropdown, and choosing `to` changes the first question to “Choose asset to receive.”
- Denomination toggles work on every editable card and exact receive quotes remain denominated correctly when the wallet uses another currency.
- The switcher is available on the wallet home screen so receive-first variants can change the first step before Swap opens.
- Reduced-motion preferences, focus states, touch targets, and existing haptic feedback remain supported.

## Scope

- `src/components/Header.tsx` — allows composed header content so V7 can place the direction dropdown directly beside `Swap`.
- `src/components/SwapVariantSwitcher.tsx` — adds the reusable temporary V1–V7 floating switcher used on both wallet and swap screens.
- `src/index.css` — styles the exact-receive surfaces, variant switcher, alternate composers, native input, responsive layouts, and header dropdown.
- `src/providers/flow.tsx` — stores the selected temporary variant across the wallet and swap screens.
- `src/screens/Wallet/Index.tsx` — exposes the variant switcher before entering Swap.
- `src/screens/Wallet/Swap/Index.tsx` — implements exact receive entry, quote protection, the seven prototype flows, denomination switching, native input, and header `from` / `to` behavior.
- `src/test/screens/mocks.ts` — supplies the default variant state to shared screen mocks.
- `src/test/screens/wallet/index.test.tsx` — verifies variant selection from the wallet home screen.
- `src/test/screens/wallet/swap.test.tsx` — covers exact receive quoting, stale responses, asset changes, validation, all seven variants, unit switching, native input, and the V7 first-step transition.

No unrelated tracked files or behavior were changed.

## Verification

- `bun run test:unit` — 463 passed, 3 skipped across 63 test files.
- `bun run test --run src/test/screens/wallet/swap.test.tsx` — 34/34 passed.
- `bun run lint` — passed.
- `git diff --check` — passed.
- `bun run build` — passed. Existing public/outDir, mixed mnemonic import, chunk-size, and local Node engine warnings remain unchanged.
- In-app mobile QA at 375×667 — verified both header options are available on the first step, `Swap to` changes the first question, the receive card becomes primary, its unit toggle works, and a live 5 USD target resolves to the bitcoin amount to send.

Not verified: submitting a real swap or physical-device haptics.

## Visual proof

Original exact-receive drawer:

Exact receive drawer

V7 header dropdown, first-step transition, and final exact receive quote:

v7-header-dropdown-final

v7-swap-to-first-step

v7-header-swap-to-final


sahilc0:
@coderabbitai review

tiero (changes requested):
Verified locally: 29/29 tests pass (CI didn't run — GitHub Actions is currently not triggering on this repo). The core is solid: want-side quoting goes through the solver's real `setWantAmount`, money paths are asserted at the atomic level (`receive.atomic`), and the a11y work (labels, `aria-haspopup`, `aria-live`, focus-visible, reduced-motion) is done properly. Four small fixes, then this is mergeable:

**1. Show the real receive-side limit instead of the generic message.**
In want mode, `below-min`/`above-max` render "Amount below minimum". But the user is typing the receive side and `plan.limits` ARE receive-side — no conversion needed: use `Minimum ${limit.display} ${toAsset.ticker}`. Caveat: `validatePlan` also enforces the market card's give-side bounds (`min/max_base_amount` / `min/max_quote_amount`) — when the give-side check is what tripped, keep the generic message. Distinguish which check fired.

**2. Delete the exploration residue in tests.**
`expect(screen.queryByRole('button', { name: 'Choose design variant' })).not.toBeInTheDocument()` asserts the absence of a button that never existed in this code — leftover from the variants exploration branch. Remove the line.

**3. Remove the `!important`.**
`.swap-receive-amount-drawer { max-height: min(92dvh, 42rem) !important; }` — resolve the specificity conflict with the drawer library properly (a more specific selector or a variable the drawer respects).

**4. Decide on the duplicated rate line.**
The `1 BTC = X` rate now renders in both the receive card and the review drawer — that's why the `/^1 BTC = /` uniqueness assertion had to be loosened to `getAllByText`. If the duplication is intentional, say so in the PR body; otherwise drop the card copy and restore the strict assertion.

**Sequencing:** this PR and #834 rewrite the same mechanisms incompatibly (quote staleness, `buildQuoteFromPlan` signature, `SwapComposer`, same test file). Recommended order: this PR merges first — its request-key staleness design is the cleaner of the two — and #834 rebases onto it.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

---

tiero:
I'm not convinced by the concept. We have a keypad and a bottom sheet keyboard, seems we loose consistency

I can see two paths 

1. We make "flippable" the Send and Receive verb, where the top input is always the one editable (remains bigger) just we flip Send/Receive action and meaning. We should add You send and You Receive titles well visible 

2. We make both form the same size and make editable them with same keypad, Uniswap style

sahilc0:
@tiero i hear you, worked on some more options, will have some variants + my recommendation soon

sahilc0:
@tiero Closing based on our decision to keep swaps exact-in/send-only so the flow remains compatible with future chained and split multi-hop routing. #784 already includes the asset-direction flipper and exchange-rate display. We can revisit exact-out in a separate issue if routing support is designed for it.

PR #860, opened 2026-07-29
Fix swap rate display to show pre-fee market price

tiero:
## Summary

The swap review screen's quoted price read consistently *lower* than Binance (and every other exchange). The Rate row was derived from the **net** receive amount, so the market fee (`fee_bps`) was baked into the quoted rate — even though the same fee is already itemized on its own Fees row on the same card. The rate now quotes the market feed's **pre-fee** price, so it matches the exchange price and the fee appears exactly once.

## The bug

The rate was computed as `receive ÷ deposit`. But the solver's planned receive amount is already net of the market fee, so the displayed rate was always `feed price × (1 − fee_bps/10⁴)`:

- **BTC/USDT** (fee 30 bps): rate read **0.3% below** its feed.
- **BTC/DePix** (fee 110 bps): rate read **1.1% below** Binance — and this market's price feed literally *is* Binance (`api.binance.com/.../ticker/price?symbol=BTCBRL`), which is why the divergence was so conspicuous and always in the "lower" direction.

This also double-represented the fee: it was hidden inside the rate *and* shown on the Fees row, so the card's numbers didn't reconcile.

## The fix

**Commit 901ba49** — the Rate row quotes the solver's pre-fee feed price instead of the net-derived rate, flipped (`1/P`) when the user gives the quote side. Swapping 10,000 sats at a $100,000 feed with a 30 bps fee:

| Row | Before | After |
|---|---|---|
| Swap | 0.00010000 BTC | 0.00010000 BTC |
| Receive | 9.97 USDT | 9.97 USDT |
| Fees | 0.03 USDT | 0.03 USDT |
| Rate | 1 BTC = 99,700 USDT | 1 BTC = **100,000** USDT |

The card is now arithmetically consistent: **Receive = Swap × Rate − Fees** (0.0001 × 100,000 − 0.03 = 9.97 ✓).

**Commit 525aff1** — follow-ups from review:

- **Exact rate math**: the first commit read `plan.priceDisplay`, which the SDK truncates at 8 fraction digits ("display only, never pricing"). A market whose quote-per-base display price is below 1e-8 (e.g. a registry publishing BTC as the *quote* asset over a sub-satoshi-priced base token) would render a `0` rate, and small prices lost significant digits — amplified through `1/x` on the give-quote side. The rate is now derived from the plan's exact `price` rational via the SDK's `displayPrice`. The helper lives in `lib/swap/markets.ts` (`preFeeDisplayRate`), unit-tested for both orientations and the sub-1e-8 case, and `formatLimitMessage`'s duplicate orientation flip is folded into it.
- **Receipt consistency**: the transaction receipt's "Price rate" row still derived a net-of-fee rate from the stored amounts, so the same swap showed `100,000` at confirm time and `99,700` in activity — and the receipt double-counted the fee against its own gross "Swap to" and "Swap fees" rows. It now grosses the stored net payout back up by the persisted `feeBps`, matching the composer. Restored swaps without `feeBps` keep the net fallback, consistent with the net "Swap to" row shown in that case.
- **Coverage**: the review drawer's give-quote (USD→BTC) inverted rate is now asserted (`1 USD = 0.00001587 BTC` at a $63,000 feed) — this also addresses CodeRabbit's review comment.

## How the fee deduction works (math unchanged by this PR)

No payout amounts changed, and the wallet does no manual fee arithmetic — the fee lives in the solver's payout math, where it always was:

1. The market's `price_feed` returns the raw market price (no fee in it).
2. The solver SDK (`@arkade-os/solver-discovery`, `computeWantAmount`) computes the payout with the fee deducted from fair value: `receive = deposit × price × (1 − fee_bps/10000)`. This net amount is what gets locked into the offer and enforced by the covenant. The wallet passes `safetyBps: 0`, so `fee_bps` is the *only* concession off the feed price.
3. The wallet's Fees row displays the gross−net gap (`received × fee/(1−fee)`), i.e. the same deduction the solver applied — shown once, in the receive asset.

So the fee was never "in the feed" and is not "added manually" now: what the user receives, what the solver enforces, and what gets persisted are byte-identical to before. The Rate rows (review drawer and receipt) simply stopped disguising the fee as a worse price.

Residual (small) divergence vs Binance can still come from the feed source itself — e.g. BTC/USDT prices off CoinGecko's index per the solver registry — and from quote staleness (the wallet caches feed responses for 30s and doesn't re-quote an idle review drawer). Those are registry/product decisions, out of scope here.

## Test plan

- `swap.test.tsx` pins the exact pre-fee rate for both orientations: `1 BTC = 100,000 USD` (give base) and `1 USD = 0.00001587 BTC` (give quote, inverted).
- New `preFeeDisplayRate` unit tests in `markets.test.ts` cover both orientations plus the sub-1e-8 display price the truncated `priceDisplay` would zero out.
- `transaction.test.tsx` asserts the receipt's gross-derived rate (`1 ALP = 0.55159174`) reconciling with its gross "Swap to" row.
- Full unit suite: 64 files, 510 passed / 3 skipped. Lint and Prettier clean.

https://claude.ai/code/session_01YLtvQ4AoGG11FyGQfbwByR

### DIFF
--- a/src/lib/swap/markets.ts
+++ b/src/lib/swap/markets.ts
@@ -1,6 +1,7 @@
 import {
   bestMarket,
   discover,
+  displayPrice,
   isNetwork,
   type DiscoveredMarket,
   type OfferPlan,
@@ -114,6 +115,21 @@ export const findMarket = (
   const givingBase = bestMarket(markets, { baseId: fromId, quoteId: toId, wantSide: 'quote' })
   if (givingBase) return { market: givingBase, give: 'base' }
   return { market: bestMarket(markets, { baseId: toId, quoteId: fromId, wantSide: 'base' }), give: 'quote' }
+}
+
+/** The market feed's pre-fee price oriented give→receive, in whole display
+ * units. Derived from the plan's exact price rational — plan.priceDisplay
+ * truncates at 8 fraction digits, which zeroes or skews small prices, and the
+ * give-quote inversion would amplify that loss. Assumes the wallet's
+ * safetyBps of 0 (QUOTE_OPTIONS): fee_bps is then the only gap between this
+ * rate and the plan's net payout. */
+export const preFeeDisplayRate = (plan: OfferPlan): number => {
+  const { num, den } = displayPrice(plan.price, {
+    baseDecimals: plan.market.base_asset.decimals,
+    quoteDecimals: plan.market.quote_asset.decimals,
+  })
+  const rate = plan.give === 'base' ? Number(num) / Number(den) : Number(den) / Number(num)
+  return Number.isFinite(rate) && rate > 0 ? rate : 0
 }
 
 export type PlanError = 'insufficient-balance' | 'side-disabled' | 'below-min' | 'above-max' | 'below-dust'
--- a/src/lib/swapDisplay.ts
+++ b/src/lib/swapDisplay.ts
@@ -104,8 +104,12 @@ export function swapAmountBeforeFee(tx: Tx): SwapDisplayAmount | undefined {
   return swapAssetDisplayAmount(toAmount + fee, toDecimals, toTicker)
 }
 
-/** Rate the covenant enforced: receive-asset units bought by one unit of the
- * sent asset, recomputed from the stored atomic amounts. */
+/** Rate the swap was priced at, quoted pre-fee like the live composer's Rate
+ * row: the covenant pins the net payout, so gross it back up by the stored
+ * feeBps — the receipt's Swap fees row itemizes that same fee, and a
+ * net-derived rate would count it twice. Restored swaps may lack feeBps;
+ * then the net rate is all we have (matching the net Swap to row shown in
+ * that case). */
 export function swapPriceRateLabel(tx: Tx): string | undefined {
   const swap = tx.assetSwap
   if (!swap) return undefined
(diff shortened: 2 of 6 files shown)


tiero:
The **E2E (core / Mobile Chrome)** failure on `525aff1` is an infrastructure flake, not related to this diff: the job died during setup while downloading Chrome (`curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR` fetching `google-chrome-stable_current_amd64.deb` from dl.google.com) — no tests ran, no containers started ([job log](https://github.com/arkade-os/wallet/actions/runs/30476573548/job/90659669209)).

I tried to re-run the failed job but the integration doesn't have workflow re-run permission (403). Could someone with access hit "Re-run failed jobs" on the run? I'd rather not push an empty commit to retrigger, to keep this PR's history clean.

---

tiero:
Second failure on the same run, same story: **E2E (assets-send / Mobile Chrome)** failed in `should send usds (some and max) to ark address` — the Send form's `$1.19 available` fiat label didn't render within its 2.1s timeout, on the first attempt and all 5 retries ([job log](https://github.com/arkade-os/wallet/actions/runs/30476573548/job/90659669160)). Not caused by this diff:

- The PR only touches the swap Rate row, the receipt rate label, and swap limit hints — nothing on the Send flow, fiat conversion, or balances (`fundWallet` is a plain offchain receive + reading the main balance).
- CI is green on master at `7aa3b6a`, the exact base commit these two commits sit on.
- This runner had degraded external network across the whole run — the core job couldn't even download Chrome (`curl (92)` from dl.google.com), and the Send form's fiat label depends on the external `blockchain.info/ticker` fetch, which fits the label never settling in 2.1s.

A single "Re-run failed jobs" on [the run](https://github.com/arkade-os/wallet/actions/runs/30476573548) covers both failed jobs (I still can't trigger it — 403 for the integration).

---

PR #861, opened 2026-07-29
add Solvers Settings

bordalix:
## Summary by CodeRabbit

* **New Features**
  * Added a Solvers settings screen to create, edit, and remove local solver cards.
  * Introduced a Solvers navigation entry with a dedicated solver icon.
  * Enabled market discovery refresh that can bypass cached results.
* **Bug Fixes**
  * Invalid or malformed stored solver cards are safely ignored.
  * Solver cards are now deduplicated by both label and target network.
  * Improved user error handling when saving solver cards fails.
* **Tests**
  * Added unit and end-to-end coverage for adding/removing solver cards and validation errors.

### DIFF
--- a/src/lib/storage.ts
+++ b/src/lib/storage.ts
@@ -1,7 +1,7 @@
 import { AssetDetails } from '@arkade-os/sdk'
 import { Config, Wallet } from '../lib/types'
 import { consoleError } from './logs'
-import { LocalCardInput } from '@arkade-os/solver-discovery'
+import { LocalCardInput, validateCard } from '@arkade-os/solver-discovery'
 
 // clear localStorage but persist config (with asset data reset)
 export async function clearStorage(): Promise<void> {
@@ -130,5 +130,17 @@ export const saveSolverCardsToStorage = (cards: LocalCardInput[]): void => {
 }
 
 export const readSolverCardsFromStorage = (): LocalCardInput[] => {
-  return getStorageItem('solverCards', [], (val) => JSON.parse(val))
+  const items = getStorageItem('solverCards', [], (val) => JSON.parse(val))
+  return Array.isArray(items) ? items.filter(isLocalCardInput) : []
+}
+
+const isLocalCardInput = (obj: unknown): obj is LocalCardInput => {
+  const input = obj as LocalCardInput | null
+  return Boolean(
+    input &&
+      typeof input.network === 'string' &&
+      typeof input.label === 'string' &&
+      typeof input.card === 'object' &&
+      validateCard(input.card).ok,
+  )
 }
--- a/src/lib/storage.ts
+++ b/src/lib/storage.ts
@@ -129,6 +129,6 @@ export const saveSolverCardsToStorage = (cards: LocalCardInput[]): void => {
   setStorageItem('solverCards', JSON.stringify(cards))
 }
 
-export const readSolverCardsFromStorage = (): LocalCardInput[] | undefined => {
-  return getStorageItem('solverCards', undefined, (val) => JSON.parse(val))
+export const readSolverCardsFromStorage = (): LocalCardInput[] => {
+  return getStorageItem('solverCards', [], (val) => JSON.parse(val))
 }
--- a/src/lib/storage.ts
+++ b/src/lib/storage.ts
@@ -1,6 +1,7 @@
 import { AssetDetails } from '@arkade-os/sdk'
 import { Config, Wallet } from '../lib/types'
 import { consoleError } from './logs'
+import { LocalCardInput } from '@arkade-os/solver-discovery'
 
 // clear localStorage but persist config (with asset data reset)
 export async function clearStorage(): Promise<void> {
@@ -122,4 +123,12 @@ export const readAssetMetadataFromStorage = (): Map<string, CachedAssetDetails> 
     Object.values(obj).forEach((x) => (x.supply = BigInt(x.supply)))
     return new Map(Object.entries(obj))
   })
+}
+
+export const saveSolverCardsToStorage = (cards: LocalCardInput[]): void => {
+  setStorageItem('solverCards', JSON.stringify(cards))
+}
+
+export const readSolverCardsFromStorage = (): LocalCardInput[] | undefined => {
+  return getStorageItem('solverCards', undefined, (val) => JSON.parse(val))
 }
--- a/src/lib/swap/markets.ts
+++ b/src/lib/swap/markets.ts
@@ -88,7 +88,7 @@ const writeMarketsCache = (network: Network, registry: string, markets: Discover
  * stale cache backstops an unreachable registry (quotes stay live either way).
  */
 export const discoverMarkets = async (network: Network): Promise<DiscoveredMarket[]> => {
-  const localCards = readSolverCardsFromStorage()?.filter((c) => c.network === network) ?? []
+  const localCards = readSolverCardsFromStorage().filter((c) => c.network === network)
   const registry = getSolverRegistryUrl(network)
   if (!registry || !isNetwork(network)) return []
   const cached = readMarketsCache(network, registry)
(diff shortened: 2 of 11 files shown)


src/lib/swap/markets.ts
```diff
@@ -88,11 +88,12 @@ const writeMarketsCache = (network: Network, registry: string, markets: Discover
  * stale cache backstops an unreachable registry (quotes stay live either way).
  */
 export const discoverMarkets = async (network: Network): Promise<DiscoveredMarket[]> => {
+  const localCards = readSolverCardsFromStorage()?.filter((c) => c.network === network) ?? []
   const registry = getSolverRegistryUrl(network)
   if (!registry || !isNetwork(network)) return []
   const cached = readMarketsCache(network, registry)
   if (cached && Date.now() - cached.fetchedAt < MARKETS_CACHE_TTL_MS) return cached.markets
```
tiero: Most pressing item: `localCards` only reaches `discover()` on the fetch path. On a fresh-cache hit this line returns `cached.markets`, so a just-added card produces no markets until the hourly TTL expires. And on the fetch path, `writeMarketsCache(network, registry, markets)` below stores the *merged* result (registry + local) into the registry cache — so a **removed** solver keeps serving quotes for up to an hour.

Suggest merging local cards outside the cache on every call (and caching only `sourceType === 'registry'` entries), so add/remove take effect immediately and the cache stays registry-only.

Separately: `if (!registry || !isNetwork(network)) return []` means a network with *no* registry can't use local cards at all — arguably the main use case for hand-added cards.

---

src/screens/Settings/Solvers.tsx
```diff
@@ -0,0 +1,206 @@
...
+import { readSolverCardsFromStorage, saveSolverCardsToStorage } from '@/lib/storage'
+import FlexRow from '@/components/FlexRow'
+import FlexCol from '@/components/FlexCol'
+import ErrorMessage from '@/components/Error'
+import Shadow from '@/components/Shadow'
+import Modal from '@/components/Modal'
+
+const addSolverCard = (input: LocalCardInput) => {
+  const existingCards = readSolverCardsFromStorage() ?? []
+  const withoutSameCard = existingCards.filter((card) => card.label !== input.label)
```
tiero: This filters by `label` across **all** networks, so adding an "acme" card on bitcoin silently deletes the "acme" card pinned on mutinynet (same in `removeSolverCard`). Keying the filter on `(network, label)` keeps same-named cards independent per network.

---

src/screens/Settings/Solvers.tsx
```diff
@@ -0,0 +1,206 @@
...
+    try {
+      card = JSON.parse(inputValue)
+      const result = validateCard(card)
+      if (!result.ok) throw new Error(`invalid card: ${result.errors.join('; ')}`)
+    } catch (err) {
+      setError(`invalid JSON: ${(err as Error).message}`)
+      return
+    }
+    const input: LocalCardInput = {
+      network: aspInfo.network as Network,
```
tiero: `aspInfo.network` is cast unchecked: while the ASP is still resolving this stores cards under `network: ''`, and on a testnet ASP under a network `discover()` doesn't know — both are silently dead pins that the list still shows. Gating the screen on `isNetwork(aspInfo.network)` (with a "not supported on this network" state) avoids persisting pins that can never produce a market.

---

src/screens/Settings/Swaps.tsx
```diff
@@ -0,0 +1,206 @@
...
+    addSolverCard(input)
+    setLocalCards(readSolverCardsFromStorage() ?? [])
+  }
+
+  const title =
+    localCards && localCards.length > 0
+      ? `You have ${localCards.length} swap card${localCards.length > 1 ? 's' : ''} stored in your wallet.`
+      : 'You have no swap cards stored in your wallet.'
+
+  function Editor({ card, onSave, onCancel }: { card?: Card; onSave?: () => void; onCancel?: () => void }) {
```
tiero: `Editor` is declared inside the component body, so its function identity changes on every render and React **remounts** it — the `setError` re-render resets the uncontrolled textarea back to `defaultValue`. Net effect: paste imperfect JSON, hit Save, and the input is wiped at the same moment the error appears. Hoisting `Editor` to module scope (passing the ref and `clearError` as props) fixes it.

The single shared `editorRef` is also cross-wired if the "+ Add new" editor and a card's Edit editor are open at the same time — the ref points at whichever textarea mounted last, so Save can read the wrong editor's content.

---

src/screens/Settings/Swaps.tsx
```diff
@@ -0,0 +1,206 @@
...
+      setShowEditor((e) => !e)
+    }
+
+    const handleRemove = () => {
+      removeSolverCard({ network: input.network as Network, label: input.label, card: input.card as Card })
+      setLocalCards(readSolverCardsFromStorage() ?? [])
+    }
+
+    const card = input.card as Card
+    const pairs = card.markets.map((m) => m.pair).join(', ')
```
tiero: One malformed stored entry (a hand-edited blob, or an entry missing `markets`) makes this line throw and takes down the whole screen render. Worth shape-filtering entries on read so one bad entry doesn't brick Settings → Swaps for every card.

---

src/lib/storage.ts
```diff
@@ -123,3 +124,11 @@ export const readAssetMetadataFromStorage = (): Map<string, CachedAssetDetails>
     return new Map(Object.entries(obj))
   })
 }
+
+export const saveSolverCardsToStorage = (cards: LocalCardInput[]): void => {
+  setStorageItem('solverCards', JSON.stringify(cards))
```
tiero, line 129: `setStorageItem` throws on quota / private mode and nothing catches it here, so "Save" can throw straight out of the click handler while the UI reports success-by-silence. There's also no size cap on a pasted card, so one large paste can consume the wallet's whole localStorage budget. Suggest `setStorageItemSafely` (surfacing the failure to the screen) plus a per-card byte cap and a max card count.

---

tiero (commented):
Really like the instincts here: storing the SDK's own `LocalCardInput` shape and folding `localCards` into the one existing `discover()` call keeps dedupe and fee-ranking SDK-native with no hand-rolled merge — that's the right layer for this feature. The edit-in-place editor and the remove confirmation dialog are also nice touches the parallel implementation (#859) doesn't have.

That said, a few pressing items, anchored inline — the first one effectively keeps the feature from working:

1. **Cache interaction** (markets.ts): local cards are skipped on the fresh-cache path and leak into the registry cache on the fetch path — adds don't show up and removes keep quoting until the hourly TTL expires.
2. **No re-discovery after add/remove**: `AssetSwapsProvider` only runs discovery on mount, network change, and tab return, so even with the cache fixed a new card produces no markets until one of those happens. A `refreshMarkets()` on the context (as in #859) closes that loop.
3. **Cross-network name clobber** in add/remove (inline).
4. **Unchecked network cast** → dead pins under `''`/testnet (inline).
5. **Editor remount wipes input on validation error** + shared ref cross-wiring (inline).
6. **Unguarded storage write, no caps, screen crash on a malformed entry** (inline).

Also worth a few tests around storage/dedupe/cache behavior — `markets.test.ts` already has the fetch-mock + cache-key harness for it. FWIW, #859 (and the combined branch on #860) has a tested lib layer covering exactly these edges (cache bypass, caps, corrupt-blob recovery, network gating); combining that layer with this PR's editor UI might be the fastest path to landing the feature.

---

src/lib/types.ts
```diff
@@ -81,6 +81,7 @@ export enum SettingsOptions {
   Reset = 'reset wallet',
   Server = 'server',
   Support = 'support',
+  Swaps = 'swaps',
```
tiero: Naming: consider calling this entry **Solvers** rather than Swaps. The screen manages solver cards specifically — not swap behavior in general — so "Solvers" says what you'll find there, keeps "Swaps" free for actual swap settings later, and matches the naming in #859, which makes converging the two implementations easier.

---

tiero:
Let call it Solvers, not Swaps please

bordalix:
@arkana-ai-bot full review

PR #877, opened 2026-08-04
Disable server config on mainnet

gringokiwi:
Don't allow setting operator URL on mainnet + small refactor to extract the existing `isMainnet` helper

Screenshot 2026-08-04 at 12 01 46 AM

## Summary by CodeRabbit

* **New Features**
  * Advanced settings now hide the Server option when connected to a mainnet network.
  * Server settings now display a warning and prevent changes when connected to a reachable mainnet server.

* **Bug Fixes**
  * Network detection is now applied consistently across wallet initialization and settings.
  * Server configuration remains available when the mainnet server is unreachable.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -36,6 +36,9 @@ export const defaultArkServer = () => {
   return mainServer
 }
 
+export const isMainnet = (network: NetworkName | string): boolean =>
+  !!network && network !== 'testnet' && network !== 'mutinynet' && network !== 'signet' && network !== 'regtest'
+
 const DELEGATE_URL: Record<NetworkName, string | null> = {
   bitcoin: 'https://delegate.arkade.money',
   mutinynet: `https://delegator.mutinynet.arkade.sh`,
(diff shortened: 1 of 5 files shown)


gringokiwi:
Addressed the review feedback and rebased onto master to resolve the conflict:

- **Fixed the empty-network bug** (flagged by both review passes): `isMainnet('')` returned `true`, so the Server option in Advanced settings — and, more importantly, the mnemonic/nostr key derivation in `wallet.tsx` — treated an unreachable ASP or the pre-load state as mainnet. `isMainnet` now returns `false` for an empty network.
- **Guarded the Server screen itself**, not just the Advanced menu row, per CodeRabbit's "enforce at the connection boundary" comment — navigating to `Server` directly now shows a blocked state on mainnet (with the unreachable-ASP recovery path preserved), instead of relying solely on the menu filter.
- **Added a unit test** for `isMainnet` (bitcoin/future networks, testnet/mutinynet/signet/regtest, and the empty-string case) per the CodeRabbit nitpick.
- Rebased onto current master to resolve the merge conflict; lint, format check, and the full unit suite (634 tests) pass, and `pnpm build` compiles clean.

PR #885, opened 2026-08-07
Show both legs of a Lightning send on its receipt

tiero:
Stacked on #876 — base is that PR's head.

The Lightning-send receipt showed a single **Transaction ID**. It now shows the pair an asset swap shows:

- **Funded** — the tx that funded the lockup covenant
- **Completed** / **Refunded** — the tx that spent it

## Why this needed more than a UI edit

A Lightning send is two transactions and the wallet only signs the first. The second — the solver's claim after it pays the invoice, or the covenant refund — is not a transaction of ours unless it *is* the refund, so there was nothing to render. Three pieces:

1. `requestLnSend` carries the covenant's `swapPkScript` through. The swap client already returned it (*"for watching the lockup and its spend"*); `lnSwap.ts` was dropping it.
2. `Send/Details.tsx` records it against the funding txid at funding time — the only moment it exists, since the quote is gone once that screen unmounts.
3. `useLnSendReceipt` resolves the spender from the indexer when the receipt is opened, persists the answer, and seeds from that store on later visits — which also matters because a swept covenant stops being answerable at all.

## Telling a claim from a refund

The refund path pays the refund address the client handed the solver, so a refund is a transaction of the wallet's own and a claim never can be.

The failure mode that matters is reporting returned money as a paid invoice. Two rules narrow it:

- **Ordering.** The ownership check runs *after* the indexer reports the spend, never from a copy taken before it.
- **Failing loudly.** The check must fail rather than answer "not ours" when it cannot tell. This is why it uses the raw SDK history read rather than `getTxHistory`, which reports failure as an empty history — that would have read as "no refund of ours" and persisted a refunded swap as completed. A lookup that cannot answer leaves the swap unresolved and retries on the next visit.

Both are pinned by tests.

**What that still does not buy**, stated plainly: the wallet's own view and the indexer are separate stores, so asking second gives sequencing, not currency. A wallet a sync cycle behind can miss a refund the indexer already knows about, and that answer persists. The two rules remove the avoidable half — a stale snapshot, and a failure disguised as a verdict — and the residual is the same trade-off `assetSwaps` already accepts for the same question. It also narrows as the refund ages, which is when a receipt is usually read.

Asking the solver instead was the alternative, and was rejected: `RfqStatus.profile` is `Record<string, unknown>` with no guaranteed claim txid, and it would make the receipt depend on a third party still being reachable and still remembering the negotiation.

## Two deliberate deviations

- **"Refunded", not the asset swap's "Cancelled"** — nobody cancelled anything, and no cancel action exists on this path.
- **The Status row still reads "Settled"**, referring to the Ark funding tx. Next to this branch's "sent means paid" stance that is arguably worth revisiting, but changing status semantics is beyond showing the two txids, so it is left alone here.

## Known scope limits

Lightning sends are tracked on `TransactionActivityMetadata` and resolved lazily on the receipt, rather than getting the dedicated store + monitor + chain-restore that asset swaps have. That is deliberate for a two-row receipt change with one consumer and no user action to perform, but it means: no activity-list merge of the send/refund pair, no restore after wallet recovery, and the covenant handle lives in a 250-entry LRU. When the M3 receive leg lands, that is the thing to build rather than extend.

## Verification

629 tests across 75 files, plus `tsc --noEmit`, eslint and prettier clean. `vite build` is covered by the Cloudflare Pages checks on this PR.

Note that `ci.yml` and `playwright.yml` are filtered to PRs targeting `master`/`next`, so neither runs here — the above was verified locally.

Reproducing locally needs one workaround: `pnpm install` cannot fetch `@arkade-os/swap` from a restricted network (`codeload.github.com` 403s), so it was built from ts-sdk at the pinned commit and placed in `node_modules` by hand. None of that scaffolding is committed.

### DIFF
--- a/src/components/Details.tsx
+++ b/src/components/Details.tsx
@@ -177,14 +177,7 @@ export default function Details({ details, variant }: { details?: DetailsProps; 
     ['Destination', destination, <TypeIcon key='destination-icon' />],
     ['Funded', fundedTxid, <HashIcon key='funded-icon' />, offchainTxOnClick(fundedTxid)],
     [spendLabel ?? 'Completed', spendTxid, <HashIcon key='spend-icon' />, offchainTxOnClick(spendTxid)],
-    // A two-leg receipt names its funding tx on the Funded row, so a
-    // Transaction ID row would only repeat it under a vaguer label.
-    [
-      'Transaction ID',
-      fundedTxid ? undefined : txid,
-      <HashIcon key='txid-icon' />,
-      showTxidLink ? txidOnClick : undefined,
-    ],
+    ['Transaction ID', txid, <HashIcon key='txid-icon' />, showTxidLink ? txidOnClick : undefined],
     ...assetIdRows,
     ['Direction', direction, <DirectionIcon key='direction-icon' />],
     ['Type', type, <TypeIcon key='type-icon' />],
--- a/src/components/Details.tsx
+++ b/src/components/Details.tsx
@@ -177,7 +177,14 @@ export default function Details({ details, variant }: { details?: DetailsProps; 
     ['Destination', destination, <TypeIcon key='destination-icon' />],
     ['Funded', fundedTxid, <HashIcon key='funded-icon' />, offchainTxOnClick(fundedTxid)],
     [spendLabel ?? 'Completed', spendTxid, <HashIcon key='spend-icon' />, offchainTxOnClick(spendTxid)],
-    ['Transaction ID', txid, <HashIcon key='txid-icon' />, showTxidLink ? txidOnClick : undefined],
+    // A two-leg receipt names its funding tx on the Funded row, so a
+    // Transaction ID row would only repeat it under a vaguer label.
+    [
+      'Transaction ID',
+      fundedTxid ? undefined : txid,
+      <HashIcon key='txid-icon' />,
+      showTxidLink ? txidOnClick : undefined,
+    ],
     ...assetIdRows,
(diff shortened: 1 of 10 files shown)


tiero:
Thanks — acted on one of the two, skipped the other deliberately.

**Sync-lag subcase → fixed the claim, not the code (a925dbd).** You're right that the ordering argument was doing work it can't do. "A lookup issued after the indexer sees at least as much as it did" is a sequencing claim wearing a causal one's clothes: the wallet's own view and the indexer are separate stores, so asking second buys ordering, not currency. The comment on `lnSendSpender` and the PR body now say what the two rules actually remove — a stale snapshot, and a failure disguised as a verdict — and name the residual instead of implying it was designed away. No behaviour change; the trade-off is the same one `assetSwaps` accepts for the same question.

**Migration note → skipped.** The old `{ swapPkScript, spentTxid, outcome }` shape only ever existed on 676d257, which has never been merged or released — the only entries with it come from this PR's own preview deploys. `saveTransactionActivityMetadata` replaces `lnSend` wholesale rather than deep-merging, so a re-resolve overwrites cleanly and the stale top-level keys are unreferenced dead data in the meantime. Documenting a shape that existed for one commit of an unmerged branch seemed likelier to confuse a future reader than help them, so I left the type alone. Happy to add it if you'd rather have the breadcrumb.

Also confirming your read on `paidUs` using `[key.arkTxid, key.boardingTxid, key.commitmentTxid]`: that rename to the raw SDK key shape is the point of the change, not incidental — it is what makes the read throw instead of degrading to `[]`.

---

PR #892, opened 2026-08-11
refactor(rfq): adopt the SDK's Nostr transport, on the ephemeral kind

Kukks:
Adopts the Nostr RFQ transport from `@arkade-os/swap` and deletes the wallet's local copy, which also carries the ephemeral-kind fix this PR originally opened for.

## Why it started

[arkade-os/lightning-swap-service#51](https://github.com/arkade-os/lightning-swap-service/pull/51) is **merged**, moving RFQ directed traffic from kind `4859` to `24859`. The two ends subscribe by `kinds`, so a solver on `24859` and a maker on `4859` do not see each other — and **neither can report it**. No error, no refusal. Every request times out appearing to blame the solver.

## What it does now

[arkade-os/ts-sdk#718](https://github.com/arkade-os/ts-sdk/pull/718) merged the transport into `@arkade-os/swap`, beside `httpTransport` and `relayTransport` where `rfq.ts` always said it belonged. So rather than patching the constant in a duplicated file, this deletes the duplication:

- `src/lib/nostrRfq.ts` — **deleted** (386 lines)
- both send screens and `scripts/probe-rfq-solver.mjs` import from `@arkade-os/swap/nostr`
- the kind constant now has exactly one definition on this side of the wire

`nostr-tools` stays a direct dependency here and is an **optional peer** of the package, so the wallet's dependency graph is unchanged — the subpath is what stops consumers who don't use Nostr from resolving it.

## The SDK moved with it

The packed `@arkade-os/swap` declares `@arkade-os/sdk: 0.4.58`, and `pnpm.overrides` is what redirects that to the vendored tarball — so both had to move together:

```
vendor/arkade-os-sdk-0.4.57-pr691-752ed57.tgz  ->  arkade-os-sdk-0.4.58-pr718-975ed0f.tgz
vendor/arkade-os-swap-0.0.1-pr691-752ed57.tgz  ->  arkade-os-swap-0.0.1-pr718-975ed0f.tgz
```

That is the wallet's entire SDK surface, not just RFQ. **It appears to have been an improvement rather than a risk:**

| | test files | tests executing |
| --- | --- | --- |
| before | 33 failing to collect, 42 passing | 425 |
| after | **75 passing, 0 failing** | **619** |

Roughly 194 tests that were silently not running now run and pass.

**Caveat, stated honestly:** the cause is not isolated. The re-vendor bumped the SDK *and* regenerated `node_modules` in one step, so this may be 0.4.58 fixing a resolution failure, or simply a clean install. Either way nothing regressed and a third of the suite stopped being dark.

## On the tests

The transport's behaviour — quotes, refusals, reply correlation, relay loss, timeouts — is covered by the package's own suite now. Duplicating it here would only test a vendored copy of already-tested code.

What replaces it is narrower and is genuinely this repo's problem: the wallet pins the package as a **vendored tarball**, so a re-vendor that packs a stale build, drops the `./nostr` subpath from `exports`, or lands without a resolvable `nostr-tools` breaks the import at runtime while every upstream test stays green. Four assertions cover exactly that, including one that the vendored tarball carries a *post-ephemeral* build — a stale vendor would still export a working transport, on `4859`, and simply never see the solver.

## Test plan

- Full suite — **75 files / 619 passed / 0 failed**
- `pnpm lint` — clean
- `pnpm build` — clean (18.1s)
- Prettier clean on all four changed sources. `pnpm format:check` fails repo-wide (467 files) on a Windows checkout due to CRLF; pre-existing and unrelated.
- Verified out-of-band that `@arkade-os/swap/nostr` resolves from the vendored tarball and exports `RFQ_AD_KIND, RFQ_DIRECTED_KIND, RelayUnavailable, nostrRfqTransport`, and that the packed `dist/index.js` still has **zero** nostr-tools references — the optional-peer isolation survives packing.

## Not included

The **solver is deliberately untouched**. Its `src/relay/nostr.ts` is a `WireCodec` behind a separate `RelayConnection` — framing only, bidirectional, handling the broadcast kind and verifying signatures on every inbound event. That is a different abstraction from a maker-side request/reply transport, and importing the SDK subpath there would pull `@arkade-os/sdk` and the whole maker swap stack into a server whose transport file currently depends on no arkade package at all.


Kukks:
**Correction to this PR's description.**

The description claims the SDK bump took the suite from "33 test files failing to collect" to all 75 passing, and credits `@arkade-os/sdk` 0.4.58. **That was wrong**, and I'd rather correct it than leave it as a misleading data point for the next person who bumps the SDK.

The 33 collection failures were:

```
Failed to resolve import "../_gitCommit" from "src/components/LoadingLogo.tsx"
```

`src/_gitCommit.ts` is a **generated** file — `pnpm build` produces it via `pnpm git-info`. In the worktree used for this PR I had run `pnpm build` before `pnpm test`, so the file existed and those suites collected. In a fresh worktree they fail regardless of SDK version.

Isolated afterwards while rebasing #889: a fresh worktree showed 34 failing files, and running `pnpm git-info` **alone** — no dependency change at all — took it to 1. The SDK bump neither caused nor fixed it.

The description did flag the attribution as unisolated, which was the right instinct, but it should not have presented the number as evidence at all. Everything else in the PR stands: the transport swap, the vendored tarballs, and the suite being green.

PR #908, opened 2026-08-12
Lightning over Arkade RFQ; remove Boltz

pietro909:
Lightning payments now run over the Arkade RFQ protocol against the released `@arkade-os/swap@0.0.2` + `@arkade-os/sdk@0.4.60`, and the Boltz integration is removed end to end.

## How a Lightning send works now

1. The solver comes from **discovery, not configuration**: the bundled `beta-solver` card (pubkey + Nostr relay + card bounds, currently 500–1,000 sats) plus any user-stored cards, both visible under Settings → Solvers.
2. The wallet negotiates a sealed quote over `wss://nostr.arkade.sh` (the package's Nostr transport, ephemeral kind 24859), derives the lockup covenant **locally**, and refuses on any mismatch — funding the derived address is the acceptance.
3. **Success is reported at the funding txid** ("Payment is on the way"); the solver pays the invoice and claims with the preimage in the background. Receipts show both legs (funding + completion/refund) and the terminal status.

The covenant co-signer key is pinned per network (matching the SDK's own pin), never fetched from an endpoint at swap time. Lightning-address (LNURL) sends stay; a receive flow ships in code but the preview solver disables that corridor for now.

## Verified

- **Two live mainnet settlements** to a Lightning address, one per wire generation during the solver migration — both 0-fee, receipts **Settled**.
- 655 unit tests / 79 files, e2e coverage for the send path, plus a regtest solver setup doc for end-to-end testing.

## Removed with Boltz

Chain swaps, Lightning receive via reverse swaps (and LNURL receive, which depended on them), the Boltz app screens, swap history, and fee-API-driven limits. Old Nostr backups still restore — the loader ignores keys it no longer reads.

## Known gaps

- Lightning receive waits on a solver serving `lightning:BTC → arkade:BTC`.
- `ServiceWorkerWallet` lacks the HD-allocation surface, so swap secrets fall back to random (restart-recovery of the sender key pairs with ts-sdk#727).

## Summary by CodeRabbit

- **New Features**
  - Added direct Lightning send and receive through Arkade solver discovery.
  - Added built-in solver cards alongside user-managed solver cards.
  - Added Lightning transaction receipts showing funding, completion, or refund details.
  - Added improved asset-swap persistence and status updates.
- **Improvements**
  - Lightning invoices now validate expiration and network compatibility.
  - Sensitive Lightning invoices are redacted from error reports.
  - Added TLS support and diagnostics for local Lightning testing.
- **Documentation**
  - Added instructions for running Lightning end-to-end tests with a real regtest solver.


tiero:
### Bug 1: balance counts an unspendable VTXO

The displayed balance includes funds locked in a swap covenant.

`requestLightningSend` registers the lockup as a `vhtlc-v2` contract in the wallet's own ContractManager (`@arkade-os/swap` → `registerLockupContract`, written with `metadata.genericallySpendable = false`). `createOffer` does the same for asset swaps. Once registered, the covenant's VTXOs come back from `getContractsWithVtxos()`, and the SDK's `computeOffchainBalance` adds them to `settled` and therefore to `total`. It excludes them only from `available`. The `vhtlc-v2` handler says so directly: "The gate covers `available` and renewal."

The wallet displays `total`:

- `src/providers/wallet.tsx:466,479`
- `src/hooks/usePortfolioFiat.ts` (`totalSats += balance`)
- `src/screens/Wallet/PortfolioHero.tsx:32`

Assets have the same problem: `setAssetBalances(assets)` uses `assets` (owned) rather than `availableAssets` (selectable).

Effect: after paying a Lightning invoice the balance does not drop. The spent input leaves and a covenant VTXO of the same size arrives in its place.

This is new in this PR. On master, `src/lib/swap/offer.ts` registered no contract and Boltz VHTLCs lived outside the wallet's contract repository, so escrow never reached the balance. `balance = total` is unchanged from master; the registration is what makes it visible.

Suggested fix: display `available` and `availableAssets`, and surface `total - available` separately if in-flight funds should remain visible.

---

tiero:
### Bug 2: transaction history does not account for failed swaps

Four separate causes:

1. **The outcome is never polled.** `awaitLnSendOutcome` (`src/lib/lnSwap.ts:387`) has no production call site — the only reference is `src/test/lib/lnSwap.test.ts`. Nothing asks the solver for the RFQ state, so `{ kind: 'failed', state }` is never produced anywhere in the app.
2. **`rfqId` is discarded at funding time.** `src/screens/Wallet/Send/Details.tsx:158` persists only `{ swapPkScript }`, so the status cannot be queried later either.
3. **There is no Lightning equivalent of `mergeAssetSwapActivity`** (`src/providers/wallet.tsx:189`). Asset swaps are folded into history with real statuses (pending / completed / cancelled / recoverable). A Lightning send is a plain `sent` row, and `src/lib/asp.ts:213` forces `settled: true` on every send. When a refund lands it appears as an unrelated `received` row, with nothing linking the two.
4. **`useLnSendReceipt` only runs on the detail screen** and only labels the second txid "Refunded" or "Completed" (`src/hooks/useLnSendReceipt.ts:102`). The history list never learns the outcome.

Related removals worth noting: `src/components/SwapsList.tsx`, the Nostr swap backup, and `src/test/e2e/restore.test.ts`, which existed to verify that a restored wallet had the correct transaction history and swap history.

`@arkade-os/swap` already provides the missing pieces — `RfqSwapManager` (with `onSwapFailed` / `onSwapUpdate`), `isRfqSwapTerminal`, and `RFQ_SWAP_TERMINAL_STATES`. None of them are imported. `RfqSwapManager` would also cover two related gaps: it pushes the refund when the solver goes silent (`pushRefundWithoutReceiver` / `refundIfUnresolved` are likewise unused), and its `retireContract` stops watching a finished lockup — asset swaps get this via `watchOfferSwaps`, Lightning lockups stay watched permanently.

---

tiero:
### Correction to bug 2, and two additions

Point 3 of my previous comment is wrong in a way that matters. I said a failed Lightning send shows as a plain `sent` row. In most cases **no row appears at all**.

**Mechanism.** History is built in the worker from the whole contract repository — `GET_TRANSACTION_HISTORY` → `getVtxosFromRepo()`, which iterates `manager.getContracts()` and flattens every contract's VTXOs (`chunk-AUEXB42T.js:16955`, `:17707`). `buildTransactionHistory` then treats any output whose txid matches the spend as change:

```js
// chunk-AUEXB42T.js:6655-6669
const changes  = fromOldestVtxo.filter((_) => _.txid === vtxo.arkTxId);
const allSpent = fromOldestVtxo.filter((v) => v.arkTxId === vtxo.arkTxId);
txAmount = spentAmount - changeAmount;
if (txAmount !== 0 || assets) { sent.push(...) }
```

Because the lockup is now a registered contract, its VTXO is in that set. Paying a 900-sat invoice from a 1,000-sat VTXO creates lockup(900) + change(100), both wallet-owned, so `txAmount = 1000 − 1000 = 0` and no SENT row is emitted. With a nonzero intent fee you get a few-sat SENT row instead. The refund is suppressed the same way: the lockup VTXO carries `arkTxId === refundTxid`, so the RECEIVED branch at `:6641` skips it. Money leaves, comes back, history shows nothing.

**Two consequences.**

1. **The receipt never renders.** `src/lib/asp.ts:206` gates `lnSend` on `type === 'SENT'`, and `useLnSendReceipt.ts:96` returns undefined without it. No SENT row means no receipt at all.
2. **A refunded swap is recorded as "Completed."** `paidUs` (`useLnSendReceipt.ts:76-79`) asks `getTransactionHistory()` whether the spender txid is one of ours. The refund nets to zero by the same rule, so it produces no row, `paidUs` returns false, and `lnSwap.ts:356` writes `outcome: 'completed'`. Line 69 then gates the lookup shut, so the wrong verdict is permanent. This is the exact failure the hook's own docstring at `:73-75` says it uses the raw SDK read to avoid.

For (2), `@arkade-os/swap` has the correct primitive: `readLockupFate(indexer, { swapPkScript, paymentHash })` decides `open`/`claimed`/`returned`/`unknown` from the spend witness rather than by inference. It needs `paymentHash`, which is computed at `lnSwap.ts:87` and discarded at `:306`.

**Also worth knowing:** the outcome poll was not an oversight. `abcd7bf` ("`Sent` now means the invoice was paid, not merely funded") wired `awaitLnSendOutcome` into `payLightning` with explicit refund reporting. The squash-merge `3631893` (#876) restored the fund-and-return body and left the function orphaned. `git log -S awaitLnSendOutcome origin/master..HEAD -- src/screens/Wallet/Send/Details.tsx` shows both commits. This was a working behaviour lost to a merge.

One note if it gets re-wired: `awaitLnSendOutcome` gates on `isRfqTerminal`, whose set is `['settled','refused','expired','refunded','stuck']`. Per the package, `refused`, `expired` and `stuck` are terminal for the negotiation but say nothing about whether the sats are still at the lockup. `RFQ_RESOLVED_STATES` (`['settled','refunded']`) is the right gate.

---

PR #913, opened 2026-08-13
fix(assets): cap selectable amounts at spendable, not owned

pietro909:
Addresses the asset half of the balance bug reported in https://github.com/arkade-os/wallet/pull/908#issuecomment-5278434021.

## Root cause

`@arkade-os/swap`'s `registerLockupContract` writes the swap covenant into the wallet's own ContractManager with `metadata.genericallySpendable: false`. That puts the covenant's VTXOs into the wallet's snapshot.

The SDK's `computeOffchainBalance` applies the `genericallySpendable` gate only to `available` and `availableAssets`:

```js
if (vtxo.isPreconfirmed) preconfirmed += vtxo.value; else settled += vtxo.value;   // no gate
if (isGenericallySpendable(vtxo) && isUnlocked(vtxo)) {                            // gate here only
  available += vtxo.value;
  addAssets(spendable, vtxo);
}
```

So an escrowed asset stays in `assets` (owned) while dropping out of `availableAssets`.

## Why this is more than a display bug

Every selectable asset amount read `assetBalances`, i.e. owned:

- `Send/Form.tsx` asset options
- `Swap/Index.tsx` composer max
- `BitcoinDetail.tsx` → `sendAccount.balance`, and `sourceAsset.balance`, which `Send/Form.tsx:817` spends verbatim on "use all"

With an offer in flight the composer offered a maximum coin selection cannot fund. The send failed at spend time rather than being blocked at entry.

## Change

`availableAssets` is threaded through `WalletProvider` as `availableAssetBalances` and used wherever an amount is selectable. `PortfolioRow` gains `spendableBalance` alongside `balance`, and `sourceAsset.balance` now carries the spendable amount since it fulfills the send.

Reporting surfaces keep showing owned — portfolio rows, the fiat total, the asset screens. Escrow is still the wallet's money.

Send-form list membership still follows owned assets, so an asset held entirely in escrow appears with nothing selectable rather than disappearing from the picker.

## Tests

Three regression cases; all 616 unit tests pass.

- `usePortfolioFiat.test.tsx` — owned/spendable divergence propagating to `sourceAsset`, and an asset absent from `availableAssets` reading as entirely unspendable
- `swap.test.tsx` — end to end: an asset held partly in escrow offers only the spendable part through to `createSwap`

I verified the swap test fails against pre-fix code rather than passing vacuously. Existing helpers now mirror owned → spendable by default, so prior tests keep their original meaning.

## Deliberately not in scope

**The BTC home-screen total is still wrong**, for the same root cause. It cannot be fixed cleanly at the display layer today: `total - available` also folds in `recoverable`, `pendingRecovery` and intent-locked funds, and `WalletBalance` has no bucket isolating escrow. Showing `available` instead would erase swept and awaiting-recovery funds from the hero — a regression in the opposite direction. This wants an escrow bucket on `WalletBalance` upstream; `computeOffchainBalance` already has `isGenericallySpendable` in hand.

**`BitcoinDetail.tsx:316`** still gates the Send button on the owned balance, so an account entirely in escrow reaches a form capped at zero rather than a disabled button. Fixing that means deciding what to tell the user, which belongs with the in-flight presentation question above.

### DIFF
--- a/src/hooks/usePortfolioFiat.ts
+++ b/src/hooks/usePortfolioFiat.ts
@@ -14,8 +14,14 @@ export interface PortfolioRow {
   ticker: string
   icon?: string
   decimals: number
-  /** Raw balance in the asset's smallest unit (sats for BTC, minor units otherwise). */
+  /** Raw balance in the asset's smallest unit (sats for BTC, minor units otherwise).
+   * Everything the wallet owns, escrow and awaiting-recovery included — this is
+   * the reporting figure, not a spending limit. */
   balance: number | bigint
+  /** The spendable subset of {@link balance}, same denomination. Funds locked in
+   * a swap covenant, intent-locked or awaiting recovery are excluded, so this is
+   * what any selectable amount must be capped at. */
+  spendableBalance: number | bigint
   /** Fiat value in the user's selected currency (0 if no price feed). */
   fiatAmount: number
   /** Equivalent value in satoshis, computed via the live BTC price feed. */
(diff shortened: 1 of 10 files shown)


tiero:
I restored my same key here, and worked

PR #915, opened 2026-08-13
Improve QR scanner camera error handling and UX

tiero:
## Summary
Enhanced the QR scanner's camera error handling to distinguish between blocked camera permissions and unavailable cameras, providing users with more helpful error messages and recovery options.

## Key Changes
- **New camera permission utilities** (`src/lib/camera.ts`): Added `queryCameraPermission()` to detect the actual camera permission state and `cameraErrorText()` to generate appropriate error messages based on permission status
- **Improved error state management**: Changed error state from boolean to string to display contextual error messages
- **Better user recovery flow**: 
  - Added "Try again" button when camera access fails, allowing users to retry after unblocking permissions
  - Hides video element when error occurs
  - Clears error state on retry attempt
- **Smarter error detection**: Since qr-scanner reports all failures as "Camera not found.", the code now queries the Permissions API to determine if the user blocked access vs. the camera being unavailable
- **Comprehensive test coverage**: Added unit tests for camera permission detection and integration tests for Scanner component error scenarios

## Implementation Details
- The `queryCameraPermission()` function gracefully handles browsers that don't support camera permission queries (like Firefox) by defaulting to 'prompt' status
- Error retry logic uses an `attempt` counter to trigger re-initialization of the QR scanner
- Button layout adapts based on error state: shows "Try again" when error occurs, "Cancel" becomes secondary
- Tests mock the qr-scanner library and camera permissions API to verify error handling paths

https://claude.ai/code/session_01RJPVarPXaWDBNU8paKJazx

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved QR scanner recovery when camera access fails.
  - Added clearer messages for blocked permissions versus unavailable cameras.
  - Prevented unnecessary scanner resets while sending.
  - Added “Try again” and “Cancel” actions when scanning errors occur.
  - Scanner now hides the camera view while displaying an error and can recover successfully after retrying.

- **Tests**
  - Added coverage for camera permissions, error states, retry behavior, and successful recovery.

### DIFF
--- a/src/components/Scanner.tsx
+++ b/src/components/Scanner.tsx
@@ -7,6 +7,7 @@ import Padded from './Padded'
 import { QRCanvas, frameLoop, frontalCamera } from 'qr/dom.js'
 import { useRef, useEffect, useState } from 'react'
 import { extractError } from '../lib/error'
+import { cameraErrorText, queryCameraPermission } from '../lib/camera'
 import QrScanner from 'qr-scanner'
 
 const videoStyle: React.CSSProperties = {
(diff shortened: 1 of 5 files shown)


bordalix:
@tiero there's a weird edge case in Firefox, where we enter a infinite loop:
- Firefox asks for permission twice, at the Android level and at the browser level
- If I say YES at the Android level and NO at the browser level, I enter a infinite loop

See video.

https://github.com/user-attachments/assets/46262d84-c5cc-4855-ac07-0dc8741962a3

Should be fixed by https://github.com/arkade-os/wallet/pull/915/commits/2bd4231e04cb60ed25274ecfcc3fc61cb11bacf7

tiero:
Pushed 668a8cb, addressing the review round.

On the Firefox loop @bordalix caught — the cause was `const Scan = () => <Scanner … />` defined inside `SendForm`'s render body. That creates a new component *type* on every parent render, so React unmounted and remounted the scanner each time, and every remount asks for the camera again. My error path made it visible: a failed start calls `onError`, which sets parent state, which re-renders, which remounts the scanner, which fails again. 2bd4231 is the right fix; I kept it and folded the duplicated props into one shared element (per Copilot's note on `Form.tsx`), with a comment on why it must stay an element and never a component defined during render.

The rest of Copilot's review:

- **`camera.ts` logging noise** — agreed and fixed. `'camera'` being unsupported in `permissions.query()` is routine on Firefox, not something support should have to read past in the persisted log. The fallback is now silent.
- **`handleRetry` state update** — switched to the functional form.
- **Both test files leaking `navigator.permissions`** — captured and restored in `afterEach`.

Unit suite (78 files, 623 tests) and typecheck pass locally.

---

tiero:
CodeRabbit and Arkana both flagged the unmount race, and 42c7404 already carries the `cancelled` guard — I've added the regression test it asked for in 81f162e: the scanner unmounts while the permission lookup is still pending, then the answer arrives. Verified it earns its keep — with `if (cancelled) return` removed it fails, with the guard it passes. Full unit suite 78 files / 624 tests green.

The race was real and worth closing: `onError` fires into the send form, so a late answer painted "Camera access is blocked…" on a screen the user had already left.

On the two other notes from Arkana's review, both deliberate:

- **`handleRetry` → `onError('')` → `smartSetError('')` can repaint the ASP-unreachable banner.** Agreed it's benign — the form is behind the scanner overlay at that moment, and clearing our stale camera error is worth more than the edge case of an unrelated banner returning a moment early.
- **`ScannerMills` parity** — untouched on purpose. It's the fallback behind the header's `M` switch, not the default path, and giving it the same treatment means duplicating the permission handling into a second library's error shape. Happy to do it in a follow-up if we want the two to converge.

---

PR #918, opened 2026-08-14
Lightning receive: drive the claim off the screen, then persist it

pietro909:
Implements `plans/arkade-swap-receive-stage-1.md` and Phases A, B and E of `plans/rfq-persistence-and-activity.md`.

**Scope grew after the first review.** This PR opened as Stage 1 alone — moving the claim off the Receive screen's `useEffect` and onto `RfqSwapManager`. Stage 2's persistence work was to be a separate PR stacked on top; it now lands here instead, because the branch already carried it and splitting would have meant re-reviewing the same provider twice. Reviewers who read the earlier description should re-read: roughly two thirds of this is new.

Depends on [arkade-os/ts-sdk#773](https://github.com/arkade-os/ts-sdk/pull/773), released as `@arkade-os/swap` 0.0.8. Stacked below [#927](https://github.com/arkade-os/wallet/pull/927), which restacks onto this branch and closes the Activity API's last blocked phase.

## What it buys

A Lightning receive survives navigation, screen unmount, a failed claim attempt — **and now a reload, a closed tab, and a crash.** The claim used to live in a screen effect, so leaving the screen lost the payment and one throw was terminal. Then it lived on a manager owned by a provider, which turned that into "keep the tab open". Wiring the manager's repository turns it into "open the wallet again".

**What is not yet proven:** the reload-survival claim is verified by unit tests against a stubbed manager, not by a real receive. The bundled solver card disables the corridor (`min/max_base_amount: "0"`), so confirmation needs a Mutinynet preview deploy — negotiate, reload mid-flight, watch it claim. Please don't describe the corridor as reload-safe until that's done.

## Stage 1 — the claim leaves the screen

**`src/lib/lnReceive.ts`** — `requestLnReceive` stops recomputing `expectedAmount` and `invoiceExpiresAt`; the package returns both and a second copy is a second place to drift. The `decodeInvoice` gate stays — not a duplicate of either. Forwards the configured co-signer key as the package's compressed `emulatorPubkey` override; without it a local stack derives from the SDK's placeholder pin and `verifyLockupAddress` refuses at quote time. New pure `toReceiveSwap`. `claimLnReceive` → `claimReceive`: callback-shaped, no wait of its own, no third copy of the value gate.

**New `src/providers/lnReceive.tsx`** — one manager for the wallet's lifetime. Page-side rather than in the service worker, on the `watchOfferSwaps`/`AssetSwapsProvider` precedent: the worker hosts `MessageBus` and the wallet reaches this side as a proxy, so moving the manager in would mean a second wallet inside it.

**`src/screens/Wallet/Receive/QrCode.tsx`** — tracks each negotiation before the invoice is shown; renders `refunded` as the loss it is rather than leaving the user on a QR forever; clears a stale negotiation when the amount changes; treats `LockupRegistrationFailed` as retryable, since the failed attempt is inert and the fix is to call again rather than resume.

**`src/providers/flow.tsx`** — `pendingLnReceive` shrinks to what the screen renders. Leaving the whole request in flow state invites a second claim path written against it.

## Phase A — the repository becomes the sink

**The manager owns persistence; the wallet writes no record format of its own.** `repository: assetSwapRepository` is wired into the manager's deps, and the `saveSwap` callback is **deleted rather than demoted** — with a repository wired it is a *second* sink, so keeping it would double-write every record. The wallet's remaining part is building an `RfqSwapOrigin` once and handing it to `addSwap`.

- **`toReceiveOrigin`** (`src/lib/lnReceive.ts`) — the pure request-time → origin mapping. Its profile is written with `rfqSecretsProfile`, never by hand: copying `signingDescriptor` and `preimageHex` across drops `preimageSaltHex`, and a static wallet's swap is then unclaimable with nothing to say so until claim time.
- **The `tracked` side map is gone.** `payoutAddress` and the secrets both live on the stored profile. What replaced it is a `Set<rfqId>` holding ids and nothing else, because `addSwap` still *replaces* a monitored swap and a second `track` for one in flight would reset it to `pending` and un-say a claim already sent.
- **`claimLockup` reads the store** — `getRfqSwap` + `rfqClaimSecretOf`, so a swap restored three reloads later claims by the same path as one negotiated a second ago.
- **Boot restore** — `restoreFromRepository()` before `start()`. One unrebuildable record reports its error and never strands the others; a restore that fails outright still lets `start()` run for the swaps this session negotiates.

### Only one tab may drive — please review this closely

Persistence *creates* a cross-tab hazard that did not exist before. With records in shared IndexedDB every open tab would restore every swap and race the others to claim it. The manager's guards (`inProgress`, `claimedOutpoints`, `dirty`) are private instance fields — they coordinate one manager with itself and say nothing about a second one. Two tabs means two `pushClaim`s over the same VTXOs: one lands, the other fails as a double-spend, and both write records that disagree about `claimArkTxid`. A record whose `claimArkTxid` is the losing tab's is *worse* than a missing one, because `claimIfFunded` reads `partiallyClaimed` off exactly that field.

The manager's driving lifetime is therefore wrapped in a Web Lock. Two details that are easy to get wrong and are the reason to look:

1. **The lock is released by returning from the callback, never by aborting.** Per spec an `AbortSignal` drops a lock request only while it is still *pending*; once granted it has no effect. An abort-only cleanup would hold the lock forever and queue the next mount behind it — and this is a live path, not a page-close curiosity: `setSvcWallet` re-runs the effect in-page on reinit and unlock, and `StrictMode` would double-mount it. The callback is held open on a promise that cleanup resolves.
2. **A passive tab constructs no manager and refuses `track` with a named error.** `LnReceiveHeldElsewhere`, rendered as *"Another tab is handling Lightning receives — close it to receive here"*. The alternatives were a misleading "Lightning unavailable" or an invoice that silently never appears.

Treat any change that moves the manager's construction out of the lock callback as a correctness change, not a refactor.

## Phase B — one stub deleted, one kept

`claimOnchain` is gone: the two claims are optional at installation as of ts-sdk#770, a receive leg never reaches that one, and the manager refuses a kind whose claim is missing at runtime — a better answer than a stub that throws a string.

**`refundArkade` stays.** The plan said to delete both; the type disagrees. `AvailableRfqSwapManagerCallbacks` is `Omit<…, 'claimOnchain' | 'claimLockup' | 'saveSwap'> & Partial<Pick<…>>` — only the two *claims* and `saveSwap` were made optional. A receive leg genuinely has no trader refund, so the throw is the true answer rather than a placeholder, and the comment now says so.

## Phase E — the covenant's own inputs

`LnReceiveRequest` no longer retains the whole `RfqQuote` for one field, nor the `payoutPubkey` nothing read. Both are on `treeParams`, and the difference is provenance: the quote is the *solver's* document — its `refund_locktime` is what the solver says its deadline is, and its `profile.payment_hash` is the solver echoing back a hash the wallet generated. `treeParams` is what the wallet fed into the derivation the lockup address commits to.

That collapsed a signature too: `toReceiveSwap` and `toReceiveOrigin` both read `paymentHash` off `treeParams`, so neither takes it as a parameter and the provider no longer computes `hex.encode(request.secrets.paymentHash)` to thread it through. There is now no way to hand the two mappings different hashes.

## Risks worth naming

- **The `DB_VERSION` 2 bump is one-way.** A browser at 2 cannot be served a bundle pinning an older `@arkade-os/swap`: the open fails `VersionError` across the *whole* database — asset swaps and markets cache included, not just the RFQ half. Rolling back this release means rolling back the data, not the bundle.
- **No backfill.** Receives negotiated by the previous build were never persisted, so this recovers none of them. Anything in flight across the deploy is still lost on reload.
- **A lost contract row is a lost claim.** `rebuildRfqSwap` takes the covenant from the contract manager, so a record whose row is gone cannot be restored by any means. That is the intended trade — one source for the covenant — but it makes the contract store load-bearing for money, and a partial loss is silent until boot.
- **Terminal records hold preimage material for 30 days**, only on the arms where `mustPersistPreimage` forced `preimageHex` into the profile, and only for swaps that settled, where `P` is already public on-chain. No new exposure, but it is the first time the wallet keeps preimage material at rest past a swap's end.

## Testing

`pnpm test:unit` — 83 files, **677 passed, 1 skipped**, against the published `@arkade-os/swap` 0.0.8.

New coverage beyond Stage 1's: `toReceiveOrigin`'s mapping across all three provisioning arms, asserted through the real write-then-read path (`rfqSecretsProfile` → record → `rfqClaimSecretOf`) rather than a hand-built projection, which is the shape that lost `preimageSaltHex` once; `toReceiveSwap` timing the claim off `treeParams` with the quote seeded to disagree; the repository wired as the only sink with no `saveSwap` installed; `claimLockup` recovering payout and secrets from a stored record nothing in the session tracked; boot restore surfacing `failed` and starting anyway; and the two Lightning failure copies on the receive screen.

**Three properties are mutation-checked rather than merely asserted**, because a lock test that passes against a broken lock is worse than no test. Each was confirmed to fail before being reverted:

| Mutation | Test that catches it |
|---|---|
| cleanup aborts without releasing | lock release on effect teardown (fails as a timeout, not a hang) |
| `navigator.locks` bypassed | only one tab drives |
| held-elsewhere branch removed | the receive screen names the other tab |

No e2e. The bundled solver card disables the corridor, so it needs the preview deploy described above.

## Summary by CodeRabbit

* **New Features**
  * Lightning send and receive swaps now persist across reloads and coordinate safely across browser tabs.
  * Receive payments offer clearer status messages and retry options for recoverable failures.
  * Transaction history now groups Lightning and asset swaps with improved labels, outcomes, and metadata.
  * Wallet activity and payment receipts update more reliably after completed transactions.
  * Added support for configured emulator public-key overrides on supported networks.

* **Bug Fixes**
  * Improved restoration, refund handling, and error recovery for interrupted swaps.
  * Prevented stale invoices and outdated wallet balances from appearing after changes or completed payments.


pietro909:
### P0 — `pushClaim` enforces the gate

`packages/swap/src/claim.ts`:

```ts
const locked = input.vtxos.reduce((sum, vtxo) => sum + vtxo.value, 0)
assertFiniteAmount(locked, "lockup_malformed", "the lockup's summed value")
if (!input.partiallyClaimed) {
  assertFiniteAmount(input.expectedAmount, "invalid_gate_input", "expectedAmount")
  if (locked < input.expectedAmount) {
    throw new LockupAmountMismatchError(input.expectedAmount, locked)
  }
}
```

Strictly stronger than the local check it replaced:

- Throws before `signAndSubmitOffchainTx`; `P` only reaches the server at submit, so nothing leaks.
- `expectedAmount` is required, not optional — the guard can't be skipped by old records.
- `assertFiniteAmount` closes a bypass the local check never had: `NaN`/`undefined` fail every comparison, so an unchecked value deletes the gate rather than failing it. Covered by an `it.each`.
- Swept outputs are refused first, so the sum is over live outputs only.

Test: `claim.test.ts` — "refuses a lockup funded below the agreed amount, with nothing signed or submitted", asserting `ark.submitted` is empty. Only bypass is `partiallyClaimed: true`, where `P` is already public.

Delegation stands; no code change.

### P1 — fixed in b6ba80f6

Rollback in a catch, keeping the set-before-`addSwap` ordering (the manager polls immediately, so the secrets must land first).

One amendment: impact isn't quite nil. The orphan also sits in `tracked.current`, which the idempotency guard keys on — a retry reusing the rfqId would early-return and monitor nothing. Regression test asserts the second `track()` reaches `addSwap` again; verified failing without the fix.

Also rebased onto master (swap 0.0.5 → 0.0.7, sdk → 0.4.64). Typecheck clean, 665 passed / 1 skipped.

pietro909:
Stack: this PR ← [#927](https://github.com/arkade-os/wallet/pull/927) (rebased onto this branch; Activity Phase 4 lands there).

Upstream dependency [arkade-os/ts-sdk#773](https://github.com/arkade-os/ts-sdk/pull/773) is released — the branch now builds against published `@arkade-os/swap` 0.0.8 and `@arkade-os/sdk` 0.4.65, no local link.

PR #919, opened 2026-08-14
Seed a two-sided standing book on regtest

tiero:
Groundwork for **Standing** — a peer-to-peer order book built on the offer covenant, with no solver in the loop.

## Why

Standing offers are already the wallet's swap primitive, but the UI hides it. `createOffer` derives a covenant whose `fulfill` leaf is signed by the server alone yet constrained to pay output 0 at least `wantAmount` to the maker's script, and whose `cancel` leaf is a 2-of-2 of maker+server needing **no counterparty signature**. Neither leaf carries a timelock. That is a self-custodial limit order that rests until it's filled or pulled.

A set of those on both sides of a pair is a book. Nothing renders it yet, and nothing can be built against it while the chain holds no orders — no solver rests two-sided inventory here, so the book has to be seeded by a user. This is that user.

## What this adds

`scripts/seed-book.mjs` issues a 6dp `SEED` asset and funds N covenants per side around a mid price:

- **asks** deposit the asset and want sats (`offerAsset`, want-btc program)
- **bids** deposit sats and want the asset (`wantAsset`, want-asset program)

Both sides carry the same asset size per rung, so the seeded position is **flat in the asset** — a full fill changes only the spread, in sats. That's what makes it a book rather than a directional bet.

Each rung is a real offer funded with the type `0x03` packet embedded, so these are indistinguishable from orders any wallet posts. The reader under development will discover them off the global `/v1/txs` stream exactly as it will discover anyone else's.

```
$ node scripts/seed-book.mjs --dry-run

  SEED/BTC (planned)                sats per SEED
    ask   100 SEED        51.5
    ask   100 SEED        51
    ask   100 SEED        50.5
    ————— spread 1.00 —————
    bid   100 SEED        49.5
    bid   100 SEED        49
    bid   100 SEED        48.5

nets out: asset net 0, sats edge if fully filled +600
```

## For the reviewer

- **`--dry-run` is the test.** It plans and prints the book without resolving the SDK or touching the network, and asserts the four properties that make it a book: equal asset per side, ask side taking in more sats than the bid side pays out, best ask above best bid, and no rung at or below dust. I verified the guards fire rather than just pass — `SPREAD_BPS=0` trips the sats-edge assertion, a tiny rung trips the dust check. The SDK imports are deliberately dynamic so the pure path stays network-free.
- **The covenant is keyed on the receive side, not the deposit.** Asks use `offerAsset`, bids use `wantAsset`. Inverting this binds an asset amount as a sat want, which a taker could fill for dust. Commented at the call site because it's the one easy-to-invert line in the loop.
- **Rounding goes against the maker in both directions** — asks up, bids down — so a rounded rung is never quoted better than intended. The covenant honours whatever integer lands there.
- **Liveness is checked by funding txid, not count.** Identical offers derive identical addresses, so a stale deposit at the same script would otherwise pass the wait.
- Rungs are written to `scripts/.seeded-book.json` (`offerHex` + funding txid, all a cancel needs), now gitignored.
- Nothing under `src/` changes; no runtime behaviour is affected.

## Testing

```bash
node scripts/seed-book.mjs --dry-run        # pure, no stack needed
git submodule update --init regtest         # not currently checked out
pnpm regtest:start && node scripts/seed-book.mjs
```

Note the `regtest` submodule is uninitialised on a fresh clone (`git submodule status` shows a leading `-`), so `pnpm regtest:start` fails without the init above. Pre-existing, not introduced here.

## Next

`src/lib/offerBook.ts` — decoding orders off the tx stream and classifying bid/ask/price — lands on this branch next, with the seeder as its fixture. The taker path (`fulfill`) needs a reachable emulator endpoint and will be gated behind `VITE_EMULATOR_URL`.


tiero:
image

On mobile the bottomshett doesn't scroll show the bottom/but on and price

PR #927, opened 2026-08-18
Tx history from the SDK activity API, and the Lightning send leg on RfqSwapManager

pietro909:
Phases 1–3 and 5 of `plans/adopt-activity-history.md`, plus the Lightning-send half of the RFQ manager migration.

**Stacked on [#918](https://github.com/arkade-os/wallet/pull/918)** — base changed from `master`, merged in rather than rebased. Review #918 first; the diff here is only what sits on top of it. Depends on [arkade-os/ts-sdk#773](https://github.com/arkade-os/ts-sdk/pull/773), released as `@arkade-os/swap` 0.0.8.

`WalletContext.txs` now comes from `getActivityHistory()` plus an `arkade-wallet:asset-swaps` resolver. No flag — a flag defaulting to off would mean the preview deploy under test was the old path — so the Phase 5 deletions land here too.

## Activity API adoption

**Added**
- `transactionHistory.ts`: `arkTransactionToTx` / `sortLocalTxs` / `txidOfArkTransaction`, extracted from `getTxHistory`
- `activity/assetSwapResolver.ts`: correlation only. `prepare()` re-reads the store every load, so records the restore scan writes after the first load still group
- `activityHistory.ts`: pure `Activity[] -> Tx[]` adapter. Only swap groups collapse — everything else emits one row per member, so the built-in boarding/exit/mint resolvers can't change row counts
- Local metadata is grafted here from a snapshot read once per reload, not per row. It is deliberately not a resolver: an enrichment membership would double every row another resolver already groups
- `Tx.historyKey` carries row identity, since a group id is not a txid and no existing field can hold one

**Removed**
- `mergeAssetSwapActivity` — `buildActivities` + the resolver replace it; its row-construction half survives as `buildAssetSwapActivityTx`
- the per-row metadata fan-out in `getTxHistory`, which stays only as the raw ungrouped read `resolveCancellingSpend` needs

**One visible change.** Trap 5 says re-sort with `sortLocalTxs`, which floats `createdAt === 0` rows (unconfirmed boarding) to the top. `mergeAssetSwapActivity`'s final `b.createdAt - a.createdAt` sort sank them to the bottom, so that ordering flips — legacy `[30, 10, 0]`, now `[0, 30, 10]`. Worth confirming on the preview.

## The Lightning-send leg

Three commits that were written before this branch was last pushed and are new to reviewers:

- **`743d7c68`** — refresh the wallet after a send instead of waiting on a `VTXO_UPDATE` that never comes, since the send lands through the page's own provider rather than the service worker.
- **`e1edb709`** — group a Lightning send's funding, claim and refund into one activity via `swapActivityResolver`, so a failed swap stops rendering as two unrelated rows.
- **`ff9efbd1`** — drive Lightning sends with `RfqSwapManager` instead of wallet-side state, retiring `awaitLnSendOutcome`'s hand-rolled poll loop.

**A scoping note, stated rather than buried.** `plans/rfq-persistence-and-activity.md` lists the send leg as item **D**, out of scope and deferred to its own PR, on the grounds that it is a behavioural change on the send critical path — funding-is-acceptance must not regress into a spinner. That plan is stale on this point: the work is written and it is here. The reasoning for keeping it separate still has force, so **if a reviewer wants `ff9efbd1` split out into its own PR, say so and it will be** — it is the one commit here that changes money-moving behaviour rather than presentation.

## Phase 4 is unblocked, and is the next commit on this branch

Phase 4 was blocked on RFQ swap persistence. #918 supplies it, so the phase is live — **but it is not written yet.** It lands here rather than in the persistence PR because it needs both halves of a dependency that point in opposite directions along the stack: records to group (from #918, below) *and* a resolver registry to attach to (`svcWallet.activity.use`, which exists only on this branch, above).

What it will be, per that plan's Phase C:

- `rfqSwapActivityInputs({ repository, indexer })` replacing the hand-rolled `lnSendActivityInputs` in `src/lib/lnSendRecords.ts`. The shipped helper does strictly more: it takes per-corridor txids from the handler's `activityTxids(profile)` rather than reading profile keys by name — which `rfqCorridors.ts` forbids on purpose — and it drains the manager's stamped `lockupSpendArkTxids` before any network read, so a terminal swap answers for its own counterparty spend instead of paying a lockup read for it.
- The outcome-token → wallet-copy mapping, which stays ours. **`lost` is the one to get right:** the resolver emits it for a `lightning_receive` that ended `refunded`, and on that leg every non-claim leaf of the covenant is the solver's — so it is a loss, not money coming back. A send leg's `refunded` is the opposite. #918 already renders that distinction on the receive screen and the history row must agree with it.

## Testing

`pnpm test:unit` — 86 files, **707 passed, 1 skipped**, against the published `@arkade-os/swap` 0.0.8 and with #918 merged in.

Tested end to end through the real `buildActivities` (borrowing `ServiceWorkerWallet.prototype.getActivityHistory`) over a history holding a swap couple, a real asset mint, a boarding tx and a plain receive. Before `mergeAssetSwapActivity` was deleted the same test compared row-for-row against the old producer and passed; the current assertions are derived from that.

Phase 0 step 2 (which built-ins fire on regtest) wasn't run, so Phase 2 step 6 — dropping the `assetAction: 'issued'` writes in favour of `assetMintResolver` — is left alone.

## Merge conflict resolved here

`src/index.tsx` — this branch's `LnSwapsProvider` and #918's `LnReceiveProvider` both claimed the same slot in the provider tree. Both are needed and neither depends on the other, so they nest: `LnSwapsProvider` > `LnReceiveProvider`, both inside `WalletProvider` and `AssetSwapsProvider`, which is all either requires.


pietro909:
Stacked above this: [#929](https://github.com/arkade-os/wallet/pull/929) — Activity Phase 4, which closes the last blocked phase of `plans/adopt-activity-history.md`.

It also fixes something that would have bitten this branch: `lnSendActivityInputs` stores the send's funding txid under a private `profile.funding_txid` key, and the package reader cannot see it (`lightning_send` has no `activityTxids`). #929 moves it onto the record's own `fundingArkTxid`.

PR #936, opened 2026-08-31
Ts sdk 0.5

bordalix:
Closes #934  

Verification on the final merge commit:

- `pnpm format:check`
- `pnpm lint`
- `pnpm exec tsc --noEmit`
- `pnpm test:unit` — 99 files passed; 863 tests passed, 1 skipped
- `pnpm build`

One concurrent full-suite run completed every assertion but exited on a post-test Framer Motion teardown rejection from `contracts.test.tsx`; that file passed alone (8/8), then the full 99-file suite passed cleanly when rerun without concurrent build pressure.

PR #837, opened 2026-07-23
Upgrade ts-sdk 0.4.49 - boltz-swap 0.3.54

pietro909:
Replaces https://github.com/arkade-os/wallet/pull/816

## Summary by CodeRabbit

* **Chores**
  * Updated underlying Arkade swap and SDK components to newer patch versions.
  * Includes the latest upstream fixes and improvements without changing the application’s user-facing functionality.

PR #944, opened 2026-09-04
feat(send): route on-chain sends through PaymentRouter, solver rail first

Kukks:
> **This PR now targets `ts-sdk-0.5` (#936), not `master`.** It is rebased onto `bd507474` (the base moved twice during this work, most recently the rc.3 dependency bump; re-rebased and re-gated against each) and force-pushed. The 12 commits below are the original ones, replayed; a 13th (`ef10dcee`) carries the whole API translation in one reviewable diff. Nothing was merged and the base was not changed.

Supersedes #941, which stays open until this is reviewed. #941 shipped the `arkade:BTC -> onchain:BTC` send with ~2,600 lines of bespoke route selection, bounds matching and fallback; `PaymentRouter` already does all three. This is the same corridor as a rail registration.

## Why this needed a port and not a rebase

#936 replaces the two `RfqSwapManager`s with one `SwapClient` under one Web Lock, and deletes three files this branch was built on: `providers/lnSwaps.tsx`, `lib/lnSendRecords.ts` and `lib/nostrRfq.ts`. Only one of them showed up as a git conflict; the rest were semantic. `src/lib/sendRouter.ts` imported `solverOnchainRail` and `SolverOnchainSend` from `@arkade-os/swap`, and the 0.1.0-rc.2 that `ts-sdk-0.5` pins does not export them from the package root.

The mechanical rebase produced 48 type errors across 8 files. What follows is how they were resolved.

## The design call: the rails come from the package now

`@arkade-os/swap@0.1.0-rc.2` ships `onchainSwapRail(client, deps)` and `lightningRail(client)` — adapters over an already-constructed `SwapClient` — plus `SWAP_ROUTER_PRIORITY`, which is `["ark", "lightning", "onchain-swap", "onchain"]`. That is this PR's own ranking, upstreamed: the solver rail ahead of the collaborative exit, both registered, the preference self-healing by amount.

So the two solver rails are the package's, over the driving tab's client. The wallet keeps only the rails that have no counterparty — the collaborative exit and the asset send.

**The alternative, and what it cost.** `RfqSwapManager` and `solverOnchainRail` do still exist, in `@arkade-os/swap/protocol`, so the smaller change was to retarget the imports and keep `lnSwaps.tsx` as it was. That was rejected: it also needs `lib/lnSendRecords.ts` and `lib/nostrRfq.ts` resurrected, and it stands a second swap driver with a second record store beside the client that #936 made the single owner — outside the Web Lock #936 introduced to stop exactly that. It would have been a smaller diff that reintroduced the hazard the base branch exists to remove.

`createSwapPaymentRouter` was not used wholesale either: it registers the SDK's own `arkRail()` and `onchainRail()`, and this wallet's exit is `collaborativeExitWithFees` (see "does NOT change", below). The rails are registered by hand instead, in the same order.

## The Web Lock argument no longer holds, and it inverts

This PR's `lnSwaps.tsx` docblock said `claimLockup` stays unwired *because* the receive leg's own manager holds a Web Lock over the same Arkade lockup, so wiring it here would put a second unsynchronised claimer on it.

In v2 that reasoning is void. There are not two managers — there is one client, and `providers/swaps.tsx` says so directly: *"the client claims on behalf of every corridor, so the Web Lock covers the whole client."* There is no second claimer to avoid, and there are no manager callbacks to leave unwired. Carrying the comment forward would have described machinery that no longer exists.

## What the package now owns, and this branch no longer does

- **`reserveOnchainSend`.** The pre-funding persist was this branch's most load-bearing ordering rule. `accept()` now persists before it funds, inside the corridor, so there is no wallet-side ordering left to get wrong — and no wallet-side seam that could get it wrong.
- **The negotiation, the market pick and the rendezvous.** The rails are adapters; `Form.tsx` no longer picks a solver.
- **`claimLockup`.** One client, one lock, every corridor.

## What the wallet must still supply: the onchain claim

`OnchainCorridorDeps.claim` is *absent by default and deliberately not defaulted* — the package will not guess a fee rate or a signer. Without it the drive reports an `arkade -> onchain` swap's L1 half **blocked**: the send funds an HTLC this wallet then never claims, and the money comes back only by refund.

`ts-sdk-0.5`'s `makeSwapClient` configures the lightning corridor only, so that hole was live on the base branch. This PR's `claimOnchain` body moves out of the deleted provider and into `src/lib/swapClient.ts` as that dep, with two changes: it reads `payoutPkScript` off the swap rather than rebuilding it from the stored profile, and it reads the record through `corridorRecordStore` over the client's own repository.

**That last part was wrong in the first push of this port and is worth calling out.** It read `assetSwapRepository.getRfqSwap(rfqId)` — the **v1 `rfqSwaps` store**, which the v2 client never writes (`corridorRecordStore` reads and writes `getAllSwapRecords`/`saveSwapRecord` instead). It would have answered `undefined` for every swap, so the claim would have thrown `no stored record for rfq …` *after the HTLC was funded* — exactly the failure the dep exists to prevent. The first version of the test missed it because it mocked the repository with a `getRfqSwap` the real one would never be asked for. The test now stands up a fake speaking the **v2 record API only**, and reverting the fix fails it with `TypeError: assetSwapRepository.getRfqSwap is not a function`.

`l1PayoutPubkey` and `claimPayoutScript` are deleted with the provider.

## A translation hazard worth naming

`quoteIsForThisInvoice` compared `quote.meta?.invoice` to the routed invoice. The v2 rail's `meta` carries `quoteId`, `expiresAt`, `paymentHash`, `market` and `solver` — and never the BOLT11. Ported literally, that guard returns `false` for **every** Lightning send, refusing all of them with "Quote is for a different invoice". It now compares payment hashes. Mutating it back kills 6 tests.

## Three things this deliberately does NOT change

1. **Coin selection.** The collaborative exit stays this wallet's `collaborativeExitWithFees`, not the SDK's `onchainRail`. That one settles through `Ramps.offboard` with its own selection, while this wallet spends soonest-expiring VTXOs first. Adopting it is a coin-selection change and belongs in its own PR. This is the reason `createSwapPaymentRouter` is not used directly.
2. **What a typed number means.** `RouteQuote` is receiver-exact; this wallet's form lets the user type what *leaves*. The translation happens at the screen boundary, not in the rails — interpreting the typed number as receiver-exact would gross it up past the balance it came from, so "send max" could never succeed.
3. **The exit's fee model.** `walletExitRail` charges `calcOnchainOutputFee()` and calls `collaborativeExitWithFees(wallet, payout + fee, payout, address)` — byte-identical to the pre-router call, asserted in the tests.

## The wrong-address guard

#941 fixed a real bug: navigate to sign, go back, type a different address, continue — and the payment went to the **first** recipient, because the carried quote held their payout script.

`RouteQuote` has no address field, so **nothing is carried**: `PaymentOption.quote()` is lazy by contract, so `Details.tsx` quotes against the address it is displaying at the instant of the spend. A stale quote cannot exist. `quoteIsForThisSend` is the belt to those braces: the screen must be showing the address that was routed to, the recipient must get exactly what the screen displayed, and no more may leave than the screen displayed (less is allowed). All three arms are mutation-tested and unchanged by this port.

## Address predicates

`isBTCAddress` uses the SDK's `isBtcAddress` — so the form classifies as `btcTarget` does — plus a decode gate, now imported from `@arkade-os/swap/protocol`. Acceptance changes in both directions, pinned in `address.test.ts`:

| Case | Before | After |
| --- | --- | --- |
| testnet/regtest legacy (`m`, `n`, `2`) | rejected | **accepted** |
| mixed-case bech32 | accepted | **rejected** — BIP173 forbids it |
| undecodable bech32 (bad checksums) | rejected | still rejected |
| mainnet p2wpkh / p2tr / p2pkh / p2sh, regtest bech32 | accepted | accepted |

## Dependencies

**This branch does not touch `package.json` or `pnpm-lock.yaml`.** It runs on `ts-sdk-0.5`'s own dependency state — now `@arkade-os/sdk@0.5.0-rc.3`, `@arkade-os/swap@0.1.0-rc.3`.

The rc.3 bump (`bd507474`, *"the promoted names come home, record reading moves to /advanced"*) moved two of this branch's imports, exactly as that commit's own note in `swapRecords.ts` warns it can: `corridorRecordStore` and `type OnchainClaim` are now on `@arkade-os/swap/advanced`. Nothing else this branch imports moved — the claim primitives stayed on `/protocol` and the two rails stayed on the root. Caught by `tsc` (5 errors) on the rebase, not at runtime. The `.gitattributes` commit from the previous round was dropped during the rebase: `ts-sdk-0.5` already carries the identical `* text=auto eol=lf` rule with a fuller comment.

## The `as unknown as` casts (#950)

#950 reports that this branch introduces two kinds of double cast. One kind is not load-bearing and is now gone: `collaborativeExitWithFees(ctx.wallet, …)` in the exit rail and `sendAssets(ctx.wallet, …)` in the asset rail. `RouterContext.wallet` is already the concrete `Wallet`, which satisfies `IWallet`, so both assign directly. Re-verified on this base rather than taken from the issue — `ts-sdk-0.5` is not the master #950 was filed against — and `npx tsc --noEmit` stays at 0 errors with both removed.

The remaining one is forced and stays:

```ts
wallet: deps.wallet as unknown as ConstructorParameters<typeof PaymentRouter>[0]['wallet'],
```

`RouterContext.wallet` is typed as the concrete `Wallet` class, and this app holds a `ServiceWorkerWallet`, which *implements* `IWallet` but does not *extend* `Wallet` — so the assignment is illegal by nominal typing, with no narrower cast available. That reason is now in the comment at the call site rather than a bare "intentional". It can be deleted once the SDK types `RouterContext.wallet` as `IWallet`; that change is being made upstream in `arkade-os/ts-sdk`, and #950 tracks the wallet side.

## `49124dff` — the `onchain-swap` rail could not win an on-chain send

Reported from a mutinynet preview build. Every ark -> L1 payment silently took
the collaborative exit; the solver rail was never usable:

```
onchain send: onchain-swap refused: quote pays 40000 for 40754, screen shows 40000 for 40000 to tb1qmt3ue...
onchain send: paying via onchain
```

The send succeeded, so no money was lost — the rail was simply unreachable.

**The chain.** `Send/Details.tsx` computed the total the user signs for *before*
any rail had quoted, from `calcOnchainOutputFee()`:

```ts
const total = pendingLnSend ? pendingLnSend.total : satoshis
const amount = direction === 'Paying to mainnet' ? satoshis - calcOnchainOutputFee() : satoshis
const fees = total - amount > 0 ? total - amount : 0
```

That fee is the **collaborative exit's**, and it is subtracted from the amount
rather than added to the total — so `total` is just what the user typed. The
solver rail is receiver-exact: the recipient gets `amount`, and the spread plus
the L1 claim go **on top**, making `total > amount` for any rail that charges
anything. `quoteIsForThisSend` then compares the two:

```ts
quote.total <= (screen.total ?? 0)
```

With `screen.total === screen.satoshis` that is false for every rail with a fee.
On this deployment `calcOnchainOutputFee()` is 0, so there was not one satoshi
of headroom. The screen and the rail held incompatible fee models — one taking
the fee out of the amount, the other putting it on top — and the guard resolved
the disagreement by always rejecting the rail.

**The shape chosen: quote first, show that rail's real total, sign that figure.**
The sign screen now asks the router what the send actually costs before it offers
anything to sign, displays that rail's fee and total, and holds the button until
the quote lands. The spend re-quotes, and the guard — unchanged — refuses any
rail that has since got dearer than the figure on screen.

The guard is not the bug and was not weakened. `quoteIsForThisSend` and
`payOnchain` are byte-for-byte unchanged, including the deliberate absence of a
catch around `send()`: a rejection there can mean a covenant that *is* funded,
and trying the exit next would pay the recipient twice. All that changed is that
the ceiling the guard measures against is now a figure a rail will honour, and
one the user actually saw.

**Alternatives, and what they cost.**

- *Carry the quote from the form, as the Lightning path does.* This is the
  smallest-looking option and was rejected: it would make the economic half of
  the guard structurally satisfied — the screen's total would derive from the
  very quote being checked — which is a guard that passes vacuously. It also
  reintroduces the carried-stale-quote exposure this PR deliberately removed for
  the on-chain path, and moves routing across a screen boundary along with new
  flow state to clear on address change.
- *A fixed budget or ceiling on the screen.* No principled number exists without
  a quote. Any constant would be invented policy, and a ceiling shown higher than
  what is charged is its own dishonesty.
- *Invert each rail's fee so exactly what the user typed leaves the wallet.* Not
  available: the router API is receiver-exact and rail fees are amount-dependent,
  so this needs a quote-then-requote that does not converge.

The cost of the option taken is a second quote per on-chain send (one to price
the screen, one to spend) and a short wait before the button arms.

**The Lightning and asset paths do not share the defect.** Lightning already
works this way — `Form.tsx` quotes before navigating and the screen shows
`pendingLnSend.total`, then sends that same quote object. The asset path prices
the screen and the rail from the same `defaultFee` constant, so they agree by
construction (it carries no total guard at all, which is noted, not changed
here).

### `748330df` — the claim fee no longer floors at 1 sat/vB

The finding reported in the previous round is now **fixed on this branch**.
`src/lib/claimFee.ts` asked for Esplora's `/fee-estimates`, which every Arkade
mempool deployment answers with a **404**. The catch returned
`MIN_CLAIM_FEE_RATE`, so **every L1 claim was priced at the 1 sat/vB floor on
every network, mainnet included** — silently, which is how it survived this long.

**The evidence is runtime, not a test assertion.** The real `claimFeeRate` was
run against the live deployments, before and after the change:

| base URL | before | after |
| --- | --- | --- |
| `https://mempool.arkade.sh/api` (**mainnet**) | 1 sat/vB | **3 sat/vB** |
| `https://mempool.mutinynet.arkade.sh/api` | 1 sat/vB | 1 sat/vB |
| `https://mempool.signet.arkade.sh/api` | 1 sat/vB | 1 sat/vB |
| `https://mempool.space/testnet/api` | 1 sat/vB | 1 sat/vB |
| `https://mempool.arkade.sh` (no `/api`) | 1 sat/vB, silent | 1 sat/vB, **logged** |

Mainnet moving **1 → 3** is the whole justification for this change; it matches
that host's live `fastestFee: 3` at the time of the probe. **The rows that stay
at 1 are not a partial result.** mutinynet, signet and mempool.space testnet each
report `fastestFee: 1` right now, so 1 sat/vB is the correct answer on a
genuinely calm network. Before the fix all five rows read 1 for the wrong reason;
after it, four read 1 for the right one.

**Why the 404, precisely.** `GET https://mempool.arkade.sh/api/fee-estimates`
returns a body reading `Cannot GET /api/v1/fee-estimates` — the deployment
rewrites an unrecognised `/api/X` to `/api/v1/X`, so `fee-estimates` is simply
not in its Esplora-compat allowlist, while `/api/blocks/tip/height` is (200).
The base URL was never wrong.

**The mapping, and the judgement in it.** mempool serves named tiers, and there
is no exact two-block tier. `fastestFee` (next block) is taken first, with
`halfHourFee` (~3 blocks) as the looser fallback. Erring fast is the safe
direction on an asymmetric loss: an unconfirmed claim loses the entire fill,
while overpaying costs the tier delta on a ~150 vB transaction — on today's
mainnet, `fastestFee 3` against `halfHourFee 1`, roughly 300 sats. It also
preserves the existing intent, since the old `['2','1','3']` order already fell
*tighter* before it fell looser.

**Both shapes are tried, and the honest version of why.** The SDK's `ESPLORA_URL`
is not homogeneous: the three Arkade hosts are mempool.space, but its `testnet`
entry is real mempool.space, which *does* serve `/fee-estimates` — at HTTP **203**
rather than 200. So no currently-configured endpoint needs the Esplora fallback
*exclusively*; today it is insurance rather than coverage. It is kept because
`swapClient.ts` builds `new EsploraProvider(esploraUrl)` against the very same
base URL — the codebase already asserts this URL is an Esplora — and a
self-hosted electrs is the plausible deployment. It costs nothing on the common
path: the second shape is only fetched when the first is not served at all. That
heterogeneity is now recorded in a comment at `FEE_SOURCES`, since it is the kind
of fact that otherwise gets re-derived painfully.

**`MIN_CLAIM_FEE_RATE` is untouched, both of its jobs intact** — the fallback
rate, and the believability bar it sets through `usableRate`. The fix never
needed to move it, so the two constants were not split.

**The decoy is refused, not parsed.** `/fee-estimates` at a base URL missing
`/api` returns HTTP 200 with `text/html` — the explorer SPA — and so does
`/v1/fees/recommended`. A response is now refused on `content-type` before it is
parsed, rather than reading as an empty fee set.

**It no longer fails closed in silence.** A source that is *read* and reports a
cheap network returns the floor quietly, as it should. A source that cannot be
read at all — 404, non-JSON, or a throw, on every shape — now goes through
`consoleError`, which persists to localStorage for the in-app log view as well as
hitting the console. Confirmed live: the `/api`-less probe above emitted
`claim fee: no fee source answered at https://mempool.arkade.sh, claiming at 1 sat/vB`,
where the original emitted nothing at all.

**The strictest new rule, named.** Requiring a JSON `content-type` is the part of
this change most likely to bite: a deployment serving correct fee JSON with a
missing or unusual `content-type` header would now be refused. The blast is
bounded — it floors, which is exactly the old behaviour — and it is now logged
rather than silent. **If a deployment ever reports fee estimates not being picked
up, that content-type gate is the first thing to check.**

**Upstream.** The SDK's own `EsploraProvider.getFeeRate()` has the identical bug
against the identical hosts — `${baseUrl}/fee-estimates`, returning `fees["1"]` —
differing only in that it throws instead of flooring. This repo never calls it,
so it is unreachable here; filed as arkade-os/ts-sdk#858 for other consumers.

**CI does not run on this base** (`ci.yml`/`playwright.yml` gate on
`master`/`next`), so the local gate below is the only evidence for this commit.

Baseline re-measured on the branch head `49124dff` before any change, after
`pnpm git-info`:

| | Baseline (`49124dff`) | With `748330df` |
| --- | --- | --- |
| `npx tsc --noEmit` | 0 errors | **0 errors** |
| `pnpm test:unit` | 91 files, 775 passed, 1 skipped, **0 failing** | **91 files, 780 passed, 1 skipped, 0 failing** |
| `pnpm lint` / `format:check` / `csp:check` | clean | clean (pre-commit hook ran; no `--no-verify`) |

The +5 tests reconcile exactly: `claimFee.test.ts` goes 5 → 10 cases, file count
unchanged at 91. The `useBounceMorph` flake described above did not fire in
either run.

| Mutation | Result |
| --- | --- |
| Revert to `/fee-estimates` | **killed**, 5 tests — `reads the named tiers an Arkade mempool deployment actually serves`: `expected 1 to be 3` |
| Treat the HTML decoy as a valid empty response | **killed**, 2 tests — `refuses a 200 carrying HTML instead of reading it as an empty fee set`: `expected 1 to be 6`; and `reports a fee source it could not read at all`: `expected "spy" to be called once, but got 0 times` |
| Pick the looser tier (`halfHourFee` first) | **killed** — `prefers the tightest tier, because an unconfirmed claim loses the fill`: `expected 2 to be 7` |
| Drop the `usableRate` bar so a `0` tier is believed | **killed** — `skips a tier at or below the floor rather than believing it`: `expected +0 to be 5` |

No mutation survived. Returning the floor is the *same value* whether the HTML
decoy is refused or parsed, so that mutation is killed on observable behaviour
instead: a refused source falls through to the other shape, and logs when nothing
is readable at all.

#### `39404b9b` — a regression in `748330df`, caught and fixed

`748330df` moved `baseUrl.replace()` outside the try block, so a base URL the
endpoint map could not name threw out of `claimFeeRate` instead of falling back
to the floor, leaking the abort timer with it. The original code caught that;
the guarantee is restored, and the cause is now reported rather than swallowed.

It is reachable. `providers/swaps.tsx` reads the endpoint through an unchecked
`aspInfo.network as NetworkName` cast, so `ESPLORA_URL[network]` is `undefined`
for any network outside the SDK's map. Throwing there blocks the send router
outright, where the floor would merely have made the claim slow — strictly worse
than the behaviour being fixed.

Found by re-reading arkana-ai-bot's observation on `7710aeb9` against the *new*
code rather than the old one it was written about. Its premise is not quite
right — `mutinynet` **is** in `ESPLORA_URL` (`https://mempool.mutinynet.arkade.sh/api`),
so it is not an example of an unnamed network — but the underlying point about
the unchecked cast was sound, and it caught a real bug I had just introduced.

| | Baseline (`49124dff`) | With `748330df` + `39404b9b` |
| --- | --- | --- |
| `npx tsc --noEmit` | 0 errors | **0 errors** |
| `pnpm test:unit` | 91 files, 775 passed, 1 skipped, **0 failing** | **91 files, 781 passed, 1 skipped, 0 failing** |
| `pnpm lint` / `format:check` / `csp:check` | clean | clean (pre-commit hook ran; no `--no-verify`) |

| Mutation | Result |
| --- | --- |
| Remove the restored `catch` | **killed** — `never throws on a base URL the endpoint map could not name`: `TypeError: Cannot read properties of undefined (reading 'replace')` |

## Test plan

> **CI runs no tests on this PR. The gate below was run by hand.**
>
> `ci.yml` and `playwright.yml` both filter on `branches: [master, next]`, and
> `pull_request.branches` matches the BASE — so a PR into `ts-sdk-0.5` triggers
> neither the unit suite nor E2E. The only checks here are the two Cloudflare
> Pages builds, which say the site compiled, not that it works. Every number in
> this section comes from `pnpm test:unit` and `npx tsc --noEmit` run locally.
> Filed as #966; please do not read the green checkmarks as verification.

Baseline recorded on `origin/ts-sdk-0.5` (`bd507474`) before any change, after `pnpm install` and `pnpm git-info`:

| | Baseline | This branch |
| --- | --- | --- |
| `npx tsc --noEmit` | 0 errors | **0 errors** |
| `pnpm test:unit` | 86 files, 701 passed, 1 skipped, 0 failing | **91 files, 781 passed, 1 skipped, 0 failing** |
| `pnpm lint` | clean | clean |
| `pnpm format:check` | clean | clean (pre-commit hook ran; no `--no-verify`) |

The +5 files / +80 tests reconcile exactly: 61 from the five new unit test files, +4 in `address.test.ts` (17 → 21 cases), +5 for the no-fallthrough, record-store, real-parser and two take-leg tests described below, +4 for `49124dff`, and +6 for `748330df` with `39404b9b` (`claimFee.test.ts`, 5 → 11 cases).

**The suite's timeout margin, and why this branch changes it.** `vitest.config.ts` set no `testTimeout`, so the 5 s default applied. On `bd507474` the suite is 3/3 green; with this branch's five extra test files it was **4 of 5 runs red** — always the same two `swap.test.tsx` tests, always `Test timed out in 5000ms`, never an assertion. They pass 40/40 when that file runs alone. The base has hit the same wall in a different file (`receive-lightning-error.test.tsx`, 4 tests, green on re-run), which is what says the margin is short rather than either test being broken.

So the fragility is pre-existing but this branch is what makes it fire, and "pre-existing" is not a reason to hand over a red suite. `7710aeb9` sets `testTimeout: 15_000`. After it: **3/3 green**. It is a separate commit so it can be taken, moved to #936, or dropped on its own — and it is the one change here that touches shared config rather than this feature. `git diff origin/ts-sdk-0.5 -- src/test/` is **1242 insertions, 0 deletions** — no test on the base branch was removed. Two test files this PR had itself added (`lnSwaps.test.tsx`, `onchainSendRecords.test.ts`) were deleted with the code they covered; `swapClientOnchainClaim.test.ts` keeps that coverage against the client dep.

The router's tests are rebuilt on a fake `SwapRailClient` — the structural seam the package documents for exactly this — instead of mocking the negotiation, the transport and the record store.

### Mutations

| Mutation | Result |
| --- | --- |
| Rank the exit rail ahead of the solver rail | **killed**, 4 tests — `expected [ 'onchain', 'onchain-swap' ] to deeply equal [ 'onchain-swap', 'onchain' ]` |
| Unwire `corridors.onchain.claim` | **killed**, 5 tests — `expected 'undefined' to be 'function'` |
| Restore the `meta.invoice` comparison | **killed**, 6 tests — `expected false to be true` |
| Catch a failed `send()` and try the next rail | **survived**, then fixed — see below |
| Read the record from the v1 `rfqSwaps` store | **killed**, 5 tests — `TypeError: assetSwapRepository.getRfqSwap is not a function` |

The pre-funding-persist mutation this PR's previous round carried is no longer expressible: the ordering moved into the package's `accept()`, so there is no wallet-side seam to invert. The wallet-side analogue is the rule that a `send()` which may already have funded is **not** retried on another rail — a rejection can mean a covenant that *is* funded, and falling through to the exit pays the recipient twice.

That rule had a docblock and no test. Mutating it to catch-and-continue passed the entire suite — 91 files, 764 tests, exit 0. `send-onchain-route.test.tsx` now pins it, and the mutation fails with `expected undefined to be defined`.

### `49124dff`: gate and mutations

**CI does not run on this base.** `ci.yml` and `playwright.yml` are gated on
`branches: [master, next]`, so a `MERGEABLE`/`CLEAN` state on this PR carries no
signal about the code. The local gate below is the only evidence.

Baseline re-measured on the branch head `7710aeb9` before any change, after
`pnpm git-info`:

| | Baseline (`7710aeb9`) | With `49124dff` |
| --- | --- | --- |
| `npx tsc --noEmit` | 0 errors | **0 errors** |
| `pnpm test:unit` | 91 files, 771 passed, 1 skipped, **0 failing** | **91 files, 775 passed, 1 skipped, 0 failing** |
| `pnpm lint` / `format:check` / `csp:check` | clean | clean (pre-commit hook ran; no `--no-verify`) |

The +4 tests reconcile exactly: +3 in `sendRouter.test.ts` (`previewOnchainCost`)
and +1 in `send-onchain-route.test.tsx` (7 → 8). No test was removed; the
existing `never funds a rail that would spend more than the screen showed` was
rewritten in place, because its old scenario — a rail quoting 10,001 against a
screen showing 10,000 — is now shown to the user at 10,001 rather than refused.
Its intent is preserved by the only way a rail can still exceed what was shown:
getting dearer between the screen's quote and the spend's.

| Mutation | Result |
| --- | --- |
| `quoteIsForThisSend` always returns `true` | **killed** — `never funds a rail that would spend more than the screen showed`: `expected "spy" to be called 1 times, but got 0 times`; and `refuses a quote that spends more than the screen displayed`: `expected true to be false` |
| `quoteIsForThisSend` always returns `false` | **killed**, 7 tests — `funds a solver that charges on top, having shown that total before signing`: `expected "spy" to be called 1 times, but got 0 times` |
| Drop the `quote.total <= screen.total` conjunct (let a rail charge more than was shown) | **killed** — same two tests, `expected "spy" to be called 1 times, but got 0 times` / `expected true to be false` |
| Stop applying the quote to the screen (restore `total = satoshis`) | **killed** — both new tests: `Expected element to have text content: /40.?754/ · Received: 40,000 sats`, which is the operator's symptom exactly |
| `previewOnchainCost` accepts a quote for a different payout | **killed** — `skips a rail quoting a different payout rather than showing its total`: `expected { amount: 5000, fee: 45000, …(1) } to deeply equal {…}` |

No mutation survived.

### A pre-existing suite flake this change makes more likely to surface

`pnpm test:unit` can exit 1 with **every test passing** and `Errors 1 error`:

```
Error: controls.start() should only be called after a component has mounted.
 ❯ bounceAndMorph src/hooks/useBounceMorph.ts:62:33
 ❯ runLoop src/hooks/useBounceMorph.ts:96:17
```

`useBounceMorph` drives an unbounded `while (true)` loop that awaits framer-motion's
`controls.start()` and checks `cancelled.current` only *between* awaits, so a
component unmounting mid-animation leaves the next `start()` firing on a gone
component. It is reachable from every screen rendering `LoadingLogo`.

Measured: **0 occurrences in 8 clean-base runs; 3 in 11 runs with this commit.**
It was attributed in turn to `contracts.test.tsx` and `receive-qrcode.test.tsx`
— neither touched here, and vitest itself notes the attribution is to whichever
file was running, not the origin. Running the suite with this commit's source fix
but its two new screen tests removed was clean 3/3, so the added mid-animation
unmounts are what raise the odds.

An attempt to contain it inside these tests — waiting for the success animation
to finish before teardown — cost 3.2 s and still failed 1 in 5, with the next
failure landing in a third unrelated file. It was reverted rather than carried.
The root cause is in a file this PR does not touch and the fix belongs in
`useBounceMorph`'s teardown; flagged here rather than fixed, to stay in scope.
No test fails when this fires — only the run's exit code.

## Not verified

- **No live regtest run.** The solver rail's `quote()`/`send()`, the persist-before-fund inside `accept()`, and the L1 claim through the new corridor dep are covered by types, unit tests and the package's own tests, but have not been exercised end to end here. A real claim against a real funded HTLC has not been observed.

  Three links in that path *are* now checked without a stack. The claim reads through the same store API the client writes; one test runs the **real** `corridorRecordStore` and the **real** `rfqClaimSecretOf` against each other, so the two packages are known to agree on the record shape rather than assumed to; and two tests pin what happens when the solver's answer is not what the gross-up asked for.

  That last pair matters because the rail asks for `amount + claimFee` on the take leg and trusts the answer. A solver returning a **short** take leg produces a `RouteQuote.amount` below what the screen is showing, and `quoteIsForThisSend` refuses it — so the recipient is not silently short-paid, the rail is skipped and the exit ranks instead. A solver returning legs that do not satisfy `give == take + fee` is refused outright by `receiverExact`, before a lockup exists. What remains unverified is the step after that: whether the HTLC a real solver actually locks on L1 matches the take leg it quoted.
- `src/test/e2e/send-onchain-fallback.test.ts` was updated for the rail id rename (`solver-onchain` → `onchain-swap`) and its fallback assertion tightened to an exact match, since `includes('paying via onchain')` would otherwise also match `onchain-swap`. It compiles and is formatted, but Playwright was not run.

## Summary by CodeRabbit

* **New Features**
  * Added on-chain payment routing through solver-based routes and wallet exits.
  * Payments now select suitable routes based on destination, amount, fees, and availability.
  * Added recovery and cleanup for interrupted or expired on-chain payments.
  * Added automatic on-chain claim support with fee estimation and wallet signing.
* **Bug Fixes**
  * Improved Bitcoin address validation across mainnet, testnet, and regtest formats.
  * Prevented mismatched or invalid payment quotes from being used.
  * Improved handling of unavailable routes and missing payment destinations.


Kukks:
Thanks — three of the four were real, and one was a money bug I had carried over from #941 without questioning it.

**Taken:**

1. **`isUnfundedReservation` (`onchainSendRecords.ts`) — this was the serious one.** The record is written *before* the funding is broadcast, and the indexer learns of it later still, so a successful-but-empty lockup lookup does not distinguish "never funded" from "funded ten seconds ago". A restart in that window dropped the record and lost an L1 claim only this wallet can make. Now anything younger than `RESERVATION_GRACE_SECONDS` (10 min) is kept regardless. The asymmetry is the argument: a stale row prunes itself, a dropped funded one costs the fill. Covered by `src/test/lib/onchainSendRecords.test.ts`, and mutating the age guard away kills it.

2. **`handleSent` fee (`Details.tsx`).** Correct — `quoteIsForThisSend` deliberately admits a quote whose `total` is *lower* than displayed, so persisting `details.fees` could record a fee nobody paid. Now persists `quote.fee`.

3. **`isBTCAddress` (`address.ts`).** Taken, though for a narrower reason than stated: `isBtcAddress` is format-only *by design*, because a rail's `match()` runs before anything is known, so the fix belongs at the form rather than in the SDK predicate. `isBTCAddress` now also requires the address to decode on one of the three L1s. Note it was never a spend risk — `Address.decode` throws before any output is built, which I verified at runtime — but failing at the form beats failing at the signature.

   This also caught a bad fixture of my own: `2N2JD6wb56AFK4d9ppKAftvBoT9Cw3xL1Tm`, which I had cited in the PR description as a testnet p2sh, has an invalid checksum and fails on all three networks. Replaced with a verified one and the description is corrected.

**Declined:**

4. **`reserveOnchainSend`'s optional `addSwap`.** The optional chaining is deliberate and matches `trackLnSend` directly above it. The record is saved *first*, and that durability is the invariant that matters: a reservation the manager never hears about still restores and is driven on the next start. Throwing would refuse a send the wallet can make, in exchange for a case that self-heals. If you think the deadline on this corridor makes "driven this session" load-bearing enough to warrant it, I'll reconsider — but it would be a behaviour change, not a bug fix.

Pushed `e964c542`.

Kukks:
Rebased onto `be14ac55` — pushed `b6cbb709`.

**The vendoring is gone.** #945 landed while this was open and bumped `@arkade-os/swap` to **`0.0.12`**, which ships #837. So the tarball, `vendor/README.md` and the `pnpm.overrides` block are all deleted: this branch no longer touches `package.json`, `pnpm-lock.yaml` or `vendor/` at all, and sits on master's own dependency state (`sdk@0.4.69`, `swap@0.0.12`). Verified after install that `0.0.12` carries `solverOnchainRail` and that `SolverOnchainRailDeps` matches the shape this was written against. Nothing left to roll back on the dependency side.

Rebase was clean — the boundary listed only my four commits, so no `--onto` was needed, and conflicts were confined to the two dependency files.

Re-verified after the rebase rather than assumed:

- all six money-path invariants still present textually (boundary fee translation, `quoteIsForThisSend`'s three arms, its call site, exit-rail amounts, `persist`-before-fund, the reservation grace period);
- 12 mutations re-run against the new SDK — all still killed;
- gate: 91 files / 788 tests, `tsc` unchanged at the one pre-existing error, lint clean, all 11 touched files format-clean against staged blobs;
- all 7 CI checks green on `b6cbb709`.

Also added, per review of my own uncertainty: `RESERVATION_GRACE_SECONDS = 600` now says in the code that it is a judgement rather than a measurement, and names the symptom of it being too short (a funded record dropped on restart, losing an L1 claim — a silent failure). If anyone has real indexer-lag figures, that constant should be set from them.

@coderabbitai review

Kukks:
Fee-bucket finding taken — `e0111f1c`, all 7 checks green.

`estimates['2'] ?? estimates['1'] ?? estimates['3']` falls through only on null/undefined, so an esplora reporting `{"1": 25, "2": 0}` had the zero taken, rejected by the validity check, and the claim dropped to the 1 sat/vB relay floor with a good 25 in the next bucket. That matters more than it looks here: `claimOnchainFill` refuses inside the refund leaf's margin, so an unconfirmed claim is a lost fill — underpaying by 24 sat/vB on a transaction racing a consensus deadline is exactly what this function exists to prevent. Now takes the first *usable* bucket; `src/test/lib/claimFee.test.ts` covers it and reverting to `??` kills the test.

Worth recording: both real bugs review has found — this and the reservation race — came from code **ported from #941**, not from the router adoption. Neither was introduced here, and neither was questioned when I carried it over.

Gate: 92 files / 793 tests, `tsc` unchanged at the one pre-existing error, lint clean, 12 touched files format-clean against staged blobs, 18 mutations all killed.

Kukks:
All six looked at. Five taken, one pushed back on, and finding 1's fix is not the one suggested — the endpoint that was missing turned out to be the wrong map entirely.

## 1 — mainnet solver absence: real, but `mempool.space/api` would have been the wrong fix

Confirmed the chain: `explorers.bitcoin` carries only `web`, `onchainClaimEndpoint` returned it verbatim, and `createSendRouter` gated registration on it — so mainnet got the exit rail alone.

What the suggested fix would have done, though, is point mainnet HTLC claims at `mempool.space`. The SDK already ships the map this wanted:

```
ESPLORA_URL: Record<NetworkName, string>   // total by type, no optional api
  bitcoin    https://mempool.arkade.sh/api
  signet     https://mempool.signet.arkade.sh/api
  mutinynet  https://mempool.mutinynet.arkade.sh/api
```

`src/lib/explorers.ts` is a different thing — it builds explorer *links*, `api` is optional there, and where the two overlap they disagree (it points signet and mutinynet at `mutinynet.com`). Reading the SDK's map is the fix; `onchainClaimEndpoint` now returns `string`, not `string | undefined`.

That makes the `if (onchainClaimEndpoint(...))` gate unconditional, so it's gone. What drops the solver is `available()` — no card serving the corridor, or an out-of-range amount — which is the real gate and already had tests. Two branches in `lnSwaps.tsx` became unreachable for the same reason (`esploraUrl` can no longer be undefined) and went with it, including a throw that could no longer fire.

Worth flagging on your framing: the absence was **deliberate and test-pinned**, not an oversight. `sendRouter.test.ts` had `never registers the solver where there is no L1 endpoint (was: no_l1_endpoint)` using bitcoin as the example of a network without one. That test now asserts the opposite, and a second case pins that every `NetworkName` resolves an endpoint.

**This turns mainnet solver sends on.** That was an explicit call, not a side effect of the refactor.

## 2 — `l1NetworkOf` — taken

Comment added. Also narrower than described: `NetworkName` includes `mutinynet`, and `explorers.ts` had entries for both it and `signet`, so both already got the rail at `l1Network: 'testnet'`. Correct either way — they share HTLC script parameters — and it now says so rather than looking like a fallthrough.

## 3 — `utxoTxsAllowed()` + `quoteIsForThisSend` — taken

Reproduced exactly. Guarded at the `else if (address)` callsite beside its siblings. The test asserts through `consoleError` rather than the DOM, because the error renders behind the loading animation's exit callback; removing the guard fails it with `expected 'Error: No route for this payment' to match /not permitted/i`, which is the bug verbatim.

## 4 — `reserveOnchainSend` optional chain — taken as a pointer

Added, but as one line rather than a block: `trackLnSend` three lines above does the identical thing and already carries the rationale, so a second copy would be the same explanation twice.

## 5 — the `wallet` cast — pushing back

`sendRouter.ts` already carries `// Typed as concrete Wallet; an IWallet satisfies what the rails call.` at that line. The ask was to add an SDK issue reference, but there isn't an open one to point at — a comment referencing a nonexistent issue is worse than the existing one. Happy to add it if you know of the issue.

## 6 — unmount mid-quote — agreed, filed

Pre-existing, as you say. arkade-os/wallet#946.

## One note on the review itself

The line numbers are consistently off — `sendRouter.test.ts:807-808` in a 145-line file, `lnSwaps.tsx:498-500` in a 225-line file. The findings were accurate and I located all of them, but the citations pointed elsewhere; worth knowing if the clone was stale.

## Also in this push

`c380181a` adds `* text=auto eol=lf` to `.gitattributes`. The pre-commit hook runs `prettier --check`, which failed on all 499 files from a Windows checkout under `core.autocrlf=true`. The stored blobs were already LF, so `git add --renormalize .` reports no content change and the diff is that one file. The `regtest` submodule is untouched.

Kukks:
> please confirm a final human pass covers the `claimOnchain` integration-test gap before cutting a production release

Covered in `529c0e18` — `src/test/providers/lnSwaps.test.tsx`, 7 tests. It drives the installed `claimOnchain` callback end to end from a record written by `onchainSendSwapRecord`, with real `rfqClaimSecretOf`, `preimageForSwapRecord`, `claimPayoutScript` and `claimFeeRate`; only `RfqSwapManager` and `claimOnchainFill` are stubbed. `mustPersistPreimage` puts the raw P in the record, so the round trip needs no signing wallet.

What it pins, and the mutation that proves each:

| pinned | mutation | failure |
| --- | --- | --- |
| payout script comes from the record, never derived | `claimPayoutScript(...)` → a constant | `expected '0101…' to be '5120 0909…'`, plus the no-payout-script test stops rejecting |
| P round-trips and still hashes to the payment hash | `preimageForSwapRecord(...)` → a constant | `expected '0808…' to be '0707…'` |
| fee rate is fetched from `ESPLORA_URL[network]`, mainnet included | `onchainClaimEndpoint` → a map without `bitcoin` (the pre-`c8d284ec` shape) | mainnet case fails with `Number of calls: 0` — `claimFeeRate(undefined)` throws inside its own try and silently returns the floor, so **no estimate is ever fetched on mainnet**. That is finding 1 caught at the integration level rather than at the map |
| an unreachable estimator drops to the floor instead of failing the claim | `claimFeeRate(esploraUrl)` → `Promise.resolve(2)` | 4 tests, `expected 2 to be 13` / `expected 2 to be 1` |
| the sighash goes to `signMessage(sighash, 'schnorr')` | `'schnorr'` → `'ecdsa'` | spy call-args mismatch |
| a missing record refuses **by name** | drop the `if (!record) throw` | `expected [Function] to throw error matching /no stored record/ but got 'Cannot read properties of undefined (…'` |
| a record missing its hashlock refuses and never reaches the fill | read `profile.signer`/`profile.hashlock` directly instead of `rfqClaimSecretOf` | `expected … /claim secret is unreadable/ but got 'this wallet cannot produce the preima…'` |

Unit suite: **92 files / 794 passed / 1 skipped** before, **93 / 801 / 1** after — +7, no regressions. All 8 checks green at `529c0e18`.

One observation while writing it, not a bug and not fixed: for `onchain_send` the `if (!secrets) throw … was stored without its hashlock` branch at `src/providers/lnSwaps.tsx:117` is unreachable. `rfqClaimSecretOf` only returns `undefined` when the corridor handler has no `claimSecret`, and `OnchainSendCorridor` has one — a malformed profile throws `PreimageNotRecoverableError` out of `parseHashlock` instead. Both refuse and neither claims, so the behaviour is right; the message just never appears. The test asserts the refusal that actually happens.

The remaining e2e gap is #948, stacked on this branch: four Playwright tests for the solver rail's unhappy paths, including the `claimOnchain`-adjacent one where the user walks away mid-quote (#946). It is not a blocker for this PR.

@arkana-ai-bot this leaves nothing open from your review — could you re-review? A `COMMENTED` reply will not clear the existing `CHANGES_REQUESTED`, so the stale review may need dismissing.

Kukks:
Both reviews read. The HD-derivation follow-up is taken; two notes back.

**HD-preimage arm — covered in `5103bb45`.** You were right that only `mustPersistPreimage: true` was exercised. The other arm now has a test: a record stored with no `preimageHex`, against a wallet that cannot re-derive, must reject and never reach `claimOnchainFill`. Mutation: `preimageForSwapRecord(...).catch(() => new Uint8Array(32))` → `promise resolved "{ txid: 'claim-txid', … }" instead of rejecting`. That is the fund-loss shape worth pinning — an L1 claim built on the wrong P burns the window and drops the swap onto the refund path.

I stopped short of asserting a *successful* HD derivation. The derived value would have to be recomputed in the test to know what to expect, which means reimplementing `contractPreimage` beside it; and the derivation is the package's, tested there. What is ours is that `svcWallet` and the record's secrets are the two things handed over, and that a failure to derive refuses — both are now pinned.

**`claimFee` floor — the assertion no longer pins the number.** Your point that 1 sat/vB will not confirm in two blocks under pressure is well made, and it is a policy call I did not want a test to veto. The estimator-down case now asserts against `MIN_CLAIM_FEE_RATE` rather than the literal `1`, so raising the floor is a one-line change that does not read as a regression. Mutation strength is unchanged — a hardcoded rate of `2` still fails it. The decision itself is still yours/@Kukks's to make.

**`payOnchain` — `send()` does not fall through: deliberate, and now observed.** Your read of the contract is right, and #948 (stacked on this branch) pins what the user actually gets. With both rails failing, the screen returns to *Sign transaction* with `Settlement failed: Error: arkade request failed`, the amounts still shown, and *Tap to Sign* available to retry — no stuck spinner, no silent loss. Mutation: dropping the `.catch(handleError)` leaves the spinner up forever, which the test catches. So the current behaviour is "surface it and let the user retry" rather than "silently try the other rail", which I think is the right call while `send()` can throw after partial work — but it is now a pinned, auditable choice rather than a default.

**One correction.** Your "Known gaps" list closes with *"`l1NetworkOf` signet/mutinynet folding filed as #946"*. #946 is the unmount-mid-quote issue; the `l1NetworkOf` fold is not filed anywhere and, per your own finding 2 closeout, is deliberate and documented rather than a gap. Nothing to do — flagging it so the ledger does not carry a gap that is not one.

**Still `CHANGES_REQUESTED`.** Your last three reviews are all `COMMENTED`, and a `COMMENTED` review does not clear a blocking one — `reviewDecision` is still `CHANGES_REQUESTED` at `5103bb45` with all 8 checks green. The `23:15:37Z` review needs dismissing by a human; that is the only thing left between this and mergeable.

Kukks:
Both discussion points answered in `f33d0554`. Comments only, no behaviour change — but the second one turned out not to be the change you proposed.

## 1. `payOnchain` — `send()` can throw after the funding. Not catching is correct.

You framed it as conditional on whether the rail's `send()` throws clean. It does not. Traced rather than reasoned from the comment:

`solverOnchainRail`'s `send()` (`@arkade-os/swap@0.0.12`) does three things in order:

```
assertFundable(...)        pure validation — throws clean, nothing done
await deps.persist(swap)   may write a record; no funds move
await ctx.wallet.send(...) THE FUNDING BROADCAST
```

`ctx.wallet` is the `ServiceWorkerWallet` — typed at `src/providers/wallet.tsx:112`, passed in at `Details.tsx:200`. Its `send()` is a message-bus round trip capped at `DEFAULT_MESSAGE_TIMEOUTS.SEND = 50_000` (`@arkade-os/sdk@0.4.69`, `chunk-TA7XLEP2.js:18622`), and the worker `await`s `this.wallet.send(...)` — submitting — *before* replying `SEND_SUCCESS` (`:17948`). A lost or late reply rejects with `Send failed: …` (`:19898`) **after the covenant is funded**.

That is #949's mechanism, on this exact call. So it is your partial-work case, and the current behaviour is right: falling through to the collaborative exit after a step-3 rejection would spend a second time for the same payment while the solver may still observe the funding and fill the HTLC. The recipient gets paid twice, out of the user's balance.

Worth naming the part that makes this survivable, because it is this PR's own doing: `persist` runs **before** funding, so a rejection at the funding step leaves a record, and `RfqSwapManager` restores and drives the swap on next start. #949's funded-but-unrecorded window does not apply to this leg. That ordering is load-bearing, not incidental.

Comment added at the call site naming the case, so the choice is auditable rather than defaulted.

## 2. `MIN_CLAIM_FEE_RATE` — I am not raising it, and the reason is your own `usableRate` note.

Your read of the risk is right: 1 sat/vB will not reliably confirm in two blocks, and the module's JSDoc says two blocks is the requirement. But "raise it to 2–3" is not a one-line change, because the constant has **two jobs**:

```ts
export const MIN_CLAIM_FEE_RATE = 1
const usableRate = (v) => … && v > MIN_CLAIM_FEE_RATE   // ← the bar, not just the floor
```

It is the fallback rate *and* the threshold a real estimate must clear to be believed. Raising it to 3 would make the wallet reject genuine 1, 2 and 3 sat/vB estimates and pay 3 whenever the mempool is calm — including on every testnet, where the honest rate really is 1. A better fallback needs the two constants split first, and `claimFee.test.ts:24` (`'2': 1` deliberately unusable) pins the current coupling.

So this is a design change to what users pay, not a constant bump, and by the same rule as point 1 it belongs in its own diff with its own review. Three things that lower its urgency:

- The failure is bounded, not fund-loss: `claimOnchainFill` refuses inside `ONCHAIN_CLAIM_MARGIN_SECONDS`, and the Ark lockup still refunds via `refund_locktime`. A stuck claim costs the swap, not the money — as you said.
- A fixed 3 is barely less arbitrary than 1 under real pressure; mainnet has sustained 20–100+ sat/vB for long stretches. The mitigation is real but modest.
- The actual fix is re-fetching on retry, not a better constant.

The reasoning is recorded at the definition so the next reader does not bump it and quietly break estimate acceptance. **This also settles your "cosmetic ambiguity" note on `usableRate`: it is not cosmetic.** The strictly-greater comparison is exactly what makes the constant dual-purpose, and therefore exactly why the obvious fix to the floor is wrong.

Gate after both comments: 93 files / 802 passed / 1 skipped, twice, `format:check` and `lint` clean. The diff is comment-only — verified by filtering the diff for non-comment lines, which is empty.

Kukks:
Finding 5 tracked as #950 — credit to @arkana-ai-bot for raising it.

The issue records what the cast papers over (`RouterContext.wallet` is the concrete `Wallet` class at `index.d.ts:6135`, while what this app has is a `ServiceWorkerWallet` that *implements* `IWallet` but does not extend `Wallet`), what breaks if the shape drifts (a member on `Wallet` but not on `ServiceWorkerWallet` throws at runtime inside a send, invisible to `tsc` because both hops go through `unknown`), and that the real fix is upstream — typing `RouterContext.wallet` as `IWallet`, the interface the rails already restrict themselves to. No existing ts-sdk issue to link; I searched.

Two things I checked rather than assumed while writing it:

**The second cast is not load-bearing.** `ctx.wallet as unknown as IWallet` at `sendRouter.ts:47` can just be `ctx.wallet` — it is already typed `Wallet`, and `Wallet implements IWallet`. With the cast removed, `tsc --noEmit` reports the same single pre-existing error (`lnSendRecords.test.ts(33,7)`) it reports with it. Not changing it here; noted in the issue as the half that is fixable without an SDK change.

**It is not inherited.** `src/lib/sendRouter.ts` does not exist on `master` — it arrives with this PR (`f2340385`), which is the first production code here to construct a `PaymentRouter`. The only comparable `as unknown as IWallet` on `master` are in test files building fake wallets. The *constraint* is pre-existing; the casts are not. Not a reason to hold this PR — the cast is the honest encoding of what the SDK demands — but the tracking should not record it as something that was already there.

Nothing else outstanding from either of your reviews.

Kukks:
Correcting my previous comment: I got finding 5's provenance wrong, and #950 has been fixed.

I said @arkana-ai-bot had marked finding 5 "No longer actionable" and that I filed the issue anyway. That was true of its 21:20 and 23:15 passes but **not** of its latest: review `5120650991` (2026-09-05T09:14:47Z) reopens it as *"Still open"* and asks directly — *"Low risk; not a blocker, still worth a follow-up issue to track the drift surface."* My mistake was mechanical: I grepped a hardcoded list of review IDs captured earlier in the session instead of re-fetching, so the 09:14 review could not have matched. #950's Provenance section now quotes the current ask rather than the superseded closure.

The lesson generalises past this one finding, so it is recorded in #950 too: **that ledger is unreliable in both directions.** Earlier on this PR findings 1–4 were marked fixed at `e0111f1c`, where the diff shows they were not yet applied (self-corrected in the next pass); now a finding has been self-closed and then reopened. Neither a carried "fixed" nor a carried "no longer actionable" is load-bearing — the newest review plus the code is the only reliable pair.

The other two corrections in that comment stand and are unchanged in #950: the second cast (`ctx.wallet as unknown as IWallet`, `sendRouter.ts:47`) is dead — `tsc --noEmit` reports the same single pre-existing error with and without it — and the casts are **not** inherited, since `src/lib/sendRouter.ts` does not exist on `master`.

One note on the review's own currency: `5120650991` is stamped at `f33d0554` but its incremental section assesses `5103bb45`, so it has not yet read that commit. Its remaining open point — *"Raising the floor remains an open policy call … worth a follow-up"* — is answered there: `MIN_CLAIM_FEE_RATE` is simultaneously the fallback value **and**, via `usableRate`'s `v > MIN_CLAIM_FEE_RATE`, the bar a real estimate must clear, so raising it to 3 would make the wallet reject genuine 1–3 sat/vB estimates and overpay whenever the mempool is calm. That is a behaviour change on what users pay, not a constant bump, and it is recorded as the maintainer's call for its own diff. The reasoning is at the definition site in `src/lib/claimFee.ts`.

Kukks:
Force-pushed onto `ts-sdk-0.5` (`9bc14c7e` → `ef10dcee`). The 12 original commits are replayed unchanged; `ef10dcee` carries the whole API translation. The base was not changed and nothing was merged. Full rationale is in the updated description.

Three things reviewers should know:

**1. CI does not run on this PR.** `ci.yml` and `playwright.yml` are both gated on `pull_request: branches: [master, next]`. With the base on `ts-sdk-0.5`, neither the unit suite nor the e2e is triggered — the green checks here are the two Cloudflare Pages previews and a CodeRabbit check that **skipped** (`Review skipped: reviews are disabled for this base branch`). `mergeStateStatus: CLEAN` carries no quality signal on this base.

The gate was run locally instead, against a baseline recorded on `origin/ts-sdk-0.5` (`d7923437`) before any change:

| | Baseline | This branch |
| --- | --- | --- |
| `npx tsc --noEmit` | 0 errors | 0 errors |
| `pnpm test:unit` | 86 files, 699 passed, 1 skipped, 0 failing | 91 files, 765 passed, 1 skipped, 0 failing |

Adding `ts-sdk-0.5` to those two workflow triggers would fix it, but that is a repo-wide CI change on a release branch and belongs to #936 rather than to a feature PR — happy to do it if you want it here.

**2. Prior findings against the deleted code.** CodeRabbit's `lnSwaps.tsx:195` finding (*"reject the reservation until the swap manager is ready"* — the rail could fund an HTLC whose record was not durable) is resolved by construction: `reserveOnchainSend` is gone and `accept()` persists before it funds, inside the corridor. Finding 5 / #950 is still real — `RouterContext.wallet` is still the concrete `Wallet` on `@arkade-os/sdk@0.5.0-rc.2`, so the double cast is preserved unchanged and stays tracked there.

**3. One deliberate behaviour change.** `SwapsContext.sendRouter` propagates `SwapsHeldElsewhere` rather than building a router without the two solver rails. In a second tab, a plain on-chain send therefore reports "another tab is handling swaps" instead of silently offboarding through the costlier collaborative exit. That is the honest message, but it is a change — say so if you would rather it degraded quietly.

Kukks:
Correcting point 2 of my previous comment, and re-pushed for it: `ef10dcee` → `542be6cb`.

I said the #950 double cast was "preserved unchanged". That was too broad — there are two *kinds* of cast on this branch and only one is forced.

**Removed** (not load-bearing): `collaborativeExitWithFees(ctx.wallet, …)` in the exit rail and `sendAssets(ctx.wallet, …)` in the asset rail. `RouterContext.wallet` is already the concrete `Wallet`, which satisfies `IWallet`, so both assign directly. Verified on this base rather than carried from the issue — `ts-sdk-0.5` is not the master #950 was filed against — and `npx tsc --noEmit` stays at 0 with both gone. The asset-rail one arrived with #947 and is not named in #950, but it is the same cast.

**Kept** (forced): the `PaymentRouter` constructor hand-off. `RouterContext.wallet` is typed as the concrete `Wallet` class while this app holds a `ServiceWorkerWallet` — which implements `IWallet` but does not extend `Wallet` — so it is nominally illegal with no narrower cast available. @arkana-ai-bot raised this twice and the previous comment named no upstream reference, which is why it did not close; the comment at the call site now states the nominal-typing reason and points at #950. It can go once the SDK types `RouterContext.wallet` as `IWallet`, which is being changed upstream in `arkade-os/ts-sdk`.

Gate unchanged after the amend: `npx tsc --noEmit` 0 errors, `pnpm test:unit` 91 files / 765 passed / 1 skipped / 0 failing, lint and `format:check` clean. CI still does not run on this base — see point 1 above.

Kukks:
Re-pushed: `542be6cb` → `298ec507`. Thanks @arkana-ai-bot for confirming the finding 5 split — including the two further casts inside the bespoke `lightningRail` that went with the code rather than individually. That matches what I removed.

**This push fixes a real bug I introduced in the first version of this port, in the onchain claim dep.** Flagging it rather than folding it in quietly, because it is the fund-loss-shaped one.

The dep read the stored record with:

```ts
const record = await assetSwapRepository.getRfqSwap(swap.rfqId)
```

`IndexedDbAssetSwapRepository.getRfqSwap` is the **v1 `rfqSwaps` store**. The v2 client does not use it: `corridorRecordStore` (`dist/index.js:1480-1520`) resolves a record by `rfqId` through `getAllSwapRecords()`/`getSwapRecord()` and writes with `saveSwapRecord()`. Nothing in the client calls the repository's own `saveRfqSwap`, so that store is never written.

So the read would have returned `undefined` for every swap and the claim would have thrown `no stored record for rfq …` — **after the HTLC was funded**, which is precisely the state the dep exists to avoid. It now reads through the package's own `corridorRecordStore` over the same repository the client is given.

The first version of the test did not catch this because it mocked `../../lib/swapRepository` with a `getRfqSwap` the real repository would never have been asked for — a mock that made the wrong call look right. The fake now implements the **v2 record API only** (`getAllSwapRecords`/`getSwapRecord`/`saveSwapRecord`) and uses the real `corridorRecordStore`; reverting the fix fails 5 tests with `TypeError: assetSwapRepository.getRfqSwap is not a function`.

Gate after the fix: `npx tsc --noEmit` 0 errors; `pnpm test:unit` 91 files / 766 passed / 1 skipped / 0 failing (baseline on `d7923437` was 86 / 699 / 1 / 0); lint and `format:check` clean.

Still unverified: no live regtest run. The record-read path is now pinned against the same store API the client writes, but a real claim against a real funded HTLC has not been observed — and CI does not run on this base branch, so nothing here was checked by the repo's own gate.

Kukks:
Re-pushed: `298ec507` → `743701b5`. One test added, closing the gap I named as the thing most likely to be wrong.

The claim dep reads a record through `corridorRecordStore` and hands it to `rfqClaimSecretOf`. Both were mocked, so nothing checked that the two packages agree on the record's shape — and a disagreement there fails **only after the HTLC is funded**, which is the same failure class as the store bug in the previous push, one layer deeper.

There is now a test that runs both for real (`vi.importActual`): a v2 `SwapRecord` goes through the real `corridorRecordStore.getRfqSwap`, and the real `rfqClaimSecretOf` reads the projection it returns. It resolves to `{ signingDescriptor, paymentHash, preimageSaltHex }`, so `rfqRecordOf`'s projection does carry what `OnchainSendCorridor.claimSecret` needs (`profile.signer` and `profile.hashlock`). If a future package version reshapes that projection, this fails at CI rather than at a claim.

Gate: `npx tsc --noEmit` 0 errors; `pnpm test:unit` 91 files / 767 passed / 1 skipped / 0 failing (baseline on `d7923437`: 86 / 699 / 1 / 0); lint and `format:check` clean.

What is still unverified is the part a stack is genuinely needed for: no live regtest run, so no real solver negotiation, no real funding, and no real L1 claim. And CI does not run on this base branch, so none of the above was checked by the repo's own gate.

Kukks:
Re-pushed: `743701b5` → `283a1ed2`. Two tests, narrowing the last risk I named.

`onchainSwapRail` asks the solver for `amount + claimFee` on the take leg and prices the quote off whatever comes back. Nothing here tested a solver whose answer differs from the request, which is the case where the recipient could be quietly short-paid.

- **Short take leg.** `RouteQuote.amount` is the post-claim-fee payout, so a solver returning 1,000 sat less makes it 49,000 against a 50,000 screen — and `quoteIsForThisSend` refuses. The rail is skipped and the collaborative exit ranks instead, which is the correct outcome: no silent short-pay, no funded lockup.
- **Legs that do not add up.** A quote violating `give == take + fee` is refused by `receiverExact` at `quote()`, before a lockup exists.

Both now asserted. Gate: `npx tsc --noEmit` 0 errors; `pnpm test:unit` 91 files / 769 passed / 1 skipped / 0 failing (baseline on `d7923437`: 86 / 699 / 1 / 0); lint and `format:check` clean.

The residue is genuinely stack-shaped and I am not claiming it: whether the HTLC a real solver locks on L1 matches the take leg it quoted. The wallet-side guards above fire on a bad *quote*; they cannot see a fill that disagrees with a good one. That needs a regtest run — and note CI does not run on this base branch, so none of this was checked by the repo's own gate.

Kukks:
@arkana-ai-bot — all three e2e findings taken, pushed as `283a1ed2` → `059d611c`. One correction to the finding-5 row, and one disagreement with the reasoning behind finding 3 (the change is still made).

**1. `.find()` before `.toMatch()` — taken.** Correct and worth having: `Array.find()` is `string | undefined`, so a missing log line surfaced as `TypeError: received.match is not a function` rather than a readable failure. Now:

```ts
const refused = log.find((m) => m.includes('onchain-swap could not quote'))
expect(refused).toBeDefined()
expect(refused).toMatch(refusal)
```

Worth noting this is the line I had already edited in this port — the rail id moved `solver-onchain` → `onchain-swap` — so the fragility was in code I touched and should have caught.

**2. `docker exec -t` — taken.** Dropped `-t` on both the `beforeAll` and `afterAll` calls. One-shot scripted commands that read no stdin have no use for a pseudo-TTY, and CI runners have none.

**3. 15 s balance wait — taken, but not for the reason given.** Raised to 30 s on both sites. Your supporting claim ("previous runs of similar waits in the suite use 30–60 s") does not hold: across `src/test/e2e`, `timeout: 10000` is used 24 times and is by far the most common, `30000` 6 times, and the only two `15_000` uses are the two in this file. The argument that *does* hold is in-file — the sibling balance assertions in this same file allow **120 s** (`Settlement failed`, and `${FUNDED - SENT} sats available`), so 15 s for a balance row arriving on a service-worker round-trip is inconsistent with its own neighbours. Same conclusion, sounder basis.

**Correction to the finding-5 row.** It says what remains is "the one in `createSendRouter` … and the two in `walletExitRail`/`assetRail` (`ctx.wallet as unknown as IWallet`)". Those last two are already gone — removed in `542be6cb`, which you confirmed yourself in the comment above ("Removed (not load-bearing) … `collaborativeExitWithFees(ctx.wallet, ...)` … `sendAssets(ctx.wallet, ...)`"). In the tree you reviewed:

```
$ git grep -n "as unknown as" 283a1ed2 -- src/lib/sendRouter.ts
283a1ed2:src/lib/sendRouter.ts:99:    wallet: deps.wallet as unknown as ConstructorParameters<typeof PaymentRouter>[0]['wallet'],
```

One cast, not three. The ledger regressed between your two passes — flagging it because the same row will otherwise carry forward again.

**What I cannot claim.** These three changes are in `send-onchain-fallback.test.ts`, which I have not executed: there is no regtest stack here, and — as noted above — Playwright does not run in CI on this base branch, since `playwright.yml` is gated on `branches: [master, next]`. They typecheck, lint and format clean, and that is the whole of the verification behind them.

Unit gate unchanged: `npx tsc --noEmit` 0 errors; `pnpm test:unit` 91 files / 769 passed / 1 skipped / 0 failing.

Kukks:
The base moved while this was open — `ts-sdk-0.5` went `d7923437` → `f1a302a1` (*"resolve one network for discovery, so an unnamed one keeps no solver cards"*). Re-rebased and re-gated against it: `059d611c` → `7d915d43`.

GitHub still called the PR mergeable across that move, but the base was no longer an ancestor of the head, so the tested tree and the merge preview were not the same tree. Rebasing restores both.

**No conflicts** — that commit touches `src/lib/swapMarkets.ts` and its test, and this branch touches neither. `discoveryOptions` keeps its signature (the fix is internal to how it narrows the network), so `swapClient.ts` and `Form.tsx`, which are its two consumers here, are unaffected.

Re-measured against the new base rather than carrying the old numbers forward:

| | Baseline `f1a302a1` | This branch `7d915d43` |
| --- | --- | --- |
| `npx tsc --noEmit` | 0 errors | 0 errors |
| `pnpm test:unit` | 86 files, 701 passed, 1 skipped, 0 failing | 91 files, 771 passed, 1 skipped, 0 failing |

The +70 is unchanged from the previous base; 699 → 701 is the two tests `f1a302a1` itself added. Lint clean.

**One thing worth flagging about the base.** Its first full run here failed 4 tests in `src/test/screens/wallet/receive-lightning-error.test.tsx`. They pass 4/4 in isolation, and the next full run of the same base commit was green — so it is a load-dependent timing flake on `ts-sdk-0.5`, not a regression, and not something this branch introduces. Recording it because a rebase that happened to inherit that run would otherwise read as this PR breaking the receive screen. Same class as the `swap.test.tsx` flakes already known in this suite.

Kukks:
Base moved again — `f1a302a1` → `bd507474`, the **rc.3 bump**. Re-rebased and re-gated: `7d915d43` → `a91f3b3d`, plus one new commit `7710aeb9`.

**The bump moved two of this branch's imports.** `bd507474`'s own note in `swapRecords.ts` warns that `./advanced` "is explicitly not a compatibility promise — its names move with the client's internals across minor versions", and that is exactly what happened here: `corridorRecordStore` and `type OnchainClaim` are now on `@arkade-os/swap/advanced`. Everything else this branch imports stayed put — the claim primitives on `/protocol`, the two rails on the root. `tsc` caught it as 5 errors on the rebase; there was no file-level conflict, so a merge would have been clean and the build broken.

**The suite's timeout margin — and this branch is what makes it fire.**

`vitest.config.ts` set no `testTimeout`, so the 5 s default applied. Measured:

| | result |
| --- | --- |
| `bd507474` (base), 3 runs | 3/3 green — 701 passed |
| this branch before `7710aeb9`, 5 runs | **4 of 5 red** — always the same two `swap.test.tsx` tests, always `Test timed out in 5000ms`, never an assertion |
| `swap.test.tsx` alone | 40/40 green |
| this branch after `7710aeb9`, 3 runs | 3/3 green — 771 passed |

The fragility is pre-existing — the base itself hit the same wall in `receive-lightning-error.test.tsx` (4 tests, green on re-run) while I was measuring it, which is what says the *margin* is short rather than either test being broken. But five new test files raise parallel load enough to push those two over, so this branch is the proximate cause, and "pre-existing flake" is not a reason to hand you a suite that fails four runs in five.

`7710aeb9` sets `testTimeout: 15_000` — margin, not masking; a genuinely hung test still fails. It is config rather than a per-test timeout because the two files that have hit it are *different* files. It is a **separate commit** precisely because it is the one change here touching shared config rather than this feature: take it, move it to #936, or drop it, independently of the port.

Gate on the new base: `npx tsc --noEmit` 0 errors; `pnpm test:unit` 91 files / 771 passed / 1 skipped / 0 failing against a baseline of 86 / 701 / 1 / 0 on `bd507474`; lint and `format:check` clean.

Kukks:
Pushed `748330df` — fixes the claim-fee finding this PR previously reported as out of scope, rather than carrying it forward.

`src/lib/claimFee.ts` asked for Esplora's `/fee-estimates`, which every Arkade mempool deployment 404s. The catch returned `MIN_CLAIM_FEE_RATE`, so every L1 claim was priced at the 1 sat/vB floor on every network, mainnet included, silently.

The headline evidence is runtime rather than a test assertion — the real `claimFeeRate` run against the live hosts:

| base URL | before | after |
| --- | --- | --- |
| `https://mempool.arkade.sh/api` (**mainnet**) | 1 sat/vB | **3 sat/vB** |
| `https://mempool.arkade.sh` (no `/api`, the SPA decoy) | 1 sat/vB, silent | 1 sat/vB, **logged** |

mutinynet, signet and mempool.space testnet stay at 1 sat/vB after the fix. That is **not a partial result** — all three report `fastestFee: 1` right now, so 1 is the correct answer on a calm network. Before the fix they read 1 for the wrong reason.

**CI does not run on this base.** `ci.yml` and `playwright.yml` gate on `master`/`next`, so a green/`CLEAN` state on this PR carries no signal about the code. The local gate is the only evidence:

- `npx tsc --noEmit`: 0 errors → **0 errors**
- `pnpm test:unit`: 91 files, 775 passed, 1 skipped, 0 failing → **91 files, 780 passed, 1 skipped, 0 failing** (+5 = `claimFee.test.ts` going 5 → 10 cases)
- `lint` / `format:check` / `csp:check` clean; pre-commit hook ran, no `--no-verify`
- Four mutations run, four killed, none survived

The SDK's own `EsploraProvider.getFeeRate()` has the identical bug against the identical hosts, differing only in that it throws instead of flooring. This repo never calls it, so it is unreachable here — filed as arkade-os/ts-sdk#858 for other consumers.

Full write-up, the mapping argument, and the mutation kill messages are in the PR description under `748330df`. Not merging.

Kukks:
Pushed `39404b9b` — a regression in my own `748330df`, caught before review.

`748330df` moved `baseUrl.replace()` outside the try block, so a base URL the endpoint map could not name threw out of `claimFeeRate` instead of falling back to the floor, leaking the abort timer with it. The original code caught that; the guarantee is restored and the cause is now logged rather than swallowed. Removing the restored `catch` fails the new test with `TypeError: Cannot read properties of undefined (reading 'replace')`.

This came from re-reading arkana-ai-bot's review of `7710aeb9` against the *new* code rather than the old one it was written about. Answering that observation directly, since I partly disagree with it:

> `aspInfo.network as NetworkName` cast: if the network is unknown (e.g. 'mutinynet'), `ESPLORA_URL[network]` returns undefined, `claimFeeRate` falls back to `MIN_CLAIM_FEE_RATE = 1`.

The example is wrong — `mutinynet` **is** in `ESPLORA_URL`, as `https://mempool.mutinynet.arkade.sh/api`; I probed it live and it answers. So it is not a case of an unnamed network.

The underlying point about the unchecked cast was right, though, and it is what found the bug: for a network genuinely outside the map, `undefined` reaches `claimFeeRate`, and as of `748330df` that threw rather than flooring. Throwing there blocks the send router outright, where the floor would merely have made the claim slow — strictly worse than the behaviour I was fixing. Fixed in `39404b9b`.

Local gate, since CI does not run on this base: `npx tsc --noEmit` 0 errors; `pnpm test:unit` 91 files, **781 passed**, 1 skipped, 0 failing (baseline `49124dff` was 775 passed). Pre-commit hook ran; no `--no-verify`.

The blocking review on `7710aeb9` names a different item — the test-4 balance assertion in `send-onchain-fallback.test.ts` — which is outside this commit's scope and untouched by it.

Kukks:
### Three send fixes, and a note on what verified them

Pushed `d69dd776`, `f172ef37`, `feff7a0c`.

**CI does not run on this branch.** `ci.yml` and `playwright.yml` both gate on `pull_request: branches: [master, next]`, and this PR targets `ts-sdk-0.5`. So the `MERGEABLE` state and the absence of red checks carry no signal here — the only evidence is the local gate: `tsc --noEmit` 0 errors, `pnpm test:unit` 92 files / 792 passed / 1 skipped, exit 0, against a recorded baseline of 91 files / 781 passed / 1 skipped, exit 0.

**1. Lightning sends were refused for a reason that was never checked (`d69dd776`).**
`lightningRail.available()` forwards `req.amount` into `client.resolve()`. An amount-bearing invoice already pins the take leg, so `@arkade-os/swap@0.1.0-rc.3` throws `AmountMismatch` — *even when the two amounts are identical*. Confirmed by driving the SDK's own `resolveRoute`:

```
WITH amount    -> THREW AmountMismatch: exactly one amount may be pinned; got the invoice's 2100 and amount 2100
WITHOUT amount -> OK  amount={"value":"2100n","on":"take","source":"destination"}
```

`PaymentRouter.options()` warns that error to the console and drops the rail, so every amount-bearing invoice lost its only route. This was invisible to the unit suite because the fake client's `resolve` answered `eligible: 1` to anything and never ran the real resolution; the tests now drive `resolveRoute` directly.

The message compounded it. `lnSendRefusal` returned "Amount outside solver bounds" whenever *any* lightning market existed — it performed no comparison at all, so the bounds it named were never the reason. It now names bounds only where it has checked them and none admit the amount, spans every lightning card rather than the first, and otherwise says the router does not report the reason.

The reason cannot be plumbed through `options()` without an SDK change: it returns `{railId, quote}` only, and the rail's error is logged then discarded.

**2. A destination-named amount could be overwritten (`f172ef37`).**
The sats readout drifted because the field text is rendered to two fiat decimals and `InputAmount` re-derived sats from that text; at the reported rate one cent is ~13 sats, so the round-trip cannot be lossless. Separately, the "N available" control stayed a tap target while the amount was pinned — a tap wrote the whole balance over the invoice's amount (reproduced in test: 1,000,000 written for a 2,100-sat invoice).

**3. The on-chain preview discarded why it could not price a rail (`feff7a0c`).**
`previewOnchainCost` swallowed every quote failure, on the stated assumption that the spend "walks the same rails moments later" and would log it. That is exactly what fails when the spend reaches a *different, successful* answer, which is the reported `quote pays 22000 for 22709, screen shows 22000 for 22000`. It now names the rail it could not price.

**Still open — a product decision, not a bug.** In that report the guard behaved correctly: it refused a rail asking more than the screen showed. What followed is the question — the send fell through to the collaborative exit, a different rail with different cost, privacy and batch-timing properties, with no re-confirmation and a fee row reading zero. The new screen test characterises that sequence rather than blessing it. My recommendation is that a send should re-price and re-confirm rather than change rails under the user, but that changes behaviour an existing test deliberately pins, so it should be decided rather than slipped in.

Kukks:
Thanks — findings **A** and **B** were both real and both mine, and B was the better catch of the two. Fixed in `f1316ecd`, details at the bottom.

The six carried findings are a different story. I checked each against the commit that was reviewed (`dbfc90b9`) rather than against the summary, and all six are already fixed or refer to code that no longer exists. Evidence per finding:

### 1. Mainnet solver absence — **fixed before the reviewed commit**

`onchainClaimEndpoint` does not read `explorers.ts`. At `dbfc90b9`:

```
$ git show dbfc90b9:src/lib/onchainPayout.ts
/** The SDK's map, not `explorers.ts`: that one builds explorer LINKS, has no
 *  mainnet `api`, and reading it here is what kept the solver rail off mainnet. */
export const onchainClaimEndpoint = (network: NetworkName): string => ESPLORA_URL[network]
```

`ESPLORA_URL['bitcoin']` is `https://mempool.arkade.sh/api`, read out of the installed `@arkade-os/sdk@0.5.0-rc.4`. Fixed by `8aacb5da` ("fix(send): the solver rail was reading the wrong endpoint map"), and:

```
$ git merge-base --is-ancestor 8aacb5da dbfc90b9 && echo YES
YES
```

So the rail **is** registered on mainnet. The diagnosis was right about the original bug and it is the reason `8aacb5da` exists — but it was already in the tree being reviewed.

The hazard it describes is real enough to guard, though, so `f1316ecd` adds `src/test/lib/onchainClaimEndpoint.test.ts`: the endpoint must resolve for **every** network in `networks`, not just the one that broke. Pointed back at `explorers.ts` it fails on `bitcoin` alone — which is exactly the network that map lacks.

### 2. `l1NetworkOf` fallthrough — comment is present

Same file, same commit, three lines above:

```
/** Deliberate, not a fallthrough: signet and mutinynet derive HTLC scripts from
 *  the same parameters as testnet, and `OnchainNetwork` names only the three. */
```

### 3. Error degradation when `utxoTxsAllowed()` is false — meaningful message present

`src/screens/Wallet/Send/Details.tsx:285` at `dbfc90b9`:

```ts
// Blanked by the limit, not by routing: every rail fails `quoteIsForThisSend`.
if (!details.destination) return handleError('On-chain sends are not permitted on this account')
```

The user does not see "No route for this payment" on that path.

### 4. `reserveOnchainSend` comment — the file does not exist

`src/providers/lnSwaps.tsx` was deleted in the v2 client migration. Its work lives in `src/providers/swaps.tsx`.

### 5. `wallet` double cast — removed

Removed in `1a118d2a`, once ts-sdk#851 landed in `0.5.0-rc.4` and typed `RouterContext.wallet` as `IWallet`. The line is now just `wallet: deps.wallet,`. Closes wallet#950.

### 6. No abort path on unmount mid-quote — argued, not ignored

Recorded in `ac7684da`. Briefly: none of `router.options()`, `option.quote()` or `quote.send()` accepts an `AbortSignal`, so nothing can actually be cancelled — only its result ignored. The `live`-flag precedent in this file guards a read-only price preview; applying it to the send would add a new behaviour on the money path (silently dropping a payment the user authorised by pressing Sign) without closing the window, since unmount can land a tick after the check. React is 18.3.1, so there is no setState-on-unmounted warning, and `handleSent` does its durable `saveTransactionActivityMetadata` write before any setState. Happy to revisit if the SDK grows a signal.

---

### The diagnostic worth flagging

Every line number cited above is out of range for the file it names: `onchainPayout.ts:88` in a 13-line file, `sendRouter.ts:317` and `:350` in a 216-line file, and `lnSwaps.tsx:498` in a file that no longer exists. That is consistent with a pre-v2 ledger being re-asserted rather than re-read — worth knowing, because "0 of 6 prior findings fixed" reads as a much worse signal than the tree supports.

That said: the two **new** findings on this same review were both correct and both mine, and A in particular came from a diagnostic I added on purpose and got wrong. The carried list needs re-reading; the fresh review does not.

### What changed in `f1316ecd`

**A** — `lockupTxids` now takes the record's terminality. A receive's lockup exists from the moment it is derived and holds nothing until the sender pays, so an empty read is a normal transient state and is silent; only an **ended** swap with nothing at its lockup is anomalous and loud. Exactly your point: an error that fires routinely is how the real one gets ignored.

**B** — correct, and it had a second half nobody flagged. Resetting the ref per client fixes the network-switch replay, but the flag was per **tab** and only the lock holder ever set it, so a follower tab never announced anything at all — the opposite of what the provider documents ("Every tab announces, deliberately"). So the replay bit now travels on the `DriverUpdate` itself: the driver marks its own replay and publishes the mark, and a follower applies state either way but announces only what is news.

My first test for B was worthless and I only caught it by mutation — it used unmount/remount, which creates a fresh ref, so it passed with the fix removed. It now rerenders with a changed network to re-run the effect in the same instance, and correctly fails without the fix.

Gate: 97 files / 854 passing, 0 failing, `tsc` clean.

PR #947, opened 2026-09-05
feat(send): route lightning and asset sends through PaymentRouter too

Kukks:
Stacked on #944 — base is `feat/payment-router-send`, so the diff here is only the second step. Merge #944 first.

#944 put the on-chain send behind `PaymentRouter`. The other two send paths still chose their provider by hand, so the sign screen had three shapes and the UI had nowhere uniform to read options from. This routes Lightning and asset sends through the same engine.

## What routes now

| path | before | after |
| --- | --- | --- |
| on-chain BTC | `solver-onchain` / `onchain` (#944) | unchanged |
| lightning | `lnSendRendezvous` + `requestLnSend` inline in the send form | `lightning` rail, picked by `options()` |
| assets | `sendAssets()` called directly on the sign screen | `asset` rail, picked by `options()` |
| ark sats send | `sendOffChain()` | still direct — see below |

## Both rails are wallet-local, deliberately

`@arkade-os/swap` ships `solverLightningRail` for exactly this corridor. It was the obvious candidate and it was rejected, for the same class of reason `walletExitRail` exists rather than the SDK's `onchainRail`:

- it calls `persist` **before** funding and its `RouteResult` carries no txid, so the record would be written without `fundingArkTxid`. `viewOf` in `lnSendRecords.ts` returns `undefined` for such a record, so it never reaches `lnSendViews` — and `ungroupedLnSendTx` is the *only* source of the row for a send Arkade's own history does not report, keyed on `send.fundingTxid` as its `redeemTxid`. The Lightning row would vanish.

  (To be exact about the neighbouring claim: `rfqSwapActivityInputs` *prefers* `record.fundingArkTxid` but is not broken without it — `activityInputOf` falls back to an indexer lookup by lockup script when it is absent. The row-disappearing problem is `viewOf`, not the grouping.)
- it selects markets via `selectMarkets`, not `lnSendRendezvous`. On a live money path a different provider pick is a regression whether or not it is a better one.

So the `lightning` rail wraps the calls that are live today — `requestLnSend` → `sendOffChain` → `trackLnSend`, in that order — and the `asset` rail wraps `sendAssets`. A `PaymentRequest` carries sats only, so the asset list rides the rail's deps.

If/when the SDK rail grows a funding txid and this wallet's market selection, swapping it in is a `use()` call.

## What is preserved, and how it is held

- **Which solver answers.** The rail calls the same `lnSendRendezvous` with the same per-network emulator pin, including the rule that a card disagreeing with the pin is skipped.
- **Both refusal messages.** `options()` reports absence without a cause, so `lnSendRefusal` tells "No Lightning solver available" from "Amount outside solver bounds (min-max sats)" on the form's error path. Discovery is cached, so the second lookup is free.
- **Negotiation errors.** `quote()` is called unguarded on the Lightning path — an `InvoiceRejected` ("invoice has expired", "invoice is not for regtest") or an `AddressMismatch` reaches the user verbatim rather than collapsing into "no route". The on-chain path's skip-to-the-next-rail loop is correct there because it has a fallback rail; Lightning has none.
- **Record order and contents.** Funded first, recorded second, with the funding txid — the opposite of the on-chain leg, and the reason is unchanged: a `trackLnSend` rejection is reported, not raised, because funding IS acceptance.
- **The expiry gate.** Moved into the rail's `send()`, same message, still refusing before anything is funded.
- **Failure behaviour at `send()`.** arkana asked on #944 whether `payOnchain` should fall through to the next rail when `send()` throws. The question does not arise on either new path: Lightning and assets each have exactly one rail, so there is nothing to fall through to, and `payLightning` / `payAssets` let the rejection reach `handleError` — which is what the code they replace did. Specifically, `sendOffChain` throwing after a broadcast leaves a funded, unrecorded covenant; that hazard is unchanged, because the record was written after a successful send before this PR too.

## What changed on purpose

- The negotiation stays **on the form** (the fee has to be on screen before the user signs), but now goes through `options()` + `quote()`. `SendInfo.pendingLnSend` carries the `RouteQuote` instead of an `LnSendRequest`, and the sign screen calls its `send()`.
- `quoteIsForThisInvoice` is the Lightning analogue of #944's wrong-address guard: a quote negotiated for another invoice is refused at the sign screen, not just cleared when the form's invoice changes.
- `createSendRouter` registers a rail only when the deps it needs are present. That is what keeps the Lightning path from deriving an L1 payout pubkey it will never use — a service-worker round trip and a new failure mode on a path that had neither.

## Not in scope

- **The plain ark sats send** (`sendOffChain` to an ark address) is still direct. The SDK's `arkRail` would cover it, but it and the new `asset` rail both match an ark address, so registering both needs a ranking decision — and this PR was asked for Lightning and assets.
- **`smartSetError` in `Form.tsx` takes `string` but is used as a promise `.catch` handler** (`getReceivingAddresses(...).catch(smartSetError)`), so a rejection there puts an `Error` object into React state and `ErrorMessage` renders it as a child — the send form white-screens. Pre-existing, found while building the test harness (which is why the form test supplies a wallet whose addresses resolve). Not touched here.

## Test plan

Baseline re-recorded at `f33d0554` (this PR's base, after its two new commits): **93 files, 802 passed, 1 skipped, 0 failed**.
After: **94 files, 836 passed, 1 skipped, 0 failed** — +1 file, +34 tests, nothing lost.

(The earlier figures in this PR's history were against base `529c0e18`: 93/801 → 94/835. The base has since advanced; both baselines were measured by checking the base out and running the suite, not inferred.)

One pre-existing flake seen twice on full runs and never twice in the same place — `useBounceMorph`'s `controls.start() should only be called after a component has mounted`, and a 5s timeout in `swap.test.tsx`. Both pass in isolation and on a re-run; neither is in a file this PR touches.

`pnpm lint`, `pnpm format:check`, `pnpm build` and `npx tsc --noEmit` all match baseline (tsc has one pre-existing error in `src/test/lib/lnSendRecords.test.ts`).

Every rail-selection decision was mutation-tested — the mutation applied, the suite run, the failing test and its message recorded, the mutation reverted. 15 mutations, 15 caught:

| mutation | caught by |
| --- | --- |
| ln `available`: drop the bounds check | drops itself for a size the one solver will not take |
| ln `available`: never drop the rail | no card / wrong size / discovery throws |
| ln `match`: claim every target | takes an invoice and leaves the other targets alone |
| ln `send`: record before funding | records the swap only once the covenant is funded |
| ln `send`: record without the funding txid | funds the lockup and records the swap WITH its funding txid |
| ln `send`: no expiry gate | refuses to fund an expired quote, and funds nothing |
| ln `send`: a refused record fails the send | does not fail a funded send because the store refused the record |
| `quoteIsForThisInvoice`: always true | the wrong-invoice guard + never funds a route negotiated for a DIFFERENT invoice |
| `lnSendRefusal`: branches swapped | both refusal tests + both form screen tests |
| asset `available`: send with nothing to send | drops itself when there is nothing to send |
| asset `match`: claim every target | takes an ark address and nothing else |
| Details: carry the quote amount, not its total | carries the quote's total, not the amount typed |
| Details: skip the wrong-invoice guard | never funds a route negotiated for a DIFFERENT invoice |
| Details: asset send routed to a held address | routes to the ark address the screen is showing |
| Form: quote discarded instead of carried | carries the route the Lightning rail quoted to the sign screen |
| guard: drop the defined check | refuses when the screen carries no invoice at all |
| never register the solver on-chain rail | ranks the solver ahead of the collaborative exit |
| register the exit rail unconditionally | takes an invoice and leaves the other targets alone |
| Form: ask the registry a second time on the refusal path | discovers the cards once… / still names the refusal when the registry is unreachable |
| Form: rail fetches its own cards instead of the snapshot | hands the rail the cards it already fetched |

### Not covered by CI — and the gap is wider than one file

`src/test/e2e/lnsend.test.ts` is the only end-to-end proof that a Lightning send actually pays an invoice, and it **has never run in CI at all**. `.github/workflows/playwright.yml` enumerates e2e files by name with no glob, and the lists have drifted from the directory in both directions:

- on disk but in **no** job: `lnsend.test.ts`, `solvers.test.ts`, `swaps.test.ts`
- named in the workflow but matching **nothing**: `restore.test.ts`, `swap.test.ts` (singular)

Verified with `npx playwright test --list`: `swap.test.ts` resolves to **0 tests in 0 files** — the args are regexes, and `swap.test.ts` does not match `swaps.test.ts` (there are two characters, `s.`, where the pattern allows one). `swaps.test.ts` resolves to 4 tests, none of which run today. And a non-matching argument alongside a matching one is dropped silently with exit 0 — which is why the jobs stay green and the drift is invisible. This is not a problem this PR introduces, and I am deliberately **not** editing the workflow here — someone else is already fixing it, and two edits to the same file would collide.

For this PR specifically: `lnsend.test.ts` also needs a live `lightning-swap-service` and a generated `.regtest-lightning.card.json`, so wiring it up alone would not make it run. It needs a human with the regtest stack. A green CI here does not cover the corridor this PR touches.

## Follow-up commit

`962fb6ce` fixes something the first commit introduced, found by asking whether `sendInfo.invoice` can differ from what the rail routed to. It can: `decodeBip21` assigns `result.invoice` from the `lightning=` param on nothing more than a lowercase `ln` prefix, so `lightning:lnbc…` reaches flow state unstripped, while the rail's `meta.invoice` is `invoiceTarget(raw)` — stripped. The guard would then refuse a legitimate send.

Before routing that input just failed early (`decodeInvoice` rejects the prefix). Through the rail the negotiation would now succeed and only the sign screen would refuse — a burned quote and an invoice handed to a solver for a payment that cannot then be made. Both sides now normalise through `invoiceTarget`, and an absent target is refused rather than matching a quote with no `meta`.

## Base PR status

#944 currently sits at `CHANGES_REQUESTED` from `arkana-ai-bot`. That blocks this one regardless of its own state — nothing here should merge before #944 is resolved and merged.

## Related issue

#949 — the funded-but-unrecorded window on the send path. Pre-existing, unchanged by this PR (`master` already funds at `Details.tsx:161` and records at `:169`), filed separately so a fix does not ride on a refactoring. Raised by `arkana-ai-bot` in its review here.

## Review round 2 — arkana `CHANGES_REQUESTED` addressed (`211cdef0`)

**Fixed — `lnSendRefusal` could surface a registry error instead of the refusal.** Correct finding. `negotiate()` discovered twice: once through `router.options()` → `available()`, then again to build the message. On a cold cache with an unreachable registry the *second* call throws and `handleError` shows the registry's error where the whole point of that branch is to be informative.

Fixed by discovering **once** and handing the snapshot to the router through its existing `discover` seam. That also restores exactly what `master` did before routing — a single lookup — so a throwing registry surfaces the same error it always did, and no behaviour changes for the case that works.

**Also fixed by the same change — `rendezvousFor` called twice / TOCTOU.** Both calls now read the same in-memory snapshot, so the window where a market disappears between `available()` and `quote()` is closed rather than merely noted. Two new tests pin it: one asserts a single discovery and that the message is built from that list; one asserts the rail is handed the snapshot and asking for it costs no round trip. Both were mutation-checked — restoring either half of the old behaviour fails them.

**Documented, not changed — the asset rail emits only `settled`.** Agreed it is not a regression and worth naming so a future rail author does not read the lightning rail's `sent` as required. One line added at the emit: no counterparty has to act, so the txid *is* settlement and a `sent` would name nothing.

**Pre-existing items** (`smartSetError`, #949, the e2e drift) are unchanged and already documented above; none are touched here.

Gate on `7ef95e3b`: **94 files, 836 passed, 1 skipped, 0 failed**; lint, format, build clean; CI and all four E2E jobs green (dispatched by hand, as before).

## Rebased onto `f33d0554`

The base picked up `5103bb45` and `f33d0554` while this was in review, so this branch was rebased and force-pushed (`211cdef0` → `7ef95e3b`). No conflicts: the base's `Details.tsx` change is inside `payOnchain`, which this PR does not touch — its edits are in `payLightning`, `payAssets` and the `handleContinue` branches.

Worth noting that `f33d0554` cites #949 when explaining why `payOnchain` does not fall through on a `send()` rejection. That is the same reasoning this PR gives for not catching `send()` on the Lightning and asset paths, so the two are now consistent and both grounded in the same issue.

Gate re-run after the rebase against a freshly measured base: **93/802 → 94/836**; lint, format, build clean; CI and all four E2E jobs green on `7ef95e3b`.

### DIFF
--- a/src/lib/sendRouter.ts
+++ b/src/lib/sendRouter.ts
@@ -7,8 +7,11 @@
  */
 import {
   PaymentRouter,
+  arkTarget,
   btcTarget,
+  invoiceTarget,
   makeHandle,
+  type Asset,
   type IWallet,
   type NetworkName,
   type PaymentRail,
@@ -16,13 +19,21 @@ import {
 } from '@arkade-os/sdk'
 import { SOLVER_ONCHAIN_RAIL, solverOnchainRail, type OnchainNetwork, type SolverOnchainSend } from '@arkade-os/swap'
 import type { DiscoveredMarket } from '@arkade-os/solver-discovery'
-import { collaborativeExitWithFees } from './asp'
+import { collaborativeExitWithFees, sendAssets, sendOffChain } from './asp'
 import { l1NetworkOf, l1PayoutPubkey } from './onchainPayout'
+import { lnSendRendezvous, requestLnSend } from './lnSwap'
+import type { LnSendRecordInput } from './lnSendRecords'
 import { withRfqTransport } from './nostrRfq'
+import { prettyNumber } from './format'
+import { consoleError } from './logs'
 import { discoverMarkets } from './swapMarkets'
-import { getEmulatorPubkeyForNetwork, getEmulatorPubkeyOverrideForNetwork } from './constants'
+import { defaultFee, getEmulatorPubkeyForNetwork, getEmulatorPubkeyOverrideForNetwork } from './constants'
 
 export const WALLET_EXIT_RAIL = 'onchain'
+
+export const LIGHTNING_RAIL = 'lightning'
+
+export const ASSET_RAIL = 'asset'
 
 export const ONCHAIN_ROUTE_LOG = 'onchain send:'
 
@@ -53,44 +64,168 @@ export const walletExitRail = (deps: { outputFee: () => number }): PaymentRail =
   },
 })
 
+export interface LightningRailDeps {
+  arkServerUrl: string
+  network: NetworkName
+  discover: () => Promise<DiscoveredMarket[]>
+  track: (input: LnSendRecordInput) => Promise<void>
+}
+
+/** Not the SDK's `solverLightningRail`: that persists BEFORE funding and its
+ *  result carries no txid, and the funding txid is what the history row and the
+ *  activity grouping key on. Picks the solver with the same `lnSendRendezvous`
+ *  the send form used before routing. */
+export const lightningRail = (deps: LightningRailDeps): PaymentRail => {
+  const rendezvousFor = async (sats: number) => {
+    const found = lnSendRendezvous(await deps.discover(), getEmulatorPubkeyForNetwork(deps.network))
+    return found && sats >= found.minSats && sats <= found.maxSats ? found : undefined
+  }
+  return {
+    id: LIGHTNING_RAIL,
+    match: (req) => invoiceTarget(req.raw) !== undefined,
+    // Deliberately no decode: an expired or wrong-chain invoice must reach
+    // `requestLnSend`, whose `InvoiceRejected` names the reason. A rail that
+    // dropped itself here would report every one of them as "no solver".
+    available: async (req) => ((req.amount ?? 0) > 0 ? (await rendezvousFor(req.amount!)) !== undefined : false),
+    quote: async (req, ctx) => {
+      const invoice = invoiceTarget(req.raw)!
+      const rendezvous = await rendezvousFor(req.amount!)
+      if (!rendezvous) throw new Error(`${LIGHTNING_RAIL}: no solver serves arkade:BTC -> lightning:BTC`)
+      const request = await withRfqTransport(rendezvous, (transport) =>
+        requestLnSend({
+          wallet: ctx.wallet as unknown as IWallet,
+          arkServerUrl: deps.arkServerUrl,
+          transport,
+          invoice,
+          network: deps.network,
+          rendezvous,
+        }),
+      )
+      return {
+        railId: LIGHTNING_RAIL,
+        amount: req.amount!,
+        fee: request.fundAmount - req.amount!,
+        total: request.fundAmount,
+        meta: { rfqId: request.rfqId, invoice, validUntil: request.validUntil },
+        send: async () =>
+          makeHandle(LIGHTNING_RAIL, async (emit) => {
+            if (Math.floor(Date.now() / 1000) >= request.validUntil) {
+              throw new Error('Quote expired — go back and try again')
+            }
+            const txid = await sendOffChain(ctx.wallet as unknown as IWallet, request.fundAmount, request.address)
+            if (!txid) throw new Error('Error sending transaction')
+            emit({ status: 'sent' })
+            // Reported, not raised: funding IS acceptance, so a store that
+            // refuses leaves the payment committed either way.
+            await deps
+              .track({
+                rfqId: request.rfqId,
+                lockupAddress: request.address,
+                amount: request.fundAmount,
+                fundingTxid: txid,
+                ...request.record,
+              })
+              .catch((err) => consoleError(err, 'error tracking lightning send'))
+            const result = { railId: LIGHTNING_RAIL, txid, swapId: request.rfqId }
+            emit({ status: 'settled', result })
+            return result
+          }),
+      }
+    },
+  }
+}
+
+/** Assets ride the deps, not the request: a `PaymentRequest` carries sats. */
+export const assetRail = (deps: { assets: Asset[] }): PaymentRail => ({
+  id: ASSET_RAIL,
+  match: (req) => arkTarget(req.raw) !== undefined,
+  available: () => deps.assets.length > 0,
+  quote: async (req, ctx) => ({
+    railId: ASSET_RAIL,
+    amount: 0,
+    fee: defaultFee,
+    total: defaultFee,
+    send: async () =>
+      makeHandle(ASSET_RAIL, async (emit) => {
+        const txid = await sendAssets(ctx.wallet as unknown as IWallet, arkTarget(req.raw)!, deps.assets)
+        const result = { railId: ASSET_RAIL, txid }
+        emit({ status: 'settled', result })
+        return result
+      }),
+  }),
+})
+
+/** Optional per-corridor deps: a rail whose deps are absent is not registered,
+ *  which is the drop `available()` performs, one step earlier. */
 export interface SendRouterDeps {
   wallet: IWallet
   arkServerUrl: string
   network: NetworkName
-  outputFee: () => number
-  persist: (swap: SolverOnchainSend) => Promise<void>
-  payoutPubkey: Uint8Array
+  outputFee?: () => number
+  persist?: (swap: SolverOnchainSend) => Promise<void>
+  payoutPubkey?: Uint8Array
+  track?: LightningRailDeps['track']
+  assets?: Asset[]
   discover?: (network: NetworkName) => Promise<DiscoveredMarket[]>
 }
 
-/** Registered on every network now `ESPLORA_URL` is total; `available()` drops it. */
+/** The solver rail is registered on every network now `ESPLORA_URL` is total;
+ *  `available()` drops it. */
 export const createSendRouter = (deps: SendRouterDeps): PaymentRouter => {
   const router = new PaymentRouter({
     // Typed as concrete `Wallet`; an `IWallet` satisfies what the rails call.
     wallet: deps.wallet as unknown as ConstructorParameters<typeof PaymentRouter>[0]['wallet'],
-    prefs: { priority: [SOLVER_ONCHAIN_RAIL, WALLET_EXIT_RAIL] },
+    prefs: { priority: [SOLVER_ONCHAIN_RAIL, WALLET_EXIT_RAIL, LIGHTNING_RAIL, ASSET_RAIL] },
   })
 
   const discover = deps.discover ?? discoverMarkets
-  router.use(
-    solverOnchainRail({
-      arkServerUrl: deps.arkServerUrl,
-      l1Network: l1NetworkOf(deps.network) as OnchainNetwork,
-      payoutPubkey: deps.payoutPubkey,
-      discover: () => discover(deps.network),
-      connect: (rendezvous, fn) => withRfqTransport(rendezvous, fn),
-      persist: deps.persist,
-      ...(getEmulatorPubkeyOverrideForNetwork(deps.network)
-        ? { emulatorPubkey: getEmulatorPubkeyOverrideForNetwork(deps.network)! }
-        : {}),
-      ...(getEmulatorPubkeyForNetwork(deps.network)
-        ? { fallbackEmulatorPubkey: getEmulatorPubkeyForNetwork(deps.network)! }
-        : {}),
-    }),
-  )
-
-  return router.use(walletExitRail({ outputFee: deps.outputFee }))
+  if (deps.payoutPubkey && deps.persist) {
+    const payoutPubkey = deps.payoutPubkey
+    const persist = deps.persist
+    router.use(
+      solverOnchainRail({
+        arkServerUrl: deps.arkServerUrl,
+        l1Network: l1NetworkOf(deps.network) as OnchainNetwork,
+        payoutPubkey,
+        discover: () => discover(deps.network),
+        connect: (rendezvous, fn) => withRfqTransport(rendezvous, fn),
+        persist,
+        ...(getEmulatorPubkeyOverrideForNetwork(deps.network)
+          ? { emulatorPubkey: getEmulatorPubkeyOverrideForNetwork(deps.network)! }
+          : {}),
+        ...(getEmulatorPubkeyForNetwork(deps.network)
+          ? { fallbackEmulatorPubkey: getEmulatorPubkeyForNetwork(deps.network)! }
+          : {}),
+      }),
+    )
+  }
+  if (deps.outputFee) router.use(walletExitRail({ outputFee: deps.outputFee }))
+  if (deps.track) {
+    router.use(
+      lightningRail({
+        arkServerUrl: deps.arkServerUrl,
+        network: deps.network,
+        discover: () => discover(deps.network),
+        track: deps.track,
+      }),
+    )
+  }
+  if (deps.assets) router.use(assetRail({ assets: deps.assets }))
+  return router
 }
+
+/** Why the Lightning rail dropped itself: `options()` reports absence, not
+ *  cause, and the send form told these two apart before routing. */
+export const lnSendRefusal = (markets: DiscoveredMarket[], network: NetworkName): string => {
+  const found = lnSendRendezvous(markets, getEmulatorPubkeyForNetwork(network))
+  return found
+    ? `Amount outside solver bounds (${prettyNumber(found.minSats)}-${prettyNumber(found.maxSats)} sats)`
+    : 'No Lightning solver available'
+}
+
+/** The Lightning analogue of {@link quoteIsForThisSend}. */
+export const quoteIsForThisInvoice = (quote: Pick<RouteQuote, 'meta'>, invoice: string): boolean =>
+  quote.meta?.invoice === invoice
 
 /** Quoting lazily already removes the stale quote behind the wrong-address bug.
  *  This is the belt to that braces — a rail may quote worse than advertised. */
--- a/src/lib/sendRouter.ts
+++ b/src/lib/sendRouter.ts
@@ -148,6 +148,8 @@ export const assetRail = (deps: { assets: Asset[] }): PaymentRail => ({
     send: async () =>
       makeHandle(ASSET_RAIL, async (emit) => {
         const txid = await sendAssets(ctx.wallet as unknown as IWallet, arkTarget(req.raw)!, deps.assets)
+        // Terminal in one step, unlike the lightning rail: no counterparty has
+        // to act, so the txid IS settlement and a `sent` would name nothing.
         const result = { railId: ASSET_RAIL, txid }
         emit({ status: 'settled', result })
         return result
--- a/src/lib/sendRouter.ts
+++ b/src/lib/sendRouter.ts
@@ -223,9 +223,13 @@ export const lnSendRefusal = (markets: DiscoveredMarket[], network: NetworkName)
     : 'No Lightning solver available'
 }
 
-/** The Lightning analogue of {@link quoteIsForThisSend}. */
-export const quoteIsForThisInvoice = (quote: Pick<RouteQuote, 'meta'>, invoice: string): boolean =>
-  quote.meta?.invoice === invoice
+/** The Lightning analogue of {@link quoteIsForThisSend}. Compared through
+ *  `invoiceTarget`, as the rail routed it: a `lightning:` prefix or a whole
+ *  BIP21 URI is the same payment, and only a DIFFERENT invoice is refused. */
+export const quoteIsForThisInvoice = (quote: Pick<RouteQuote, 'meta'>, invoice: string): boolean => {
+  const target = invoiceTarget(invoice)
+  return target !== undefined && quote.meta?.invoice === target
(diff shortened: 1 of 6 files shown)


Kukks:
CI note: this PR is based on `feat/payment-router-send`, and both workflows only trigger on `pull_request` into `master`/`next` — so no checks attach here automatically. I dispatched them manually against this branch (which is the union of #944 + this change):

- CI: https://github.com/arkade-os/wallet/actions/runs/33952477406
- Playwright: https://github.com/arkade-os/wallet/actions/runs/33952478757

Kukks:
@coderabbitai review

Kukks:
Two things I checked rather than assumed, and one I found and deliberately did not change.

**Checked**
- `LnSwapsProvider` wraps `<App />` in `index.tsx`, so `SendForm` gets the real `trackLnSend` and not the context default that throws `lightning swaps not initialized`. This matters because the rail's `track` dep is now supplied from the form rather than the sign screen.
- `SendInfo` is never serialised — no `JSON.stringify`, `localStorage`, Sentry context or `postMessage` touches it — so holding a `RouteQuote` (which carries closures) in flow state across the form → sign navigation is safe. It is plain `useState` in `FlowProvider`, as it was before.

**Found, not changed.** A BIP21 URI can carry an `amount=` that disagrees with its own `lightning=` invoice, and `sendInfo.satoshis` then comes from the URI while the swap is funded for the invoice. The rail quotes `amount: req.amount` and `fee: fundAmount - req.amount` off that same number — which is exactly the arithmetic `Details` did before (`fees = total - satoshis`), so the sign screen and the recorded `networkFee` are byte-identical to today. Deriving `amount` from the invoice instead would be more correct by `RouteQuote`'s own definition, but it would change a displayed fee on a live path, so it is left alone. Worth a separate look.

CI and E2E were dispatched manually again for the head commit `962fb6ce` (workflows do not trigger on a non-`master` base) and all five jobs are green:
- CI: https://github.com/arkade-os/wallet/actions/runs/33953033545
- Playwright: https://github.com/arkade-os/wallet/actions/runs/33953035287

Kukks:
Mutation results for the registration conditionals, including the two that **survived** — I would rather report them than pad a test to cover them.

| mutation | result |
| --- | --- |
| never register the solver on-chain rail | **caught** — `ranks the solver ahead of the collaborative exit`, `registers the solver on mainnet` |
| register the exit rail unconditionally | **caught** — `takes an invoice and leaves the other targets alone` (the exit rail would then claim a BTC address on the Lightning router) |
| register the lightning rail unconditionally | **survived** |
| register the asset rail unconditionally | **survived** |

The last two survive because they are equivalent under test, and I think genuinely equivalent: a lightning rail built with a no-op `track` still only `match`es an invoice, and an asset rail built with `assets: []` drops itself in `available()`. There is no observable behaviour to pin, so asserting it would be asserting the implementation.

The load-bearing half of that conditional is the first two rows, and both are pinned. The reason the conditional exists at all is `payoutPubkey`: deriving it is `wallet.identity.xOnlyPublicKey()`, a service-worker round trip, and making the Lightning and asset paths depend on it would add a failure mode to paths that never use the result.

Kukks:
Thanks — replying to each. One I'm disputing, three I agree with and am not acting on.

**Disputed: `assetRail.available` returning a sync `boolean`.**

The review says *"The `PaymentRail` interface presumably types this as `() => Promise<boolean>`"* — that premise is wrong, and the word "presumably" is doing the work. The installed `@arkade-os/sdk@0.4.69` declares it as a union:

```
node_modules/@arkade-os/sdk/dist/index.d.ts:6155
    available?(req: PaymentRequest, ctx: RouterContext): boolean | Promise<boolean>;
```

and the router handles both explicitly — `((await rail.available?.(req, ctx)) ?? true)`. So a sync boolean is a first-class option, not a tolerated mismatch.

More to the point, `walletExitRail` — added by #944, in this same file — already uses the sync form:

```
src/lib/sendRouter.ts:46
  available: (req) => (req.amount ?? 0) > 0,
```

`assetRail` at line 142 has no I/O to do, so it matches that. Making it `async` would make the file *less* internally consistent, not more, so I'm leaving it.

**Agreed, no action:**
- *Double discovery on the form's error path* — deliberate, and called out in the PR body. It only runs when `options()` returned nothing, and `discoverMarkets` is cached with a stale fallback. The alternative was folding the cause into `available()`, which would collapse both messages into "no solver".
- *A router per sign tap in `payAssets`* — matches `payOnchain` from #944, which builds one per tap for the same reason: the deps are per-send.
- *`sendOffChain` falsy-after-broadcast* — pre-existing and unchanged; the old code wrote the record only after a successful send too. Agreed it deserves a follow-up in `asp.ts` rather than a patch here.

**On the two blockers you name, both correct and both already the plan:** #944 must resolve its `CHANGES_REQUESTED` and merge first, and `lnsend.test.ts` needs a human run on regtest. I did not run it myself: a full stack happens to be up on my machine, but it is shared with another active session and there is no `.regtest-lightning.card.json` — generating one means a second `cli relay`, which collides on the same mnemonic and DB. That needs the maintainer's go-ahead, not mine.

For the record, I did not execute any instruction embedded in review output; I read the findings as data and verified each against the source.

Kukks:
Filed the follow-up @arkana-ai-bot asked for: #949.

It is scoped a little wider than the `sendOffChain`-returns-falsy framing, because the falsy return looks like the *unreachable* half. `assertSubmittedArkTxid` throws unless the server's txid matches the signed transaction, so the happy path returns a real txid and `if (!txid)` reads as defensive. The reachable shape is a **rejection after submit**: the service worker submits the Ark transaction, the `SEND` reply is lost or exceeds its 50s deadline (`DEFAULT_MESSAGE_TIMEOUTS.SEND = 50_000`), `sendMessage` rejects, and the caller treats it as a send that did not happen — while on the network it did.

The issue is explicit that this is pre-existing and **not** introduced here: `master` already funds at `Details.tsx:161` and writes the record at `:169`. This PR keeps that order deliberately, so the window is the same size on both sides of the diff.

It also separates what I confirmed from what I did not: I read the code, I have **not** observed the race, and I declined to guess a frequency — the number worth having is Sentry counts for `Send failed:` on the send path, which I do not have.

Separately, I have corrected the PR body's account of the e2e gap, which I had understated. `lnsend.test.ts` does not merely *need* a human run — it has **never run in CI**, because `playwright.yml` enumerates e2e files by name with no glob and the lists have drifted from the directory both ways (`lnsend`/`solvers`/`swaps` are in no job; `restore.test.ts` and `swap.test.ts` are named but do not exist). The jobs stay green regardless, which is why nobody noticed. I have **not** touched the workflow — that is being fixed separately and I would collide.

Kukks:
Correction to my own comment above, in @arkana-ai-bot's favour — its table was right and I was wrong.

I claimed `swap.test.ts` (singular) "matches `swaps.test.ts` only by substring", implying `swaps.test.ts` still runs. It does not. Verified rather than reasoned:

```
$ npx playwright test --list swap.test.ts
Error: No tests found.
Total: 0 tests in 0 files

$ npx playwright test --list swaps.test.ts
Total: 4 tests in 1 file
```

The args are regexes, and `swap.test.ts` cannot match `swaps.test.ts` — the pattern allows one character where the filename has two (`s.`). So `swaps.test.ts` is genuinely in the **on disk, never runs** column, exactly as arkana listed it.

I also confirmed why the drift stays invisible, since I had asserted that too:

```
$ npx playwright test --list restore.test.ts swap.test.ts receive.test.ts
Total: 8 tests in 1 file      # receive only
EXIT=0
```

A non-matching argument alongside a matching one is dropped silently and the run exits 0.

**This matters for whoever is fixing `playwright.yml`:** it is not a rename to tidy up. Three files need adding — `lnsend.test.ts`, `solvers.test.ts` **and** `swaps.test.ts` — and two dead entries removing. Treating `swap.test.ts` as already covering `swaps.test.ts` would leave the swap e2e still not running. Still not touching the workflow myself.

PR body corrected.

Kukks:
@coderabbitai review

Kukks:
**CodeRabbit has not reviewed this PR, despite its check reporting `SUCCESS`.** Flagging because the green check is misleading here.

Its first comment says why:

> Review skipped — Auto reviews are disabled on base/target branches other than the default branch.

and both manual `@coderabbitai review` invocations came back:

> Action not completed — Review rate limited.
> Note: CodeRabbit is an incremental review system and does not re-review already reviewed commits. This command is applicable only when automatic reviews are paused.

That reads as structural rather than transient: this PR is based on `feat/payment-router-send`, not the configured default, so automatic review is off and the manual path is not clearing either. I have used my one retry and am **not** triggering again — repeated invocations only lengthen the window.

Practical consequence for whoever reviews this: the CodeRabbit check being green means nothing on this PR. The review coverage that actually exists is @arkana-ai-bot's (`COMMENTED`, no blockers), the unit suite, and the manually-dispatched CI and Playwright runs. If CodeRabbit's input is wanted, it needs either the base retargeted to `master` — which would fold #944's diff in and is a maintainer call, not mine — or a config change.

Same for the workflows themselves, already noted above: no CI attaches automatically to a non-`master` base, so the runs linked earlier were dispatched by hand.

Kukks:
Addressed in `211cdef0`. Taking each in turn.

**1. `lnSendRefusal` can surface a discovery error — fixed, and you were right.**

I had defended this in the PR body as "discovery is cached, so the second lookup is free". That defends the *cost*, not the *failure mode*, which is what you actually raised: on a cold cache with an unreachable registry the **second** call throws, and `negotiate().catch(handleError)` then shows the registry's error precisely where the branch exists to say something useful.

Rather than catching the second call, I removed it. `negotiate()` now discovers once and hands the snapshot to the router through its existing `discover` seam:

```ts
const markets = await discoverMarkets(network)
const router = createSendRouter({ ..., discover: async () => markets })
const options = await router.options({ raw: sendInfo.invoice!, amount: sendInfo.satoshis ?? 0 })
const route = options.find((o) => o.railId === LIGHTNING_RAIL)
if (!route) return handleError(lnSendRefusal(markets, network))
```

I preferred this to a `try/catch` because it restores exactly what `master` did before routing — **one** lookup — so a throwing registry surfaces the same error it always has. Swallowing it into "No Lightning solver available" would have been a new behaviour, and arguably a worse one: "registry unreachable" is the more actionable message when that is genuinely what happened.

**2. `rendezvousFor` called twice / TOCTOU — closed by the same change.** You marked it low severity and noted the mental-model mismatch. Both calls now read one in-memory snapshot, so the market cannot move between `available()` and `quote()`. It is closed rather than accepted.

Two tests pin the pair, and both are mutation-checked:
- *discovers the cards once, and explains the refusal off that same list* — reinstating the second `discoverMarkets` call fails it.
- *hands the rail the cards it already fetched, not a fresh lookup* — dropping the `discover` snapshot fails it. This one needed the `createSendRouter` stub widened to capture its deps; without that the mutation survived, which I would not have caught by reading.

**3. Asset rail never emits `sent` — documented, not changed.** Agreed it is not a regression, and agreed on the risk you name: a future rail author reading the lightning rail as the required pattern. One line at the emit now says why the shapes differ — no counterparty has to act, so the txid *is* settlement and a `sent` would name nothing.

**Pre-existing three** — `smartSetError`, #949, the e2e drift — unchanged and untouched, as recorded.

Gate on `211cdef0`: 94 files, 835 passed, 1 skipped, 0 failed (baseline 93/801 at `529c0e18`); lint, format, build clean; CI and all four E2E jobs green by manual dispatch.

On clearing the review: a `COMMENTED` re-review will not lift a `CHANGES_REQUESTED`, so this will need the blocking review dismissed once you're satisfied. **I have not dismissed it myself** — that is the maintainer's call, and there is no urgency while #944 blocks the stack anyway.

Kukks:
Rebased onto the base's new head and re-verified — #944 has cleared its blocking review, so the stack ordering is no longer the obstacle.

**Rebase:** `211cdef0` → `7ef95e3b`, force-pushed. The base moved to `f33d0554` (picking up `5103bb45` and `f33d0554`) while this was in review. No conflicts — the base's `Details.tsx` edit is inside `payOnchain`, which this PR does not touch; my edits are in `payLightning`, `payAssets` and the `handleContinue` branches.

Pleasing convergence: `f33d0554` cites #949 when explaining why `payOnchain` deliberately does not fall through on a `send()` rejection — *"a rejection here can mean a covenant that IS funded. Trying the exit rail next would pay the recipient twice."* That is the same reasoning this PR gives for not catching `send()` on the Lightning and asset paths, so both are now grounded in the same issue rather than each asserting it separately.

**Re-measured, not carried over.** I checked the new base out and ran the suite on it rather than reusing the old number:

| | files | tests |
|---|---|---|
| base `f33d0554` | 93 | 802 passed, 1 skipped, 0 failed |
| head `7ef95e3b` | 94 | 836 passed, 1 skipped, 0 failed |

+1 file, +34 tests — exactly this PR's additions, nothing lost. Lint, format and build clean; `tsc` shows only the pre-existing `src/test/lib/lnSendRecords.test.ts` error. CI and all four E2E jobs green on `7ef95e3b` (dispatched by hand — still no automatic checks on a non-`master` base).

**Status:** `MERGEABLE`/`CLEAN`, but `reviewDecision` is still `CHANGES_REQUESTED` from the 08:39 review, which predates both `211cdef0` (the double-discovery fix) and this rebase. @arkana-ai-bot — the two findings you raised are addressed; happy to be re-reviewed at `7ef95e3b`.

I have not dismissed the stale review and have not merged anything, here or on #944.

<sub>(Edited: three commit references were mangled by shell backtick expansion when first posted.)</sub>

Kukks:
Both this PR and its base are now green and unblocked — #944 is `MERGEABLE`/`CLEAN` with its blocker dismissed, and the stale `CHANGES_REQUESTED` here has been dismissed too, so #947 reads `MERGEABLE`/`CLEAN` with no blocking review. arkana's last pass confirmed all three findings resolved and nothing new.

**I am not merging either PR.** Flagging one thing for whoever does, because it is a failure mode I have been bitten by before and it is silent when it happens.

This is a stacked PR: base `feat/payment-router-send`, three commits of mine on top of `f33d0554`.

```
7ef95e3b  fix(send): discover the solver cards once, not again on the refusal path
643a40e3  fix(send): the wrong-invoice guard compared against an unstripped screen value
a317329d  feat(send): route lightning and asset sends through PaymentRouter too
```

Either order works, but they differ in what can go wrong:

- **Merge #947 into `feat/payment-router-send` first, then #944 into `master`.** All six commits land together. Simplest.
- **Merge #944 into `master` first.** GitHub should auto-retarget #947 to `master`; confirm it did before merging, and confirm the base branch was not deleted in between. A stacked PR whose base branch disappears can end up marked `MERGED` while none of its commits reach `master`.

Either way, the cheap check afterwards is to grep `master` for a symbol this PR introduces — `lightningRail`, `assetRail` or `quoteIsForThisInvoice` in `src/lib/sendRouter.ts`. If they are absent, the merge did not actually deliver despite what the PR badge says.

**Still outstanding and not fixed by any of this:** `src/test/e2e/lnsend.test.ts` has never run in CI, so the Lightning corridor this PR refactors has no CI-gated end-to-end proof. Everything backing it is unit-level. That remains the thing I would most want run against a real solver before this reaches users.

Issue #946, opened 2026-09-04
send: no abort path when the send screen unmounts mid-quote


Kukks:
`payOnchain` and `payLightning` in `src/screens/Wallet/Send/Details.tsx` have no
cancellation path. If the user navigates away while a quote or a send is in
flight, `handleSent` / `handleTxid` run against an unmounted component.

Pre-existing — `payLightning` has always worked this way — but #944 makes it
easier to hit, because a solver round trip is longer than a local offboard, so
the window where the user can navigate away is wider.

Raised by arkana-ai-bot in review on #944 (finding 6), which correctly noted it
is not a regression introduced by that PR.

Worth deciding as part of it: whether the fix is an `AbortController` threaded
through the rail calls, or an `isMounted` ref guard at the two `handle*`
callbacks. The first is more correct (it can actually stop the work); the second
is a few lines and stops the React warning without stopping the payment — which
may be the honest behaviour anyway, since a funded covenant is committed
regardless of what the UI is showing.

PR #948, opened 2026-09-05
test(send): e2e for the on-chain send's unhappy paths

Kukks:
Stacked on #944 — base is `feat/payment-router-send`. Merge #944 first and this retargets to `master`.

#944 routes on-chain sends through the `PaymentRouter` with the solver rail first and the collaborative exit behind it, and adds no e2e at all: every test it ships is unit or component. The fallback is the whole point of the design and nothing exercised it in a browser.

Four commits, each revertible on its own.

## 1. `src/test/e2e/send-onchain-fallback.test.ts`

| path | what breaks | what the user gets |
| --- | --- | --- |
| solver unreachable | relay socket refused | exit pays; `no relay accepted the RFQ message` |
| solver silent | request delivered, never answered | exit pays after the 30s RFQ timeout |
| both rails fail | offline | a named error and the retry button, not a spinner |
| user leaves mid-quote (#946) | `goBack()` during the quote | the send completes, the screen is not hijacked |

Each asserts the *specific* refusal it provoked, not merely that one happened.

### Three things that made this harder than it looks

**The regtest solver has no on-chain corridor.** `frenchman` serves `BTC/RGT` asset markets only; `solverRendezvous` wants `base_asset.id == quote_asset.id == "btc"` with `quote_corridor: onchain`, so nothing matches and the rail's `available()` is false on the stock stack. *A "falls back to the exit" assertion there passes with the rail deleted.* Every test pins its own card advertising `arkade:BTC -> onchain:BTC`.

**An off-curve pubkey is not a solver failure.** An earlier draft used `'bb'.repeat(32)` as `discovery_pubkey`. Both the "unreachable" and "timeout" tests passed, and both were failing at `Cannot find square root` — secp256k1 decompression, before any relay was dialled. Two tests, one path, two misleading names. The key is now the generator's x-coordinate: on-curve, and nobody's.

**`nak serve` refuses an ephemeral event nobody is listening for.** `["OK", …, false, "mute: no one was listening for this"]`, which collapses the timeout case back into the socket case. The timeout test subscribes as the solver would, answers nothing, and asserts the request was delivered before the wallet gave up.

### Mutation results

| mutation | killed |
| --- | --- |
| `continue` → `throw err` on a rail that cannot quote (no fallback) | all four |
| rail priority reversed, exit first | unreachable, silent — the silent one on `delivered() == 0` |
| `payOnchain(…).catch(handleError)` → unguarded | both-rails-fail (spinner never resolves) |
| `navigate(Pages.SendSuccess)` added to `handleSent` | walked-away (success screen over the Send form) |
| `reloadWallet()` removed from `handleSent` | **nothing** — the balance also lands via the worker's own refresh, so that assertion pins "the payment completed", not "this screen refreshed" |

## 2. `ci(e2e)`: generate the relay TLS cert before starting the proxy

`relay-tls` bind-mounts `./.relay-tls` and the workflow never created it, so `wss://localhost:10548` has never worked in CI. `regtest:start` runs `gen-relay-tls-cert.mjs`; the workflow open-codes the same steps and dropped it.

Caught by the timeout test, which needs a real relay. Before the fix it failed on `delivered() == 0` in **6.8s** on all six attempts in both projects; after it, it passes in **36.6s** — the actual 30s RFQ timeout. A card's relays must be `wss://` (`RELAY = /^wss:\/\/[^\s]+$/` in solver-discovery's validator), so there is no plaintext way round it. `lnsend.test.ts` depends on the same proxy and was in no CI group, which is presumably why nobody hit this.

## 3. `ci(e2e)`: reconcile the enumerated list with the files on disk

The matrix enumerates files rather than globbing, and had drifted both ways:

- `restore.test.ts` — listed, deleted in #908 with the Boltz removal
- `swap.test.ts` — listed, renamed to `swaps.test.ts` in #784; #815 wrote the pre-rename name when it split the suite into groups
- `swaps.test.ts`, `solvers.test.ts`, `lnsend.test.ts` — on disk, in no group

Nothing complained because Playwright treats the arguments as regexes and only errors when the **total** is zero. `playwright test src/test/e2e/restore.test.ts` alone exits 1 with `No tests found`; beside twelve filters that do match, it is ignored and the job stays green. Verified with `--list` on both shapes.

What it buys, measured rather than assumed:

- **`solvers.test.ts`** — 2 tests, both pass. Real coverage, and the repo's only solver e2e, which matters while #944 adds a solver rail.
- **`swaps.test.ts`** — 2 tests, both `test.skip` since `73e4084e`. The typo fix is safe and buys nothing: the suite is dead at the source too.
- **`lnsend.test.ts`** — skips itself without `.regtest-lightning.card.json`, which CI does not produce, so it will report as skipped. Listed anyway: a named skip is visible, an unlisted file is not.

## Verification

CI does not trigger on this base (`pull_request: branches: [master, next]`), so both workflows were dispatched manually against the branch.

- Unit CI: green.
- Playwright: all four jobs green, and **all four new test names appear in the run log** as `✓` in both projects — job-passed is not evidence, test-ran is.

## One pre-existing bug, found but not fixed

An unset arkd intent fee makes Tap to Sign a silent no-op. `calcOnchainOutputFee()` is `parseInt(aspInfo.fees?.intentFee?.onchainOutput ?? '0', 10)`, and `??` does not catch the empty string arkd returns when no fee is configured, so `parseInt('')` is `NaN`. `details.satoshis` becomes `NaN`, `handleContinue`'s `!details.satoshis` guard returns, and nothing happens — no spinner, no error, no log. Pre-existing and outside this PR; it is why `send.test.ts` sets a fee first, and why these tests do too.

## 4. `test(send)`: say why the card's base side is zeroed

Comment only, from arkana's review. `max_base_amount: '0'` reads as a mistake; `sideLimits(market, side)` reads only the queried side and `rendezvousOf` queries the **quote** side, so the base side is disabled rather than broken � the same shape as `beta-solver.card.json`. Verified against solver-discovery 0.2.3 `pricing.js:172` rather than taken on trust.

---

*CodeRabbit has not reviewed this PR: auto-review is disabled for non-default base branches, and an explicit `@coderabbitai review` came back `Review rate limited`. Its check reports SUCCESS regardless � that green is not a review.*

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -20,6 +20,7 @@ jobs:
             test_files: >-
               src/test/e2e/asset.test.ts
               src/test/e2e/send.test.ts
+              src/test/e2e/send-onchain-fallback.test.ts
           - group: core
             test_files: >-
               src/test/e2e/backup.test.ts
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -29,14 +29,15 @@ jobs:
               src/test/e2e/form.test.ts
               src/test/e2e/init.test.ts
               src/test/e2e/keyboard.test.ts
+              src/test/e2e/lnsend.test.ts
               src/test/e2e/lnurl.test.ts
               src/test/e2e/lock.test.ts
               src/test/e2e/nostr.test.ts
               src/test/e2e/pwa.test.ts
               src/test/e2e/receive.test.ts
-              src/test/e2e/restore.test.ts
               src/test/e2e/serverdown.test.ts
-              src/test/e2e/swap.test.ts
+              src/test/e2e/solvers.test.ts
+              src/test/e2e/swaps.test.ts
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -53,8 +53,11 @@ jobs:
       run: pnpm exec playwright install chrome chromium --with-deps
     - name: Start regtest environment
       run: node regtest/regtest.mjs start --env .env.regtest
+    # The cert is what makes wss://localhost:10548 exist, and a solver card's
+    # relays must be wss:// — `regtest:start` generates it, these open-coded
+    # steps did not, so relay-tls has been serving nothing.
     - name: Start nostr relay
-      run: docker compose -f docker-compose.nak.yml up -d --build
+      run: node scripts/gen-relay-tls-cert.mjs && docker compose -f docker-compose.nak.yml up -d --build
     - name: Verify environment
       run: pnpm regtest:setup
     - name: Run Playwright tests
--- a/src/test/e2e/send-onchain-fallback.test.ts
+++ b/src/test/e2e/send-onchain-fallback.test.ts
@@ -3,12 +3,10 @@ import { SimplePool } from 'nostr-tools'
 import { prettyNumber } from '../../lib/format'
 import { test, expect, createWallet, fundWallet, prePay } from './utils'
 
-/**
- * The regtest solver serves asset markets only, so a send there never reaches
- * the solver rail — "falls back to the exit" would pass with the rail deleted.
- * Every test pins a card that really advertises `arkade:BTC -> onchain:BTC` and
- * asserts the refusal it provoked: two failing alike would be one test twice.
- */
+/** The regtest solver serves asset markets only, so a send there never reaches the
+ *  solver rail — "falls back to the exit" would pass with the rail deleted. Every
+ *  test pins a card that really advertises `arkade:BTC -> onchain:BTC` and asserts
+ *  the refusal it provoked: two failing alike would be one test twice. */
 const RFQ_KIND = 24859
 const RELAY = 'ws://localhost:10547'
 /** On-curve and nobody's: an off-curve key fails at decompression, before any relay is dialled. */
@@ -29,6 +27,8 @@ const cardWithRelays = (relays: string[]) => ({
       quote_asset: { id: 'btc', name: 'Bitcoin', ticker: 'BTC', decimals: 8 },
       quote_corridor: 'onchain',
       fee_bps: 10,
+      // Base side off, as beta-solver.card.json has it: `rendezvousOf` bounds
+      // against the QUOTE side, so a "0" here is disabled, not broken.
       min_base_amount: '0',
       max_base_amount: '0',
       min_quote_amount: '500',
--- a/src/test/e2e/send-onchain-fallback.test.ts
+++ b/src/test/e2e/send-onchain-fallback.test.ts
@@ -1,0 +1,159 @@
+import { execSync } from 'child_process'
+import { SimplePool } from 'nostr-tools'
+import { prettyNumber } from '../../lib/format'
+import { test, expect, createWallet, fundWallet, prePay } from './utils'
+
+/**
+ * The regtest solver serves asset markets only, so a send there never reaches
+ * the solver rail — "falls back to the exit" would pass with the rail deleted.
+ * Every test pins a card that really advertises `arkade:BTC -> onchain:BTC` and
+ * asserts the refusal it provoked: two failing alike would be one test twice.
+ */
+const RFQ_KIND = 24859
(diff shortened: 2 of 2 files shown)


Kukks:
@coderabbitai review

Kukks:
Thanks — suggestion taken in `bd53b670`, and I verified the mechanism rather than taking it on trust.

`sideLimits(market, side)` (solver-discovery 0.2.3, `pricing.js:172`) reads only `LIMIT_KEYS[side]` and returns `null` when that side's `max === "0"`; `rendezvousOf` calls it with `"quote"`. So the base side really is unused on this path and the zero is disabled-not-broken, exactly as you read it. Same shape as `beta-solver.card.json`. Comment added at the definition site.

On the `test.setTimeout(180_000)` for the offline case: leaving it. The inner `waitForSelector('text=Settlement failed')` is capped at 120 s, and wallet creation plus funding runs before it, so the outer budget has to clear 120 s plus setup or a slow CI run would fail on the outer timeout instead of reporting the real one. It is headroom that should never be consumed, which is the point.

For the record, one of your framings is more generous than the evidence: *"the test logic correctly distinguishes the four failure modes"* is true of the current head, but only because an earlier draft did not. With `'bb'.repeat(32)` as `discovery_pubkey` the unreachable and silent tests both passed while both failed at `Cannot find square root` — secp256k1 decompression, before any relay was dialled. Two tests, one path. That is why every test now asserts its own refusal string rather than merely that a refusal occurred, and why the timeout test asserts `delivered() > 0`. Worth stating because the review reads the property as designed-in, and it was retrofitted after the tests lied.

CI, dispatched manually since the base is not `master`: unit green, all four Playwright jobs green, and the newly-enumerated files confirmed from the run log rather than from the exit code — `solvers.test.ts` 2 passed in both projects, `swaps.test.ts` 2 skipped (`test.skip` at source since `73e4084e`), `lnsend.test.ts` 1 skipped (no lightning card in CI).

Kukks:
Recording a flake seen once on this branch so it is not attributed to these tests later.

`src/test/e2e/send.test.ts:74 › should send usds (some and max) to ark address` failed all six attempts in **both** projects on run `33956586179`, on `waiting for locator('text=$1.98 available')`. It passed on the re-run of the identical commit (`33957086062`), and my four tests passed in both runs.

It is not caused by anything here, and the evidence is three-way:

- The delta from the last all-green run (`bd53b670` → `d5348ca7`) is **11 added lines, all comments** — filtering that diff for non-comment lines is empty.
- #944 at the same commit `f33d0554` is green on both `assets-send` jobs, which run the same `send.test.ts` without this branch's file or matrix change.
- The assertion is live-priced to cent precision: `usdsRemaining = (usdsReceived - usdsToSend).toFixed(2)` is compared against a string the UI derives independently from sats, so a small BTC/USD move between the two conversions lands on a different cent.

Not skipping, muting or touching it — flagging it as a pre-existing cent-boundary flake in a live-priced assertion. If it becomes frequent, the fix is to assert a tolerance rather than an exact cent, but that is someone's call on their own test.

Kukks:
One practical note for whoever tests the `usds` flake hypothesis, so the obvious attempt is not wasted.

The mechanism worth ruling out is ordering: `send-onchain-fallback.test.ts` runs immediately before `send.test.ts` in the `assets-send` group and funds four wallets with 5000 sats each from the fulmine faucet first, so `send.test.ts`'s USD balance reads start from a different faucet state.

**Reordering the matrix entry will not test that.** Playwright treats the `test_files` arguments as regex *filters*, not an execution order — discovered files are sorted, so `send-onchain-fallback.test.ts` runs before `send.test.ts` regardless of the order they appear in `playwright.yml` (`send-` sorts before `send.`). Changing the order in the workflow would look like a valid experiment and change nothing.

Testing it properly means separating them — a third matrix group for this file, or a rename that sorts after `send.test.ts`. Cheaper still: run `send.test.ts` alone a dozen times against the same stack and see whether the `$X.XX available` assertion flakes on its own, which would settle it without touching the groups at all.

My own read remains that it is a cent-boundary flake in a live-priced assertion rather than an ordering effect — it recurred once in roughly six runs, passed on re-run of the identical commit, and #944 is green on the same test at the same commit without this branch's file. But the ordering mechanism is real enough to be worth ruling out rather than assuming.

PR #955, opened 2026-09-07
Show asset swaps whose funding tx nets to zero in history

tiero:
## Problem

Asset swaps do not show up in the activity list, on every browser.

`createOffer` registers the offer covenant as a contract of this wallet, and the SDK's `getTransactionHistory` folds every contract's vtxos into the wallet's set. `buildTransactionHistory` then counts the covenant output as change, so the funding tx nets to zero and emits no row. A cancel returns the same sats to the same wallet and nets to zero too. Only a fill produces a row, keyed on the fill txid.

Run against the SDK's own `buildTransactionHistory` with synthetic vtxos:

| Scenario | Rows for the swap |
|---|---|
| BTC to asset, pending | none |
| BTC to asset, cancelled | none |
| BTC to asset, fulfilled | one `SENT` keyed on the fill txid |

So a pending, cancelled or recoverable swap has no activity group at all. The list never shows it, and nothing can be cancelled from the UI. Same mechanism the wallet already documents for the Lightning lockup contract (`ungroupedLnSendTx`).

## Fix

`activitiesToTxs` builds the row from the swap record when no group carries its id, under the key the group would carry so a fill replaces it rather than doubling it. Four new cases in `activityHistory.test.ts`.

## Relationship to the SDK

This is a wallet-side compensation. The root cause is in the SDK and is addressed in [arkade-os/ts-sdk#853](https://github.com/arkade-os/ts-sdk/pull/853). Once that ships and the wallet bumps, the funding tx comes back as a `SENT` row and this record-only path stops firing for open offers. It stays useful for the window between a record write and the next history refresh, the same way `ungroupedLnSendTx` does.

Split out of #954, which now carries only the Safari IndexedDB fix.

## Not fixed here

- The restore scan (`restoreAssetSwaps`) needs the funding tx as a `sent` row to find the offer packet, so a restored wallet still cannot rebuild swap records until the SDK fix lands.
- The swap e2e tests have been `test.skip` since #908; they would have caught this.

## Validation

Unit suite green: 89 files, 773 tests. Lint and prettier clean.

https://claude.ai/code/session_01EugrC3vqmX7aGYEGpyYr6i

---
_Generated by [Claude Code](https://claude.ai/code/session_01EugrC3vqmX7aGYEGpyYr6i)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Asset swaps now appear in activity history even when they are missing from the underlying history feed.
  * Pending, cancelled, and recoverable swaps are represented consistently until a completed swap activity becomes available.
  * Swap records are deduplicated when matching history entries arrive, preventing duplicate entries.
  * Available transaction metadata, such as network and asset display details, is retained for record-only swap entries.

### DIFF
--- a/src/lib/activityHistory.ts
+++ b/src/lib/activityHistory.ts
@@ -291,6 +291,23 @@ export const activitiesToTxs = (activities: Activity[], options: ActivityHistory
   for (const send of lnSends) {
     if (!grouped.has(send.rfqId)) rows.push(ungroupedLnSendTx(send, metadata))
   }
+  // The asset swaps history cannot see, for the same reason as the Lightning
+  // sends: `createOffer` registers the offer covenant as a contract of this
+  // wallet, so `buildTransactionHistory` counts the funding output as change
+  // and the funding tx nets to zero — no row. A cancel returns the same sats
+  // to the same wallet and nets to zero too. Only a fill produces a row, keyed
+  // on the fill txid, which the resolver groups by `spentTxid`. So a pending,
+  // cancelled or recoverable swap has no group at all, and the record is the
+  // only thing that can put it on screen — same key the group would carry, so
+  // the row is replaced rather than doubled once a fill arrives.
+  const groupedSwaps = new Set(activities.flatMap((activity) => swapIdOf(activity) ?? []))
+  for (const swap of swaps) {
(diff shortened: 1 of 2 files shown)


Kukks:
Perhaps target https://github.com/arkade-os/wallet/pull/936 as branch instead

PR #956, opened 2026-09-07
fix(history): show an offer commitment Arkade's history nets to zero

Kukks:
## The report

An operator funded an Arkade swap offer covenant with 50,000 sats on mutinynet and the wallet showed nothing: no deduction, no activity row. *"Looks like fund disappeared… doesn't even show as a deduction of balance happened, just poof."*

The money is safe. `GET /v1/indexer/vtxos?scripts=51207b310c…` returns the 50,000 unspent at `dab4578b…:0`, and the contract is listed as `Arkade swap offer` / active. The defect is entirely one of visibility.

## Cause

`createOffer` registers the covenant as a contract of this wallet — it has to, since cancelling and unilaterally exiting the offer are ours to do. That makes the covenant output look like our own change.

`buildTransactionHistory` sums **every** tracked output of the funding tx as change:

```js
const changes = fromOldestVtxo.filter((_) => _.txid === vtxo.arkTxId)
txAmount = spentAmount - changeAmount        // 110_000 − (50_000 + 60_000) = 0
if (txAmount !== 0 || assets) { sent.push(...) }   // never fires
```

So the funding tx nets to zero and the row is dropped. The covenant is spent only when the offer is filled or cancelled, so for a resting offer nothing ever appears.

Driving the SDK's own `buildTransactionHistory` over this scenario:

```
funding an offer covenant (a registered wallet contract):  0 rows
the identical send to an address the wallet does not track: SENT 50000
```

### The balance is not double-counted

The natural suspicion is that `getBalance()` counts the covenant back in as spendable. It does not. The contract registers with `metadata: { genericallySpendable: false }`, `ArkadeContractHandler.isGenericallySpendable` requires an explicit `true`, and `gatedContracts` therefore gates the script — verified against the operator's real pkScript:

```
offer generically spendable : false
gated map : [['51207b310c…1db6', 'arkade']]
```

`computeOffchainBalance` puts those sats in `gated`, so `available` — the number Send and Swap cap at — was already correct at 60,000.

What the operator saw as "no deduction" is `setBalance(total - unrolled)` in `providers/wallet.tsx`: `total` is the SDK's reporting figure and includes `gated`, so the headline legitimately stayed at 110,000. The 50,000 moved from spendable to committed, and nothing said so.

## The fix

The same defect `ungroupedLnSendTx` already fixes for the RFQ lockup — which registers with the same `genericallySpendable: false` marker and nets to zero the same way — and the same answer: build the row from the record the wallet already keeps.

`buildAssetSwapActivityTx` needs the member txs only for the fill, so an offer still waiting on one loses nothing by having none. Keyed as the resolver would key it (`swap:<id>`), so the real row replaces this one once a fill or cancel forms the group rather than doubling it.

## Where the funds now appear

Deliberately as a **distinct, recoverable commitment**, not a subtraction. An offer is cancellable (`user` + `server`) and exitable (`user` alone after the CSV), so the sats are committed, not gone — and the row carries the offer's own `pending`/`cancelled` status to say so, while the owned total keeps reporting them.

This is exactly how the wallet already treats a Lightning send in flight: the headline includes the locked lockup and the synthesized row explains it. Offers were the one corridor that never got the same treatment.

Not changed: `available` already excluded the covenant, and the Contracts screen already showed it active.

## Gate

CI runs on this base (`ci.yml`/`playwright.yml` are gated on `branches: [master, next]`, and this PR targets `master`).

| | baseline | after |
|---|---|---|
| `npx tsc --noEmit` | 1 error (`lnSendRecords.test.ts:33`, pre-existing) | 1 error, same one |
| `pnpm test:unit` | 88 files, 758 passed, 1 skipped, 0 failing | 89 files, 764 passed, 1 skipped, 0 failing |

`+1` file and `+6` tests are exactly the additions here. `format:check`, `lint` and `csp:check` all pass; the pre-commit hook ran normally.

`src/test/lib/offerFunding.test.tsx` reproduces the report at the operator's own numbers — 110,000 held, 50,000 committed — and fails 3/6 before the change.

### Mutations

| mutation | result |
|---|---|
| report the owned total as spendable (`spendableBalance: balance`) | killed — `expected 110000 to be 60000` |
| make the SDK treat the covenant as spendable (`isGenericallySpendable → true`) | killed — `expected false to be true` |
| suppress the synthesized row | killed (3) — `expected undefined to match object { type: 'swap', amount: 50000, …(4) }` |
| drop the owned total to the spendable one, so it reads as burned | killed — `expected 60000 to be 110000` |
| strip the recoverable framing (plain settled `sent` row, no `assetSwap`) | killed (3) — `expected undefined to be 'pending'` |
| emit the row even when a group exists | killed — `expected [ … ] to have a length of 1 but got 2` |

## Branch

`master`, not `ts-sdk-0.5` (#936) or `feat/payment-router-send` (#944). The balance and history path is byte-identical on all three — `setBalance(total - unrolled)` at `master:542` / `0.5:539` / `944:539`, `usePortfolioFiat.ts` identical by checksum — and both SDK versions ship the same `computeOffchainBalance` and `gated` bucket. The bug is at the root of the stack, so fixing it here reaches both PRs on their next rebase and does not leave `master` to be fixed separately later.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where funding activity for Arkade swap offers could be missing from transaction history.
  - Swap offer funding now appears correctly even when the underlying transaction is not initially reported in history.
  - Prevented duplicate entries by replacing standalone funding activity with the corresponding swap activity when it becomes available.

- **Tests**
  - Added coverage for swap funding activity, including pending, cancelled, balance, and transaction-history scenarios.

### DIFF
--- a/src/lib/activityHistory.ts
+++ b/src/lib/activityHistory.ts
@@ -199,6 +199,24 @@ const ungroupedLnSendTx = (send: LnSendView, metadata: Record<string, Transactio
   )
 
 /**
+ * One row for an offer whose funding tx Arkade's history does not report.
+ *
+ * Same shape of problem as `ungroupedLnSendTx`, same answer: `createOffer`
+ * registers the covenant as a contract of this wallet, so the deposit reads as
+ * change and `buildTransactionHistory` drops the tx on its `txAmount !== 0`
+ * guard. A resting offer may never be spent, so nothing would ever appear.
+ * Keyed as the resolver would key it, so the real row replaces this one.
+ */
+const ungroupedOfferTx = (
+  swap: WalletAssetSwap,
+  { network, assetDisplay }: Pick<ActivityHistoryOptions, 'network' | 'assetDisplay'>,
+): Tx => ({
+  ...buildAssetSwapActivityTx(swap, [], { network, assetDisplay }),
+  amount: swap.fromAsset === 'btc' ? Number(swap.fromAmount) : 0,
+  historyKey: `swap:${swap.id}`,
+})
+
+/**
  * One row for a unilateral exit, which Arkade's history does not report.
  *
  * The sent side of `buildTransactionHistory` is gated on `isSpent`, and the SDK
@@ -290,6 +308,11 @@ export const activitiesToTxs = (activities: Activity[], options: ActivityHistory
   const grouped = new Set(activities.flatMap((activity) => rfqIdOf(activity) ?? []))
   for (const send of lnSends) {
     if (!grouped.has(send.rfqId)) rows.push(ungroupedLnSendTx(send, metadata))
+  }
+  // The offers history cannot see either — see `ungroupedOfferTx`.
+  const groupedSwaps = new Set(activities.flatMap((activity) => swapIdOf(activity) ?? []))
+  for (const swap of swaps) {
+    if (!groupedSwaps.has(swap.id)) rows.push(ungroupedOfferTx(swap, { network, assetDisplay }))
   }
   // Exits last, for the same reason: history reports none of them either.
   for (const exit of exits) rows.push(exitTx(exit))
--- a/src/test/lib/offerFunding.test.tsx
+++ b/src/test/lib/offerFunding.test.tsx
@@ -1,0 +1,153 @@
+import { renderHook } from '@testing-library/react'
+import { ReactNode } from 'react'
+import { describe, expect, it } from 'vitest'
+import { gatedContracts, type Contract } from '@arkade-os/sdk'
+import { activitiesToTxs } from '../../lib/activityHistory'
+import { ASSET_SWAP_ACTIVITY_KIND } from '../../lib/activity/assetSwapResolver'
+import { usePortfolioFiat } from '../../hooks/usePortfolioFiat'
+import { AspContext } from '../../providers/asp'
+import { FiatContext } from '../../providers/fiat'
+import { WalletContext } from '../../providers/wallet'
+import { mockAspContextValue, mockFiatContextValue, mockWalletContextValue } from '../screens/mocks'
+import type { WalletAssetSwap } from '../../lib/swapRepository'
+
+// The operator's mutinynet report: 110_000 sats held, 50_000 committed to a
+// btc-to-asset offer wanting 206 units. Balance never moved and no row appeared.
+const HELD = 110_000
+const COMMITTED = 50_000
+const FUNDING_TXID = 'dab4578b615fc6a9e3070e7041b2a6314d6515c21f6ac1d72fe8830cba2cfe8f'
+const COVENANT_SCRIPT = '51207b310c1298990c4b1dd2fba639039e75dc421e37b503428c069ae5797bad1db6'
+const WANT_ASSET = 'f1'.repeat(34)
+
+const contract = (over: Partial<Contract>): Contract => ({
+  type: 'default',
+  script: '5120' + 'bb'.repeat(32),
+  address: 'tark1own',
+  params: {},
+  state: 'active' as Contract['state'],
+  createdAt: 4_000,
+  ...over,
+})
+
+// as `registerOfferContract` writes it in @arkade-os/swap
+const offerContract = contract({
+  type: 'arkade',
+  script: COVENANT_SCRIPT,
+  address:
+    'tark1qqcpq7yq3e8hhsx6ml3fud93m7827qggaurtzu3zwsr4a0qs0gf857e3psff3xgvfvwa97ax8ypeuawugg0r0dgrg2xqdxh909a668dkrqlehc',
+  label: 'Arkade swap offer',
+  metadata: { genericallySpendable: false, kind: 'asset-swap-offer' },
+})
+
+const ownContract = contract({})
+
+const offerSwap = (over: Partial<WalletAssetSwap> = {}): WalletAssetSwap =>
+  ({
+    id: FUNDING_TXID,
+    fromAsset: 'btc',
+    toAsset: WANT_ASSET,
+    fromAmount: String(COMMITTED),
+    toAmount: '206',
+    swapAddress: 'tark1qqcpq7yq...',
+    swapPkScript: COVENANT_SCRIPT,
+    offerHex: '0100',
+    fundingTxid: FUNDING_TXID,
+    status: 'pending',
+    createdAt: 4_000,
+    ...over,
+  }) as WalletAssetSwap
+
+const empty = { swaps: [], metadata: {} }
+
+function portfolioWrapper({ children }: { children: ReactNode }) {
+  return (
+    <AspContext.Provider value={mockAspContextValue as never}>
+      <FiatContext.Provider value={{ ...mockFiatContextValue, toFiat: (sats?: number) => sats ?? 0 } as never}>
+        <WalletContext.Provider
+          value={
+            {
+              ...mockWalletContextValue,
+              balance: HELD,
+              availableBalance: HELD - COMMITTED,
+              assetBalances: [],
+              availableAssetBalances: [],
+            } as never
+          }
+        >
+          {children}
+        </WalletContext.Provider>
+      </FiatContext.Provider>
+    </AspContext.Provider>
+  )
+}
+
+describe('funding an Arkade swap offer', () => {
+  it('gates the covenant, so the spendable balance cannot select it', () => {
+    const gated = gatedContracts([offerContract, ownContract])
+
+    expect(gated.has(COVENANT_SCRIPT)).toBe(true)
+    expect(gated.has(ownContract.script)).toBe(false)
+    // fail-closed, so the funds stay gated even if the marker is ever dropped
+    expect(gatedContracts([{ ...offerContract, metadata: {} }]).has(COVENANT_SCRIPT)).toBe(true)
+  })
+
+  it('offers only what is left to spend, while still reporting the coins as owned', () => {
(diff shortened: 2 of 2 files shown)


Kukks:
On the docstring-coverage warning: `ungroupedOfferTx` does carry a JSDoc block (`src/lib/activityHistory.ts` L201-209) explaining why the row has to be synthesised at all — the check appears not to attribute it to an arrow-function `const`. Not adding more prose here on purpose: the comment budget on this diff is already 16 lines out of 166 added, and the discovery narrative belongs in the commit message and this description rather than in the file.

Kukks:
**Short answers.** Nothing to port to the activity resolver — a resolver structurally *cannot* express this case, so the three patches here are not working around a facility that exists. And the concept is not duplicated: what looks like two activity models is three stages of one pipeline. But the fix does belong upstream, one layer *below* the resolver, and it is already open as **ts-sdk#859** — which retires two of this file's three synthesized rows on its own.

Read against `origin/master` on both repos: wallet at `ea9ef551` (this PR's merge), ts-sdk at `43d002ce`, plus `pull/859/head` at `c5cfeb32`.

---

## 1. Can a resolver express the offer funding? No — it never sees the transaction

The layering is strictly one-directional:

```
vtxos --buildTransactionHistory--> ArkTransaction[] --buildActivities--> Activity[]
```

`getActivityHistory()` is one line — `buildActivities(await this.getTransactionHistory(), this.activity.all())` (`packages/ts-sdk/src/wallet/wallet.ts:1505-1506`; the worker wallet this app actually uses is identical at `packages/ts-sdk/src/wallet/serviceWorker/wallet.ts:1115-1116`).

`buildActivities` is a projection over a list it is handed: `for (const tx of txs)` (`packages/ts-sdk/src/wallet/activity.ts:143`), and every bucket it creates comes from a tx in that loop (`:161-178`). No `ArkTransaction` enters it except through its `txs` argument (`:102-105`). There is no synthesis path.

And the funding transaction is not in that list. Walking `buildTransactionHistory` over the three coins the funding leaves — the 110,000 input (`isSpent`, `arkTxId = F`), the 50,000 covenant output (`txid = F`), the 60,000 change (`txid = F`):

- neither output emits a RECEIVED row: both are excluded by `fromOldestVtxo.filter((v) => v.arkTxId === vtxo.txid).length === 0` (`packages/ts-sdk/src/utils/transactionHistory.ts:160`)
- the spent input's SENT row nets `changes` = *both* outputs (`:181`), so `txAmount = 110_000 − 110_000 = 0` (`:192`) and it dies on `if (txAmount !== 0 || assets)` (`:207`)

So `resolve(tx)` is never called with `F`.

**On `GroupMembership.amount` specifically:** it re-attributes an amount *across groups, for a transaction that already exists*. That is its documented purpose — "use it to split a batched tx across groups" (`activity.ts:30-37`) — and mechanically it is an override applied at bucketing time, `signedAmount(tx, m.amount ?? tx.amount)` (`:176`). It cannot conjure a row. So: **no, it cannot express the offer-funding case** — and not because `amount` is under-powered, but because there is nothing to attach it to.

That is why the fix here had to be a synthesized `Tx`. `ungroupedOfferTx` (`src/lib/activityHistory.ts:210`) and `ungroupedLnSendTx` (`:167`) are not routing around the resolver facility; they are routing around the row never reaching it.

## 2. The right layer already has an open PR

The asymmetry is one line wide. Both reads take the same snapshot:

- `getBalance()` keeps the contract identity and derives the gate from it — `gated: gatedContracts(snapshot.map((_) => _.contract))` (`wallet.ts:1418`), consumed as `isGenericallySpendable: (vtxo) => !gated.has(vtxo.script)` (`:1438`).
- `getTransactionHistory()` reads the same `getContractsWithVtxos()` and then discards the contract — `const allVtxos = response.flatMap((_) => _.vtxos)` (`:1484`) — handing `buildTransactionHistory` a bare `VirtualCoin[]`.

The SDK already knows the covenant is not spendable change. The history path throws that away one line after reading it, which is precisely why the netting is wrong.

**ts-sdk#859, "gate transaction history the way balance is gated", is that fix.** It adds a fifth `gatedScripts: GatedContracts` parameter and partitions gated coins out of `ownVtxos`, reading them as an external counterparty (confirmed on `pull/859/head` @ `c5cfeb32`). Open and mergeable against master, with the batch/exit corner carved out into #860.

### Two of the three patches here retire themselves when it lands

Neither needs a follow-up to delete it — their own guards stop firing:

- **Offers.** `assetSwapResolver.prepare` indexes `swap.fundingTxid → swap.id` (`src/lib/activity/assetSwapResolver.ts:25`), and `swap.id === swap.fundingTxid ===` the txid `send()` returned (`src/providers/assetSwaps.tsx:268,276`). Once the funding tx is in history it groups, `groupedSwaps` contains the id, and `ungroupedOfferTx` is skipped (`activityHistory.ts:313-316`) in favour of the real grouped row at `:284-296`.
- **Lightning sends.** Same shape through `swapActivityResolver`, which indexes `record.fundingArkTxid` (`packages/swap/src/activity.ts:163`) and groups on `swap:${rfqId}` (`:113`); `grouped` then covers the rfq id and `ungroupedLnSendTx` is skipped (`activityHistory.ts:308-311`).

The framing this PR added is not lost either: `buildAssetSwapActivityTx` derives `status` from the record, not from the transactions (`src/lib/swapDisplay.ts:220`) — `members` only contributes `amount: members[0]?.amount ?? 0` and the fill lookup. The commitment reads as recoverable the same way; it just stops being synthesized.

### `exitTx` does not retire

Its cause is unrelated: the SENT arm is gated on `isSpent`, which is documented as *"not set to true if the virtual output is unrolled or swept, only when it's spent offchain"* (`packages/ts-sdk/src/wallet/index.ts:817`). #859 touches none of that. So it is two patches sharing one root cause plus a third with its own — worth stating, since "three patches, one misplaced fix" slightly overstates it.

### One consumer a display-layer fix cannot reach

`resolveCancellingSpend` searches history for the transaction that spent the covenant — `(await getTxHistory(svcWallet)).find(...)` (`src/providers/assetSwaps.tsx:339`, via `src/lib/asp.ts:192`). By the same netting a BTC-denominated offer cancel emits no row either (covenant in, returned coin out, `txAmount = 0`), so that lookup returns `undefined` and the reconciliation gives up. **Inferred from the code path, not executed.** If it holds, it is the sharpest argument against a fourth patch in `activitiesToTxs`: this consumer does not read `activitiesToTxs` at all. Only the upstream fix serves it.

## 3. Is the activity concept duplicated? No — three stages, not two implementations

1. **`buildTransactionHistory`** (`packages/ts-sdk/src/utils/transactionHistory.ts`) — *derivation*: coins → transactions. "What moved."
2. **`buildActivities` + `ActivityResolver`** (`packages/ts-sdk/src/wallet/activity.ts`) — *grouping*: transactions → labelled logical actions. "What were those transactions, together."
3. **`activitiesToTxs`** (`src/lib/activityHistory.ts:260`) — *view model*: `Activity[]` → the `Tx[]` rows the screens read.

Stage 2 does not supersede stage 1 — it consumes it, and `getTransactionHistory` is neither deprecated nor bypassed. Not a half-finished migration: a pipeline that grew a grouping stage, with both stages load-bearing by design.

Stage 3 is a view model, not a rival activity model. This wallet is fully on the resolver contract — `svcWallet.activity.use(assetSwapResolver())` (`src/providers/wallet.tsx:663`) and `swapActivityResolver` from `@arkade-os/swap` (`:682`), rendered through `activitiesToTxs(history.activities, …)` (`:224`). `src/lib/activity/assetSwapResolver.ts` and `src/lib/activityHistory.ts` are not two answers to one question: the first is registered *into* the SDK, the second reads what comes *out*.

The only genuine overlap is much narrower than "duplicated concept": two independent resolvers share the `swap:` group namespace — `swap:${swap.id}` (`assetSwapResolver.ts:33`) and `swap:${swap.rfqId}` (`packages/swap/src/activity.ts:113`) — against the SDK's own advice to namespace by vendor (`activity.ts:7-13`). It is already recognised and handled: the comment at `activityHistory.ts:54-56` says the corridor is told apart by `swapKind`, never the group id. Since one id space is an ark txid, they cannot actually collide. Hygiene, not a defect.

## What I would port, and where

- **To `buildTransactionHistory`:** the gate. Already open as #859 — nothing new to write.
- **To the `ActivityResolver` layer:** nothing. The contract is not the wrong shape and is not missing `amount` semantics; it just runs downstream of the stage that lost the row.
- **Here, now:** nothing. When #859 lands, `ungroupedOfferTx` and `ungroupedLnSendTx` become branches that stop firing — worth deleting then, with the double-row case pinned by a test, not before.
- **Still unowned:** `exitTx`'s cause. If unilateral exits should reach history without every consumer synthesizing them, the SENT arm needs a signal other than `isSpent`. Separate SDK question.

## The claim I would most expect to be wrong

That the synthesized rows retire *cleanly* rather than doubling up. It rests on `swap.fundingTxid` being the same string as `tx.key.arkTxid` on the funding row — `send()`'s return value versus what history keys on. Every local use treats them as one value (`vtxos.find((v) => v.txid === swap.fundingTxid)`, `src/providers/assetSwaps.tsx:320`), and `@arkade-os/swap` is explicit that the checkpoint txid and the ark txid are different things (`packages/swap/src/activity.ts:211-213`). If `send()` ever returns the checkpoint, the guards miss and #859 yields a second row beside the synthesized one instead of replacing it. Cheap to pin with one test against #859's branch before it merges.

PR #967, opened 2026-09-08
fix(swaps): classify an unseen spend by its covenant leaf, not by history

tiero:
## Problem

A filled BTC→DePix swap rendered as **Cancelled**. Regression from #962.

`reconcileUnseenSpends`, added in #962, classified a deposit spend with `isCancelSpend`: read the wallet's history row for the spending tx, call it a cancel unless the want-asset is on it. That row is built by the SDK by netting the deposit against the outputs the same tx created for the wallet. The pass's two reads are not equally fresh:

- `getContractsWithVtxos` calls `syncContracts`, a live indexer fetch of the covenant script. The deposit shows spent the instant the fill lands.
- `getTxHistory` goes to the worker's `GET_TRANSACTION_HISTORY`, which builds from `getVtxosFromRepo`. The fill's output on the wallet script arrives only when the subscription delivers it.

A pass in that window sees the deposit spent and no output for it: a bare 50K-sat send, no assets. `isCancelSpend` calls that a cancel, `spendUpdate` writes it, and a stored status is terminal. The pass fires on every history change, including the reload `createSwap` triggers, so a fast fill hit it reliably.

Reproduced against mainnet on the reported swap (deposit `f5703efc…`, spent via checkpoint `fb1a9d30…` by ark tx `7ebab1ee…`): the SDK's own history builder over the real vtxos gives a row with DePix when the fill output is present and a bare send without it; `isCancelSpend` flips accordingly. `classifyDepositSpend` over both spend txs says `fulfilled`. A fresh browser rebuilt the same record as Completed because the only writer it has, the restore scan, classifies by leaf.

## Fix

Classify from the covenant leaf, as the watcher and the restore scan already do. `classifyDeposit` hands the package's `classifyDepositSpend` every tx that spent the deposit (`spendTxidsOf`: the checkpoint carries the leaf, not the ark tx), read off the indexer in one call per pass. No answer leaves the record open for the next pass. The completion time comes from the fill's own row in `ungroupedTxs` when it has landed, else now; the outcome does not wait for it. BTC-want offers, which the row could never answer, resolve too.

Both writers that start from a spent deposit, the unseen-spend pass and a cancel that threw after the deposit was gone, now share one `resolveSpentDeposit`: classify, re-read, `spendUpdate`, write, announce. That was the last caller of `isCancelSpend`, whose own TODO asked for this, so `src/lib/swapSpend.ts` goes. The x-only server key goes through the SDK's `toXOnlySignerHex` at both sites instead of a bare slice.

## Not included

Records the shipped pass already wrote as `cancelled` are left alone: terminal by design. A restore rebuilds them correctly from chain. A one-shot leaf re-verification of `cancelled` records is possible if wanted.

## Testing

- New tests: the pass fed exactly the bare row writes `fulfilled`; both spend txids and the deposit outpoint reach the classifier against the x-only key; two spent swaps are read in one indexer call; the failed-cancel path classifies by leaf.
- `src/test/providers/assetSwaps.test.tsx`: 30 pass. Full unit suite: 90 files, 794 pass.
- tsc, eslint, prettier, production build clean.

https://claude.ai/code/session_011bDQfhxXV2si3GHWrkrouU

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asset swap status reconciliation using transaction and covenant details for more accurate results.
  * Correctly recognizes completed swaps and cancellations, including offers involving BTC as the requested asset.
  * Recovers swap records after restores, cancellations, and previously unseen transactions.
  * Keeps swap records pending when transaction details cannot be conclusively classified.
  * Rechecks reconciliation when server connection or signing settings change.


tiero:
@arkana-ai-bot review

---

tiero:
@coderabbitai review

---

tiero:
CI `test` on `bc6a745` failed on one unit test: `src/test/screens/wallet/swap.test.tsx › quotes the minimum in the asset being sent (BTC), not the asset being received (USDT)`, a 5000 ms timeout at 5033 ms.

Not this PR's. That test renders the swap screen under a hand-built `AssetSwapsContext.Provider` value (`swap.test.tsx:112`), not `AssetSwapsProvider`, so none of the code this PR changes executes in it. The same job was green on the two previous heads of this branch (`8b54515`, `6454689`), and `bc6a745` changes only a `useMemo` dependency list and a comment. Locally the test passes three of three at about 1.7 s.

Re-running the failed job once. A second failure gets root-caused, not re-run.

---

tiero:
Correction to the above: the re-run was refused (403 for this integration), so no re-run happened.

Master moved under the PR meanwhile (#968, which adds tests to `swap.test.tsx`), so the failed job ran against a merge state the branch had not been tested on. Merged `master` in as `f8befa0`; the full `swap.test.tsx` file passes locally on that exact state (42 tests, including the one that timed out and #968's new one). CI runs again on the merged head.

---

tiero:
CI `test` on `f8befa0` failed on a different test in the same untouched file: `swap.test.tsx › stops a partial asset swap that cannot fund its change carrier` (added by #968), `waitFor` at the default 1 s not finding the toast while the DOM still showed a spinner. It passed on the previous run, on master's own run, and locally on this exact head (42/42).

Same reasoning as above: this screen only imports the `AssetSwapsContext` object and never renders `AssetSwapsProvider`, so none of this PR's code runs in it. Two consecutive runs missing two different `waitFor`s in one file is a slow runner, and I cannot re-run (403).

Pushed `e44262b`: that one assertion now gets the 3 s allowance the Sonner test two cases above already uses for the same toast path. One line, no behaviour change. Happy to drop it if you would rather keep #968's test untouched.

---

tiero:
`test` is green on `e44262b`. One E2E job on that head, `E2E (core / Google Chrome)`, failed before any test body ran: the regtest stack never came up (`mempool_api` in a restart loop, "mempool Esplora API did not become ready in time"). Its sibling `E2E (core / Mobile Chrome)` on the same commit passed. Regtest docker infrastructure, not this PR's, and no fix to port. I cannot re-run it (403), so it needs a re-run from someone with Actions access, or the next push.

---

tiero:
On the `fetchSpendTxs` error-propagation point: already caught one level up, deliberately. The whole cancel-recovery block in `cancelSwap` sits inside `try { … } catch { /* indexer unreachable: keep 'cancelling'; the watcher resolves it later */ }` followed by `throw err`, so an unreachable indexer during `getVirtualTxs` leaves the record `cancelling` and surfaces the original cancel error, exactly as before. In the reconcile pass the same throw lands in the queue's `.catch(consoleError)` and the next history change re-enters. Catching inside `fetchSpendTxs` would only hide the failure from that log, so leaving it as is.

Agreed on the test symmetry note: the batching test owns the txid-collection assertion now.

---

tiero:
@coderabbitai review

The Merge Risk banner is assessed against `6454689` and cites the context memo missing `aspInfo.signerPubkey`. That landed in `bc6a745`; the later commits are a master merge (`f8befa0`) and a one-line test allowance (`e44262b`).

---

PR #972, opened 2026-09-09
TEMPORARY: verification probe for #966 - do not review, closing immediately

Kukks:
Throwaway probe for #971 / #966. Base is `tmp/ci966-verify-base`, not `master`, and that base carries the trigger fix. Purpose is only to observe whether `test` and the four `E2E` checks are created on a non-master base. Closed and both branches deleted as soon as the check list appears.

### DIFF
--- a/.ci966-trigger-probe
+++ b/.ci966-trigger-probe
@@ -1,0 +1,1 @@
+temporary probe for issue 966 verification; branch is deleted immediately after


Kukks:
Probe complete. It confirmed what it was for: base `tmp/ci966-verify-base` is not `master` or `next`, and this PR still received `test` plus all four `E2E` checks (runs 34330795127 and 34330795257, both event `pull_request`). Runs cancelled to avoid burning runner time; branches deleted. See #971.

PR #973, opened 2026-09-09
TEMPORARY probe A - do not review, closing shortly

Kukks:
Throwaway. Base is a copy of `ts-sdk-0.5` WITHOUT the #971 fix. Observing whether any test check is created. Deleted shortly.

### DIFF
--- a/.ci966-probeA
+++ b/.ci966-probeA
@@ -1,0 +1,1 @@
+probe A: base lacks the fix


Kukks:
Probe complete — this one was the control. Base was a copy of `ts-sdk-0.5` **without** #971, and after 3.5 minutes it had only the two Cloudflare Pages builds: no `test`, no E2E, and no workflow run created at all. Confirms that #971 on `master` does not gate PRs into an integration branch that hasn't merged it.

PR #975, opened 2026-09-09
TEMPORARY probe C - do not review, closing shortly

Kukks:
Throwaway. Base is a copy of `ts-sdk-0.5` WITHOUT the #971 fix; the HEAD has `origin/master` merged in so it DOES carry the fix. Isolating whether the base branch or the merge commit decides. Deleted shortly.

### DIFF
--- a/.ci966-probeC
+++ b/.ci966-probeC
@@ -1,0 +1,1 @@
+probe C: head has the fix, base does not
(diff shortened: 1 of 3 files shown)


Kukks:
Probe complete. Base was a copy of `ts-sdk-0.5` **without** #971 while the head had `origin/master` merged in, so only the head carried the fix — and `test` plus all four E2E checks were still created. So the workflow is resolved from the **merge commit**, not the base branch alone: either side carrying the fix is enough. Runs cancelled, branches deleted.

PR #979, opened 2026-09-10
Show placeholders while swap restore scan is unsettled

tiero:
## Summary
This change prevents a visual flash in the transaction list by showing skeleton placeholders while the asset swap restore scan is determining which transactions are part of swaps. Without this, users would briefly see ungrouped "Sent" and "Received" rows that would then be replaced by a single "Swap" row once the scan completes.

## Key Changes
- **Added `swapScanSettled` state to AssetSwapsContext**: A boolean flag that tracks whether the restore scan has finished processing all funding transactions. Defaults to `true` to avoid placeholders on wallets that already have complete history.

- **Implemented scan settlement logic**: 
  - The gate is lowered (`swapScanSettled = false`) before fetching when there are unanswered funding transactions
  - The gate is raised (`swapScanSettled = true`) after the scan completes, even if it throws (to prevent hiding history behind placeholders during indexer outages)
  - A queued rescan keeps the gate down since it's part of the same answer

- **Added `TransactionLineSkeleton` component**: A placeholder row that matches the height and layout of real transaction rows, with animated shimmer effect to indicate loading state.

- **Updated TransactionsList to show placeholders**: When `swapScanSettled` is false and there are transactions, the list renders skeleton placeholders instead of actual transaction rows.

- **Added CSS styling**: New `.activity-row--skeleton` and `.activity-skeleton` classes with shimmer animation that respects prefers-reduced-motion.

- **Added comprehensive tests**: Three new test cases verify the placeholder behavior during unsettled scans, that settled scans don't show placeholders, and that errors don't leave placeholders stuck.

## Notable Implementation Details
- The settlement gate is raised before the first fetch (not after) by checking local state, ensuring wallets with nothing to rebuild never show placeholders
- The predicate for detecting unanswered funding transactions is duplicated from `@arkade-os/swap` with a TODO to export it from that package to prevent drift
- Skeleton rows are marked with `aria-hidden='true'` since they're temporary UI artifacts
- The shimmer animation uses a gradient that moves across the placeholder to indicate loading

https://claude.ai/code/session_014rffYCc641C2biRsyoVWXf

### DIFF
--- a/src/components/SkeletonText.tsx
+++ b/src/components/SkeletonText.tsx
@@ -1,0 +1,12 @@
+/**
+ * One shimmering bar standing in for text that has not arrived.
+ *
+ * Shared rather than per-screen: the swap composer waits on a quote, the
+ * activity list waits on the restore scan, and two shimmers tuned separately
+ * read as two different loading states.
+ */
+export default function SkeletonText({ width, className }: { width: string; className?: string }) {
+  return (
+    <span className={className ? `skeleton-text ${className}` : 'skeleton-text'} style={{ width }} aria-hidden='true' />
+  )
+}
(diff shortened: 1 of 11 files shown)


tiero:
@arkana-ai-bot review

PR #996, opened 2026-09-15
fix wallet page flickering

bordalix:
@sahilc0 do you see any diff with your iPhone?

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the staggered boot animation from the wallet home screen, allowing it to load without the introductory animation.


tiero:
I'm bit against to remove the animation tout court, since the proper [fix](https://github.com/arkade-os/wallet/pull/995) isn't that big of a change. Before wasn't happening beacuse we had the tabs guarding for "intial boots" pages

PR #1004, opened 2026-09-16
feat(lnurl): receive lightning through the registered address

Kukks:
Stacked on #1000. Moves the Receive screen's lightning rail off the wallet's own RFQ negotiation and onto the registered lnurl address.

**Draft only because #1000 is.** The registration path landed in `ff621358`, so this no longer removes a working rail without a replacement.

The server is read from `VITE_LNURL_SERVER` rather than defaulted. Which lnurl-server a wallet trusts with its receive identity is not a choice to bake into a binary: the token is a bearer credential and the registered Arkade address is where the covenant pays, so a wrong default hands both to someone the user never picked. **With none configured the screen says so and offers no button**, which is also why this ships dark until you point a build at a server.

## Why

The wallet was negotiating a solver RFQ to mint its own hold invoice, which requires the wallet **open** and an amount **typed first**. The lnurl-server does the same three rails server-side and mints an invoice per payer, so a payment can arrive while the wallet is closed — which is the entire point of the offline rails.

## What changed

`encodeBip21` already had an `lnurl` parameter emitting `lightning=<lnurl>` when there is no invoice (`bip21.ts:92`), so one QR still serves every rail and the integration is a single argument.

Removed from the screen — roughly 90 lines: solver discovery, the rendezvous bounds check, the RFQ transport, `requestLnReceive`, the three negotiation error states and the retry button. The amount gate goes too: a lightning address needs no amount where a solver-minted hold invoice did.

## What deliberately did not change

- **`LnReceiveProvider` stays mounted.** It owns the `RfqSwapManager` the rest of the app uses, and `claimReceive`/`toReceiveOrigin`/`toReceiveSwap` still serve receives negotiated before this branch. Unmounting it breaks lightning *sends*.
- **`requestLnReceive` is left in place though it is now dead**, along with `toReceiveSwap`, `toReceiveOrigin` and the provider'''s `track`. The separability is now verified rather than assumed: claiming reads the stored record via `assetSwapRepository.getRfqSwap` (`providers/lnReceive.tsx:195`), while those three run only at creation inside `track` (`:310-311`). So a receive negotiated before this branch still claims with the creation path gone. It is not removed here because `track` shares the multi-tab lock machinery (`granted`, `LOCK_GRACE_MS`) with the claim driver, and unpicking that is a careful change on money-handling code rather than a tidy-up to append to this one.
- **Assets and on-chain boarding are untouched.** lnurl-server serves neither.
- **The Arkade destination is still locally derived.** Routing it through the server would mean a per-payment covenant destination, which buys payment attribution but adds a covclaimd dependency to the one path that currently cannot fail (`covenant-destination.ts:121` throws without `covclaimdUrl` or `emulatorUrl`). Assets ride on that same address, and whether `enforcePayTo` covers asset VTXOs is unverified — getting it wrong means unclaimable funds. Both need an answer before that moves.

## Tests

`receive-lightning-error.test.tsx` covered the deleted negotiation and went with it; its only non-lightning references were fixture addresses. Replaced by `receive-lnurl.test.tsx`, which pins the unavailable line and its absence once an address is registered.

- `pnpm test` — **98 files / 852 passed + 1 skipped**, exit 0
- `pnpm exec tsc --noEmit` — exit 0
- `prettier` / `eslint` / `csp:check` — clean


Kukks:
Superseded by #1000, which now carries this PR's work (receive through a registered lnurl address), rebuilt on `ts-sdk-0.5` and ported onto the `@arkade-os/lnurl-client` facade. The solver-RFQ receive this PR removed is kept there as its own module (`swapRail`), used whenever `VITE_LNURL_SERVER` is unset.

PR #1010, opened 2026-09-18
fix(native-runtime): hide install banner and fall back to legacy copy in webviews

Delgado74:
### Summary

Two small native-runtime tweaks for webviews and in-app browsers, where the wallet is opened inside an embedded renderer instead of a real browser.

### Changes

- **`fix(pwa)`: hide install banner in in-app browsers and webviews** — `pwaCanInstall()` now also returns `false` inside in-app browsers / embedded webviews (reuses the existing `isInAppBrowser()` detector). Those contexts never fire `beforeinstallprompt` and offer no "Add to Home Screen" entry point, so the install banner was misleading there. New tests in `src/test/lib/pwa.test.ts`.
- **`fix(clipboard)`: fall back to legacy copy in webviews** — `copyToClipboard()` now falls back to the `document.execCommand('copy')` + temp textarea path when `navigator.clipboard` is missing (many webviews) or rejects the write. New tests in `src/test/lib/clipboard.test.ts`.

### Validation

- `pnpm exec tsc --noEmit` clean
- `pnpm exec eslint` clean (all `src/**`)
- `pnpm exec prettier --check src` clean
- `pnpm vitest` -> 8 new tests pass (2 files); full suite green except 7 pre-existing `asset.test.ts` failures that only reproduce when the machine locale is non-en-US (grouping asserted as `,`), not on CI runners


Delgado74:
Addressed the remaining notes from the pass — see [6b5b049b](https://github.com/arkade-os/wallet/commit/6b5b049b):

- New test for the both-paths-fail branch: `consoleError` fires with 'error copying via legacy fallback' and the copy resolves cleanly.
- `pwa.test.ts`: `serviceWorker` stub is removed in `afterEach`, and `pwaIsInstalled()` now covers the iOS `navigator.standalone` and `display-mode: standalone` paths.
- `clipboard.test.ts`: textarea visibility during `execCommand` is asserted via a captured flag instead of an in-mock `expect`.
- `clipboard.ts`: comment now calls out that the legacy path briefly attaches the textarea to the DOM (a body MutationObserver can read `textarea.value`), which matters for the Backup.tsx callsite passing mnemonics/nsec.
- Also restored `toLocaleString('en-US')` in `prettyAssetNumber` — without a locale the grouping depends on the runtime environment (e.g. es_ES produces dot separators), breaking cross-machine parity.

Full suite: 850 passed | 1 skipped; tsc/eslint/prettier clean.

